PR #19, opened 2018-07-09
New option in plot_diff : save figure

minouHub:
minor improve of the get_diff function of class spectrum


erwanp:
pushed in `develop` directly

Issue #65, opened 2020-02-04
bug in pandas conversion to np.int64


erwanp:
### Error

P, Q, R branches in HITRAN line databases are converted to numeric -1, 0, 1 to improve performances.

This recently stopped working in the test job of an unrelated change (https://travis-ci.com/radis/radis/jobs/283163452) : 

```
radis/io/hitran.py:359: in hit2df

    def replace_PQR_with_m101(df):
        if df.dtypes["branch"] != np.int64:
            df.loc[df.branch == "P", "branch"] = -1

[...]

       assert df.dtypes["branch"] == np.int64

>>> AssertionError
```

### Diagnostic:

- No change in RADIS, no change in HITRAN database (this happens on hardcoded fragments of the database too) 
- It may be that Pandas does not automatically update the column dtypes anymore?
- If the error is due to Pandas update, it will hit all existing installations and should be fixed ASAP. 

### Fix:

- [ ] update latest Pandas 
- [ ] check what is the dtypes
- [ ] try to convert to np.int64
- [ ] if it does not work, dont crash. The code will be a little slower but it shouldnt be fatal.

minouHub:
Ok guys, the panda package was updated on the 29th of January, so that may be it. I'm on it

minouHub:
Quick update: this is a known issue of pandas
https://github.com/pandas-dev/pandas/issues/4094

I did a quick and dirty fix of the issue. Also I shortened the synthax

erwanp:
#67 was fixed, so as soon as your changes are updated our tests should pass again!

PR #66, opened 2020-02-04
sort the problem raised by the update of pandas

minouHub:
To sort #65

### DIFF
--- a/radis/io/tools.py
+++ b/radis/io/tools.py
@@ -200,9 +200,9 @@ def replace_PQR_with_m101(df):
     """
 
     # Note: somehow pandas updates dtype automatically. We have to check
-    # Actuelly it's a know issue. https://github.com/pandas-dev/pandas/issues/4094
     # We just have to replace the column:
-    new_col = df["branch"].replace(["P", "Q", "R"], [-1, 0, 1])
-    df["branch"] = new_col
+    if df.dtypes["branch"] != np.int64:
+        new_col = df["branch"].replace(["P", "Q", "R"], [-1, 0, 1])
+        df["branch"] = new_col
 
     assert df.dtypes["branch"] == np.int64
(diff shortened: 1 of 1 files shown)


radis/io/tools.py
```diff
@@ -200,12 +200,9 @@ def replace_PQR_with_m101(df):
...
-
-    if df.dtypes["branch"] != np.int64:
-        df.loc[df.branch == "P", "branch"] = -1
-    if df.dtypes["branch"] != np.int64:
-        df.loc[df.branch == "Q", "branch"] = 0
-    if df.dtypes["branch"] != np.int64:
-        df.loc[df.branch == "R", "branch"] = 1
+    # Actuelly it's a know issue. https://github.com/pandas-dev/pandas/issues/4094
+    # We just have to replace the column:
+    new_col = df["branch"].replace(["P", "Q", "R"], [-1, 0, 1])
```
erwanp: does it work if you simply do 
```
df["branch"].replace(["P", "Q", "R"], [-1, 0, 1])
```
(without the `new_col`) ?

radis/io/tools.py
```diff
@@ -200,12 +200,9 @@ def replace_PQR_with_m101(df):
...
 
     # Note: somehow pandas updates dtype automatically. We have to check
-
-    if df.dtypes["branch"] != np.int64:
-        df.loc[df.branch == "P", "branch"] = -1
-    if df.dtypes["branch"] != np.int64:
-        df.loc[df.branch == "Q", "branch"] = 0
-    if df.dtypes["branch"] != np.int64:
-        df.loc[df.branch == "R", "branch"] = 1
+    # Actuelly it's a know issue. https://github.com/pandas-dev/pandas/issues/4094
```
erwanp: I'd remove that line

radis/io/tools.py
erwanp: After testing, it does not work (dtype is not changed). 
Let's do that instead (to prevent rewriting the column if nothing has to be changed)

```

    if df.dtypes["branch"] != np.int64:
        new_col = df["branch"].replace(["P", "Q", "R"], [-1, 0, 1])
        df["branch"] = new_col
```

PR #83, opened 2020-03-31
Pull request to solve issue #82. 

### LINKED ISSUE #82: Nan handeling in get_residual
Hello all,

When using `get_residual(s1, s2, normalize=True, ignore_nan=True)` I would expect an output which is not a Nan. However, if one of the spectra has Nan values, the residual will be a Nan.

**Why?**
Using the option `normalize=True`, the spectrum is normalized by (let's say) its maximum which is a nan. 

**Short term solution**
```
if normalize_how == "max":
                norm1 = np.nanmax(s1.get(var, copy=False)[1])
                norm2 = np.nanmax(s2.get(var)[1])
                # norm1 = s1.get(var, copy=False)[1].max()
                # norm2 = s2.get(var)[1].max() 
```

**Long term solution**
Should we handle each function like before, or should we ban the storage of Nan in spectrum?

Minou

adityabharadwaj198:
### Description

This pull request is to address ... the issue here https://github.com/radis/radis/issues/82. 
I've made the changes suggested in the comments related to the issue and pushed them to my forked repo. If anyone wants to run a test to see if the changes fix the issue, they can run this:
```
from radis import calc_spectrum, plot_diff, Radiance_noslit
                   molecule='CO',
                   isotope='1,2,3',
                   pressure=1.01325,   # bar
                   Tgas=700,           # K
                   mole_fraction=0.1,
                   path_length=1,      # cm
                   )
s=Radiance_noslit(s)   # keep only radiance 

import numpy as np
s._q['radiance_noslit'][0] = np.nan

# Something like this would give an error otherwise:
plot_diff(s, s*1.2, normalize=True)
```

Fixes #82

### DIFF
--- a/radis/spectrum/compare.py
+++ b/radis/spectrum/compare.py
@@ -383,13 +383,11 @@ def get_residual(
             if normalize_how == "max":
                 norm1 = np.nanmax(s1.get(var, copy=False)[1])
                 norm2 = np.nanmax(s2.get(var)[1])
-                #norm1 = s1.get(var, copy=False)[1].max()
-                #norm2 = s2.get(var)[1].max()
+
             elif normalize_how == "mean":
             	norm1 = np.nanmean(s1.get(var, copy=False)[1])
             	norm2 = np.nanmean(s2.get(var)[1])
-                #norm1 = s1.get(var, copy=False)[1].mean()
-                #norm2 = s2.get(var)[1].mean()
+
             elif normalize_how == "area":
                 norm1 = s1.get_integral(var)
                 norm2 = s2.get_integral(
@@ -785,12 +783,10 @@ def plot_diff(
         s2 = s2.copy()
         w1, I1 = s1.get(var, wunit=wunit, copy=False)
         w2, I2 = s2.get(var, wunit=wunit, copy=False)
-        #ratio = np.max(I1) / np.max(I2)
         ratio = np.nanmax(I1) / np.nanmax(I2)
         I1 /= np.nanmax(I1)
         I2 /= np.nanmax(I2)
-        #I1 /= np.max(I1)
-        #I2 /= np.max(I2)
+
         if verbose:
             print(("Rescale factor: " + str(ratio)))
 
@@ -996,7 +992,7 @@ def plot_diff(
         if show_residual:
             difftext += " (residual={0:.2g})".format(
                 get_residual(
-                    s1, s2, var=var, norm="L2", ignore_nan=True,diff_window=diff_window
+                    s1, s2, var=var, norm="L2", ignore_nan=True, diff_window=diff_window
                 )
             )
         pos = ax1[i].get_position()
--- a/radis/spectrum/compare.py
+++ b/radis/spectrum/compare.py
@@ -381,11 +381,15 @@ def get_residual(
             s2 = multiply(s2, 1 / norm2, var=var)
         else:
             if normalize_how == "max":
-                norm1 = s1.get(var, copy=False)[1].max()
-                norm2 = s2.get(var)[1].max()
+                norm1 = np.nanmax(s1.get(var, copy=False)[1])
+                norm2 = np.nanmax(s2.get(var)[1])
+                #norm1 = s1.get(var, copy=False)[1].max()
+                #norm2 = s2.get(var)[1].max()
             elif normalize_how == "mean":
-                norm1 = s1.get(var, copy=False)[1].mean()
-                norm2 = s2.get(var)[1].mean()
+            	norm1 = np.nanmean(s1.get(var, copy=False)[1])
+            	norm2 = np.nanmean(s2.get(var)[1])
+                #norm1 = s1.get(var, copy=False)[1].mean()
+                #norm2 = s2.get(var)[1].mean()
             elif normalize_how == "area":
                 norm1 = s1.get_integral(var)
                 norm2 = s2.get_integral(
@@ -781,9 +785,12 @@ def plot_diff(
         s2 = s2.copy()
         w1, I1 = s1.get(var, wunit=wunit, copy=False)
         w2, I2 = s2.get(var, wunit=wunit, copy=False)
-        ratio = np.max(I1) / np.max(I2)
-        I1 /= np.max(I1)
-        I2 /= np.max(I2)
+        #ratio = np.max(I1) / np.max(I2)
+        ratio = np.nanmax(I1) / np.nanmax(I2)
+        I1 /= np.nanmax(I1)
+        I2 /= np.nanmax(I2)
+        #I1 /= np.max(I1)
+        #I2 /= np.max(I2)
         if verbose:
             print(("Rescale factor: " + str(ratio)))
 
@@ -989,7 +996,7 @@ def plot_diff(
         if show_residual:
             difftext += " (residual={0:.2g})".format(
                 get_residual(
-                    s1, s2, var=var, norm="L2", ignore_nan=True, diff_window=diff_window
+                    s1, s2, var=var, norm="L2", ignore_nan=True,diff_window=diff_window
                 )
             )
         pos = ax1[i].get_position()
--- a/radis/spectrum/compare.py
+++ b/radis/spectrum/compare.py
@@ -25,7 +25,7 @@ Routine Listings
 """
 
 from __future__ import print_function, absolute_import, division, unicode_literals
-from radis.misc.arrays import array_allclose
+from radis.misc.arrays import array_allclose, nantrapz
 from radis.misc.curve import curve_substract, curve_distance, curve_divide
 from radis.spectrum.spectrum import Spectrum, is_spectrum
 from radis.spectrum.utils import format_xlabel, make_up, cast_waveunit
@@ -356,8 +356,11 @@ def get_residual(
             b = (w1 > wmin) & (w1 < wmax)
             if normalize_how == "max":
                 norm1 = I1[b].max()
+                # norm1 = np.nanmax(I1[b])
             elif normalize_how == "mean":
-                norm1 = I1[b].mean()
+                # norm1 = I1[b].mean()
+                # norm1 = np.nanmean()
+                norm1 = I1[b].nanmean()
             elif normalize_how == "area":
                 norm1 = np.abs(np.trapz(I1[b], w1[b]))
             else:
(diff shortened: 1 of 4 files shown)


erwanp:
Hello @adityabharadwaj198 , thanks for submitting!

Tests [fail for the moment](https://github.com/radis/radis/pull/83/checks?check_run_id=549318457). Don't worry about the Code Quality Checks, we'll fix that at the end. But you should have a look at the "Test and Coverage" one: seems there is a syntax error ! 

Fix and re-push!

radis/test/spectrum/test_compare.py
```diff
@@ -66,6 +67,20 @@ def test_compare_methods(verbose=True, plot=True, close_plots=True, *args, **kwa
...
+                   pressure=1.01325,   # bar
+                   Tgas=700,           # K
+                   mole_fraction=0.1,
+                   path_length=1,      # cm
+                   )
+    
+    s=Radiance_noslit(s)  
+    s._q['radiance_noslit'][0] = np.nan
+
+    plot_diff(s, s*1.2, normalize=True)
```
erwanp: in the test can you also try some other normalization options? Like, "max", "area", etc.

adityabharadwaj198: Sure

radis/spectrum/compare.py
```diff
@@ -381,11 +381,15 @@ def get_residual(
             s2 = multiply(s2, 1 / norm2, var=var)
         else:
             if normalize_how == "max":
-                norm1 = s1.get(var, copy=False)[1].max()
-                norm2 = s2.get(var)[1].max()
+                norm1 = np.nanmax(s1.get(var, copy=False)[1])
+                norm2 = np.nanmax(s2.get(var)[1])
+                #norm1 = s1.get(var, copy=False)[1].max()
```
minouHub: I believe you can remove the lines you commented

radis/test/spectrum/test_compare.py
minouHub: Hi, 
You even added a test! Very good :)
I just suggest you remove the lines you commented.
Minou

adityabharadwaj198: > Hi,
> You even added a test! Very good :)
> I just suggest you remove the lines you commented.
> Minou

Hey there, I did what you suggested and pushed changes. I also used black to format the code, hope the code quality checks pass this time.

adityabharadwaj198:
> Hello @adityabharadwaj198 , thanks for submitting!
> 
> Tests [fail for the moment](https://github.com/radis/radis/pull/83/checks?check_run_id=549318457). Don't worry about the Code Quality Checks, we'll fix that at the end. But you should have a look at the "Test and Coverage" one: seems there is a syntax error !
> 
> Fix and re-push!

Hey there, I fixed the 'Test and Coverage' test, it was indeed a syntax error.

radis/test/spectrum/test_compare.py
adityabharadwaj198: > in the test can you also try some other normalization options? Like, "max", "area", etc.

I'm a bit confused about that, I've tried doing setting `normalization_how` manually, and it works, but how can I write a method to test that? Since `normalization_how` is an argument of the get_residual method, and that method is called from `plot_diff` method, and there's no `normalization_how` argument in that.

erwanp:
Well done with the improvements @adityabharadwaj198 

- [ ] Looking at the [coverage report](https://codecov.io/gh/radis/radis/pull/83/changes#D1-353) I realized there are some options not covered by your current changes: around L353 :
```
if isinstance(normalize, tuple):
```
we also have some calls to `.max()` and `.mean()` that could be made nan-proof

- [ ] Also we do not handle the `normalize_how == "area"` case right now. Can you add it in your test first (it should crash) and then decide on a fix. 
In #82 is suggested to use [nantrapz](https://radis.readthedocs.io/en/latest/source/radis.misc.arrays.html#radis.misc.arrays.nantrapz)   (the description looks wrong btw, you may want to fix that!)  

Last point is a bigger suggestion:

- [ ] I realized that the same problem (nan in a Spectrum) will arise when we want to plot a single, normalized spectrum. For instance with:

```
s.plot(normalize=True)
```
We may want to use that PR to improve that aswell. In that case I would suggest to:

- [ ] add in your test a line to plot a single spectrum that has nans (as above)
- [ ] update the [spectrum.py](https://github.com/radis/radis/blob/develop/radis/spectrum/spectrum.py#L1590) accordingly. As you'll see there are two cases: one where we normalize with the max: easy fix, let's just use `nanmax`. 
- [ ] in the other option the Spectrum is normalized by using the mean value on a given range (it's very useful in data analysis). By looking at the code it seems that the function uses [norm](https://github.com/radis/radis/blob/9cec50124f7b6fac845920710f8686de6d658dbb/radis/misc/arrays.py#L27) which itself handles `nan` properly. So we should be fine. But you should add a test 
```
s.plot(normalize=(2200, 2250))  # for instance 
```

radis/test/spectrum/test_compare.py
erwanp: You're right! So far I'd include a `get_residual(s)` test line in the test file! 

@minouHub do we want to have more normalization options in plot_diff ? ("normalize=" could be given the "area" keywords, or a tuple to normalize on a certain range, like in [plot](https://radis.readthedocs.io/en/latest/source/radis.spectrum.spectrum.html#radis.spectrum.spectrum.Spectrum.plot) ). In any case that would be a separate Issue!

adityabharadwaj198:
> if isinstance(normalize, tuple):
Hey @erwanp , so I was trying to change the max and mean method calls here also, and from what I understand, to check if it's working I should be calling `get_residual` like this `get_residual(s, s*1.2, var="radiance_noslit", ignore_nan=True, normalize=(2200, 2250), normalize_how="max")
` So a tuple is passed as an argument and` if isinstance(normalize, tuple):` conditional branch can be evaluated. However when I'm doing this, I'm getting an error: _ValueError: zero-size array to reduction operation fmax which has no identity_.. Any suggestions?

radis/test/spectrum/test_compare.py
minouHub: Well, I've never used that in post-processing, so I'm not sure it is necessary.

minouHub:
I agree with @erwanp that normalizing with a tupple is very useful. I did it last week, but I did not realize there was an option for that!!!

As @adityabharadwaj198, I have some problems using this option. Here is a simple code that does not work:
```
                  molecule='CO',
                  isotope='1,2,3',
                  pressure=1.01325,   # bar
                  Tgas=700,           # K
                  mole_fraction=0.1,
                  path_length=1,      # cm
                  )
s2 = calc_spectrum(1900, 2300,         # cm-1
                  molecule='CO',
                  isotope='1,2,3',
                  pressure=1.01325,   # bar
                  Tgas=1000,           # K
                  mole_fraction=0.1,
                  path_length=1,      # cm
                  )
test = get_residual(s, s2, 'radiance_noslit', normalize=(2000, 2100))
``` 
The problem comes from the units: 'cm-1' are converted to 'nm' at some points. Therefore, the line 356 in compare.py `b = (w1 > wmin) & (w1 < wmax)` gives only False. This is why the array is 'zero-size'. 
This option should be more secure and the help re-written. I can spend more time on this, but after the 7th :)

adityabharadwaj198:
Okay, I'll look into other suggestions @erwanp made.

minouHub:
I tried to add my test to your branch using this: https://tighten.co/blog/adding-commits-to-a-pull-request
Can you "allow edits for maintainer" please. @erwanp, am I maintainer? ^^

adityabharadwaj198:
>allow edits for maintainer

@minouHub , I checked and it is enabled

adityabharadwaj198:
> * [ ]  Also we do not handle the `normalize_how == "area"` case right now. Can you add it in your test first (it should crash) and then decide on a fix.
>   In #82 is suggested to use [nantrapz]

Fixed this. I had to use `s1.get()` first and then use `nantrapz`, and I skipped using `get_integral` because of the way it is written, I hope that's okay

> s.plot(normalize=True)
> * [ ]  add in your test a line to plot a single spectrum that has nans (as above)
Added this.
> * [ ]  update the [spectrum.py](https://github.com/radis/radis/blob/develop/radis/spectrum/spectrum.py#L1590) accordingly. As you'll see there are two cases: one where we normalize with the max: easy fix, let's just use `nanmax`. 

Yes, just replacing` y.max() `with `y.nanmax()` worked

> * [ ]  in the other option the Spectrum is normalized by using the mean value on a given range (it's very useful in data analysis). By looking at the code it seems that the function uses [norm](https://github.com/radis/radis/blob/9cec50124f7b6fac845920710f8686de6d658dbb/radis/misc/arrays.py#L27) which itself handles `nan` properly. So we should be fine. But you should add a test
>> ```
> s.plot(normalize=(2200, 2250))  # for instance 
> ```

 As you said, this worked without any problems. And I added the above as a test.

erwanp:
> 
> 
> I agree with @erwanp that normalizing with a tupple is very useful. I did it last week, but I did not realize there was an option for that!!!
> 
> As @adityabharadwaj198, I have some problems using this option. Here is a simple code that does not work:
> 
> ```
> s = calc_spectrum(1900, 2300,         # cm-1
>                   molecule='CO',
>                   isotope='1,2,3',
>                   pressure=1.01325,   # bar
>                   Tgas=700,           # K
>                   mole_fraction=0.1,
>                   path_length=1,      # cm
>                   )
> s2 = calc_spectrum(1900, 2300,         # cm-1
>                   molecule='CO',
>                   isotope='1,2,3',
>                   pressure=1.01325,   # bar
>                   Tgas=1000,           # K
>                   mole_fraction=0.1,
>                   path_length=1,      # cm
>                   )
> test = get_residual(s, s2, 'radiance_noslit', normalize=(2000, 2100))
> ```
> 
> The problem comes from the units: 'cm-1' are converted to 'nm' at some points. Therefore, the line 356 in compare.py `b = (w1 > wmin) & (w1 < wmax)` gives only False. This is why the array is 'zero-size'.
> This option should be more secure and the help re-written. I can spend more time on this, but after the 7th :)

Thanks for reporting, I opened a different Issue : #85 . We'll remember to activate @adityabharadwaj198  once that Issue is fixed. 

Also, you are now officially a Maintainer :)

erwanp:
> 
> 
> > * [x]   Also we do not handle the `normalize_how == "area"` case right now. Can you add it in your test first (it should crash) and then decide on a fix.
> >   In #82 is suggested to use [nantrapz]
> 
> Fixed this. I had to use `s1.get()` first and then use `nantrapz`, and I skipped using `get_integral` because of the way it is written, I hope that's okay
> 
> > s.plot(normalize=True)
> > 
> > * [x]   add in your test a line to plot a single spectrum that has nans (as above)
> >   Added this.
> > * [x]   update the [spectrum.py](https://github.com/radis/radis/blob/develop/radis/spectrum/spectrum.py#L1590) accordingly. As you'll see there are two cases: one where we normalize with the max: easy fix, let's just use `nanmax`.
> 
> Yes, just replacing`y.max()`with `y.nanmax()` worked
> 
> > * [x]   in the other option the Spectrum is normalized by using the mean value on a given range (it's very useful in data analysis). By looking at the code it seems that the function uses [norm](https://github.com/radis/radis/blob/9cec50124f7b6fac845920710f8686de6d658dbb/radis/misc/arrays.py#L27) which itself handles `nan` properly. So we should be fine. But you should add a test
> > 
> > ```
> > s.plot(normalize=(2200, 2250))  # for instance 
> > ```
> 
> As you said, this worked without any problems. And I added the above as a test.

Well done ! I approved all changes. I'll let @minouHub approve too and then we can merge!

adityabharadwaj198:
Thanks @erwanp and @minouHub !

Issue #84, opened 2020-03-31
Database too slow - save Spectrum as H5/binary


minouHub:
Hi guys,

## Expected Behavior
I would expect the loading/writing time of the spectrum in a database to be at most of the order of magnitude of the lines calculation.

## Steps to Reproduce 
```
from radis import calc_spectrum, SpecDatabase
from time import time

t0 = time()
s = calc_spectrum(1900, 2300,         # cm-1
                  molecule='CO',
                  isotope='1,2,3',
                  pressure=1.01325,   # bar
                  Tgas=700,           # K
                  mole_fraction=0.1,
                  path_length=1,      # cm
                  verbose=False
                  )
s.apply_slit(0.5, 'nm')       # simulate an experimental slit
t1 = time()
print('------------')
print('Calculate Spectrum: {} s'.format(t1 - t0))
print('------------')


db = SpecDatabase("databaseSpectra/", nJobs=1)

t0 = time()
db.add(s, compress=True, store_name='s1', if_exists_then='replace')
t1 = time()
print('------------')
print('Save compressed: {} s'.format(t1 - t0))
print('------------')



t0 = time()
db2 = SpecDatabase("databaseSpectra/", nJobs=1)
t1 = time()
print('------------')
print('Load database: {} s'.format(t1 - t0))
print('------------')
```
Here is my output:
```
Using cache file: /home/minou/.astropy/cache/astroquery/Hitran/tempfile_CO_1_1895.00_2305.01.h5
Using cache file: /home/minou/.astropy/cache/astroquery/Hitran/tempfile_CO_2_1895.00_2305.01.h5
Using cache file: /home/minou/.astropy/cache/astroquery/Hitran/tempfile_CO_3_1895.00_2305.01.h5
/home/minou/Python/radis/radis/misc/warning.py:242: MissingSelfBroadeningWarning: Self-broadening temperature coefficient Tdpsel not given in database: used Tdpair instead
  warnings.warn(WarningType(message))
WARNING:root:json-tricks: numpy scalar serialization is experimental and may work differently in future versions
------------
Calculate Spectrum: 0.46891236305236816 s
------------
Database  initialised in databaseSpectra
*** Loading the database with 1 processor (0 files)***
Spectrum stored in databaseSpectra/s1.spec (3.0Mb)
loaded s1.spec
------------
Save compressed: 10.109835624694824 s
------------
Loading database 
*** Loading the database with 1 processor (1 files)***
loaded s1.spec
------------
[...]

PR #86, opened 2020-04-05
#84 Improve json storage size and performance

mverleg:
Don't know if anyone is following the json-tricks issue so I'll repost here in response to #84:

* Compact format was released in json-tricks v3.15.0. It saves about 20% over the old compressed mode, it's also substantially faster.
* Using ignore_comments=False where possible makes loading noticeably faster.


erwanp:
Thank you, that made things twice as fast in the example of #84! I fixed the codestyle, merging ASAP!

PR #88, opened 2020-04-06
Fixed the units' issue, added functionality to work with Nans and added a test case for issue #85

### LINKED ISSUE #85: Error when normalizing on a range
_Originally posted by @minouHub in https://github.com/radis/radis/pull/83#issuecomment-609021448_ : 

------

I agree with @erwanp that normalizing with a tupple is very useful. I did it last week, but I did not realize there was an option for that!!!

As @adityabharadwaj198, I have some problems using this option. Here is a simple code that does not work:
```
                  molecule='CO',
                  isotope='1,2,3',
                  pressure=1.01325,   # bar
                  Tgas=700,           # K
                  mole_fraction=0.1,
                  path_length=1,      # cm
                  )
s2 = calc_spectrum(1900, 2300,         # cm-1
                  molecule='CO',
                  isotope='1,2,3',
                  pressure=1.01325,   # bar
                  Tgas=1000,           # K
                  mole_fraction=0.1,
                  path_length=1,      # cm
                  )
test = get_residual(s, s2, 'radiance_noslit', normalize=(2000, 2100))
``` 
The problem comes from the units: 'cm-1' are converted to 'nm' at some points. Therefore, the line 356 in compare.py `b = (w1 > wmin) & (w1 < wmax)` gives only False. This is why the array is 'zero-size'. 
This option should be more secure and the help re-written. I can spend more time on this, but after the 7th :)

adityabharadwaj198:
[

### Description

This pull request is to address ... the issue here [#85 ](https://github.com/radis/radis/issues/85). 

- I have changed how the get function is called from inside get_residual, by adding an extra argument of wunit = s1.get_waveunit(), this fixes the problem of conversion of units from 'cm-1' to 'nm'

- Post that I added nanmax, nanmean and nantrapz which was to be done in #82 but couldn't be done because of the above problem.

- I also added a testcase to test both of the above changes.   

Fixes #85

### DIFF
--- a/radis/spectrum/compare.py
+++ b/radis/spectrum/compare.py
@@ -352,17 +352,16 @@ def get_residual(
 
         if isinstance(normalize, tuple):
             wmin, wmax = normalize
-            w1, I1 = s1.get(var, copy=False)  # (faster not to copy)
+            w1, I1 = s1.get(
+                var, copy=False, wunit=s1.get_waveunit()
+            )  # (faster not to copy)
             b = (w1 > wmin) & (w1 < wmax)
             if normalize_how == "max":
-                norm1 = I1[b].max()
-                # norm1 = np.nanmax(I1[b])
+                norm1 = np.nanmax(I1[b])
             elif normalize_how == "mean":
-                # norm1 = I1[b].mean()
-                # norm1 = np.nanmean()
-                norm1 = I1[b].nanmean()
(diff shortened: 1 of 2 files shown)


adityabharadwaj198:
> All good! I simply merged all your testing functions into one so we only calculate one Spectrum, shared for all tests (will reduce the test times)

Thanks!

PR #92, opened 2020-04-09
fixes #90

### LINKED ISSUE #90: Regenerate levels file automatically if manually changed
## Expected Behavior

Any modification to a (text) `.levels` file should regenerate the `.h5` cache. 

## Current Behavior

For pre-tabulated data, metadata currently include the molecule & format of the level database. Ex:
```
{'molecule': 'CO2', 'viblvl_label': 'pcJN', 'levelsfmt': 'cdsd-hamil', 'isotope': 1}
```
For levels calculated by Radis, it also stores the spectroscopic data used to generate the levels (and therefore will properly regenerate if the spectroscopic constants changed). 

However, it does not include the last modification time. So it won't see if the file was manually changed by the user after creation. 

## Possible Solution

Include Last modification time of the .levels in the metadata. 

- In [PartFuncCDSD_calc](https://github.com/radis/radis/blob/develop/radis/levels/partfunc_cdsd.py#L274)

- or make it a required argument of [load_h5_cache_file](https://github.com/radis/radis/blob/develop/radis/levels/partfunc_cdsd.py#L293)

Also don't forget to properly save it when creating the cache file:

- [ ] update [save_to_hdf](https://github.com/radis/radis/blob/develop/radis/levels/partfunc_cdsd.py#L310)

- [ ] and have a look at all occurences of `save_to_hdf` in the code where we could generalize the addition of this new metadata

adityabharadwaj198:
### Description

This pull request is to address ... issue #90 
I've added a variable last_modification that saves the last time .levels file was modified into the metadata. If this value is deperecated (caused by someone changing the .levels file manually) the cache file (.levels.h5) will be regenerated.

Fixes #90


erwanp:
Great! 

I see only one problem in your test:
when you modify the file, you actually add a white space 
```
with open(getTestFile(r"co2_cdsd_hamiltonian_fragment.levels"), "a") as file:
        file.write(" ")
```
Because that file is in Git history, it will appear as a change everytime we run a test. We don't want that. 

Is is possible to change the last modification time without actually changing the file? (for instance, what happens if you add nothing `file.write("")` ? Or add then delete?

adityabharadwaj198:
> what happens if you add nothing `file.write("")` ?

This is what I tried initially, and it doesn't change the modified time 

> Or add then delete?
Could you elaborate this?

erwanp:
What if you rename it? Or use `shutil.copy2` to replace it by a copy? 

Also, I'm realizing the change of the modification time may be OS dependant. Automatic testing are only run on Linux at the moment (windows is still to be implemented #9). In the meantime I'll probably run the test locally on my Windows machine.

radis/levels/partfunc_cdsd.py
```diff
@@ -272,9 +274,17 @@ def __init__(
         self.use_cached = use_cached
         self.levelsfmt = levelsfmt
         self.viblvl_label = viblvl_label
+        self.last_modification = time.ctime(
+            getmtime(getTestFile(r"co2_cdsd_hamiltonian_fragment.levels"))
+        )
+        print("Here is the last mod {0}".format(self.last_modification))
```
erwanp, line 280: this should be removed in the final version. (or hide it behind a `if verbose>=2` or something

radis/levels/partfunc_cdsd.py
```diff
@@ -286,9 +296,11 @@ def __init__(
         ]
         # (dev) locals() automatically stores all variables: levelsfmt, viblvl_label, etc.
         metadata = filter_metadata(locals(), discard_variables=_discard)
+        # print("self variables {0}".format(self))
```
erwanp, line 299: you can remove this too

radis/levels/partfunc_cdsd.py
erwanp, line 280: this should be removed in the final version. (or hide it behind a `if verbose>=2` )

radis/levels/partfunc_cdsd.py
adityabharadwaj198: Done.

radis/levels/partfunc_cdsd.py
adityabharadwaj198: Hid it.

PR #95, opened 2020-04-12
Added new tests for miscellaneous array functions

pkj-m:
### Description

This PR adds new tests for the array functions defined in `/radis/misc/arrays.py`.

### DIFF
--- a/radis/test/misc/test_arrays.py
+++ b/radis/test/misc/test_arrays.py
@@ -12,8 +12,6 @@ from radis.misc.arrays import (
     is_sorted_backward,
     find_first,
     bining,
-    shift_array,
-    find_nearest,
     calc_diff,
     autoturn,
     centered_diff,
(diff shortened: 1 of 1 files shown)


jayanthchandra:
Thanks for this @pkj-m ! There are unused imports like `find_nearest` and `shift_array`. I think you could remove this.

pkj-m:
Oh right. I was actually going to write the tests for those as well in the beginning but I got confused with a function's behavior so I skipped it. I have already asked about it in Gitter so if someone could clarify it there, I can push the tests for that function in this PR itself. Otherwise, I'll get rid of the imports and the comments.

erwanp:
@pkj-m  : 
- see [Gitter](https://gitter.im/radis-radiation/community?at=5e944bacfea5216d6976c4cf) for answer on `find_nearest`. 
- `shift_array` has no use case right now nor in a foreseable future. We should remove it entirely from the library

PR #74, opened 2020-02-13
Allow to calculate multiple molecules directly

ulyssed:
Fixes #63

Issue #63, opened 2020-02-03
Calculate multiple molecules directly


erwanp:
#  Problem

Current version of [calc_spectrum](https://radis.readthedocs.io/en/latest/source/radis.lbl.calc.html#radis.lbl.calc.calc_spectrum) is limited to one molecule only. 
We have to call it several times then use [MergeSlabs](https://radis.readthedocs.io/en/latest/los/los.html)

# Suggestion

- Allow calc_spectrum to take a dictionary as input for `mole_fraction=` 
- if a dictionary is given, then `calc_spectrum` calls himself several times, and returns a `MergeSlabs` directly 

(easy modification, great feature. Requires a test, an update to docstrings and examples in docs)

- [x] Implement a version version
- [ ] Choose syntax for user interface 
- [ ] Implement syntax

-------------

# Syntax for the Interface

Which syntax to implement in PR #74   ?

*(suggest your own if it seams more user-friendly / logical / prettier)*

### 1. Current version, where we could also allow default values

```
calc_spectrum(
    molecule={'CO2':"mole_fraction":0.8,  'CO':"mole_fraction":0.2}, 
    isotope=1,
    ... )
```
Would work. Then the default databank (fetch)  is used for everyone 

If default value check that it is the same for everyone :
```
calc_spectrum(
    molecule={'CO2':"isotope":1,  'CO':isotope=2}, 
    isotope=1, 
    ... )
```
Should fail 

### 2. Simpler syntax 

```
calc_spectrum(
    molecule={'CO2':0.8 ,  'CO':0.2 },
    ... )
```
Could work? 

Or 

### 3. Directly from the keys:

```
calc_spectrum(
    mole_fraction={'CO2':0.8 ,  'CO':0.2 }, 
    ... )
```
```
calc_spectrum(
    mole_fraction={'CO2': 0.8, 'CO':0.2}, 
    isotope={'CO2':[1,2], 'CO':[1,2,3]},
     ... )
```

### Vote!  (comment to add your own)

[![](https://api.gh-polls.com/poll/01E11RWDWA19X63PP62S3GSKD0/Syntax%201)](https://api.gh-polls.com/poll/01E11RWDWA19X63PP62S3GSKD0/Syntax%201/vote)
[![](https://api.gh-polls.com/poll/01E11RWDWA19X63PP62S3GSKD0/Syntax%202)](https://api.gh-polls.com/poll/01E11RWDWA19X63PP62S3GSKD0/Syntax%202/vote)
[...]

Issue #57, opened 2020-01-13
Allow arbitrary units in Spectrum methods


erwanp:
SpectrumFactory now read arbitrary `astropy.units` inputs after #55 

We need to do the same thing for the Spectrum post-processing methods, for instance `rescale_path_length`, `crop`, etc.   See all the methods there:  https://radis.readthedocs.io/en/latest/spectrum/spectrum.html

Most of the work will be in the [spectrum.py](https://github.com/radis/radis/blob/master/radis/spectrum/spectrum.py) file. 
- [ ] parse inputs with the `convert_and_strip_units` function as in #55 
- [ ] update examples in the docs

pkj-m:
@erwanp after working on SpectrumFactory and adding support for `wmin`/`wmax` and `wunit`, I think I gained a decent understanding of how to implement arbitrary units in the Spectrum methods, so I'd like to take this up next. Are there any specific methods that you think I should prioritize or is it okay if I start taking them up sequentially?

erwanp:
Hello! I listed all the methods above. They are all quite similar in terms of complexity. As discussed in #96 you will only have a problem in choosing what to do when dimensioned inputs contradict the default unit value (for "wunit" for instance). That should be resolved in #96 first.

pkj-m:
Hey @erwanp , I had a small question. From the docs, the `get` methods functions as follows:

`w, I = s.get('transmittance_noslit', wunit='cm-1')`

should return the wavenumber and the transmittance in `cm-1` and the _default_ unit respectively. At the same time, if I try to call the method like this: 

`w, I = s.get('transmittance_noslit', wunit='m-1')`,  

it throws an error. So if I got this right, the goal of this issue is to allow us to make the above function call work and give us the wavenumber and transmittance in `m-1` and _default_ respectively? Is that correct? And this flexibility should be offered to the `Iunit` parameter's unit as well? 
Additionally, the values being returned need to stay as numpy arrays or are we supposed to change that and return astropy's Quantity array instead?

erwanp:
1. Exactly for wunit. Currently it's only reading from harcoded set of values, either ~nm or ~cm. 
With that change we'll allow other wavelengths ("cm", "µm"), wavenumbers ("m-1") or even frequency units ("GHz"). 

[...]

PR #96, opened 2020-04-12
Support for `wmin` and `wmax` parameters in Spectrum Factory

### LINKED ISSUE #56: New optional `wmin` and `wmax` parameters in SpectrumFactory
As discussed in #55 , https://github.com/radis/radis/pull/55#issuecomment-573309359 , `wavelength_min`, `wavelength_max`, `wavenumber_min`, `wavenumber_max` can be replaced with only 2 : `wmin`, `wmax` (for instance), with one third `wunit='cm-1'` argument when people don't use Astropy units in wmin, wmax directly.

The concerned file is `lbl/factory.py`

pkj-m:
### Description

This PR adds support for new parameters `wmin` and `wmax` in the SpectrumFactory. Most modifications have been made in the `base.py` file at `radis/lbl`.
New function calls can now look like:

```
calc_spectrum(1200 * u.cm, 1400 * u.cm,...)
calc_spectrum(1200 * u.cm**-1, 1400 * u.cm**-1, ...)

calc_spectrum(1200, 1400, wunit = 'nm') 
calc_spectrum(1200, 1400, wunit = '1/m')
```
while the current function calls are still supported:

```
get_waverange(wavenum_min=1200, wavenum_max=1400)
get_waverange(wavelength_min=1200 * u.cm, wavelength_max=1400 * u.cm)
```
A few things that are still up for discussion are:

1. Can we allow users to pass units with both the `wmin`/`wmax` parameters *and* `wunit`? If so, which unit takes precedence if they differ?
2. Currently, the code thows an error if calls like ` calc_spectrum(1200, 1400, ...) are made. I would like to change that to an assertion error as I personally feel the function call looks way too vague and should not be allowed.

Will fix #56 after tests, docstring changes and further discussion.

### DIFF
--- a/radis/lbl/base.py
+++ b/radis/lbl/base.py
@@ -3338,7 +3338,7 @@ def get_waverange(
     wavenum_present = wavenum_min is not None or wavenum_max is not None
     wavelength_present = wavelength_min is not None or wavelength_max is not None
     if w_present + wavenum_present + wavelength_present >= 2:
-        raise ValueError("Cannot pass more than one set of values as input")
+        raise ValueError("Cannot pass more than one set of values as input: choose either wmin/wmax (with astropy.units), wavenum_min/wavenum_max, or wavelength_min/wavelength_max")
     assert medium in ["air", "vacuum"]
 
     # user has passed valid wmin/wmax with units
--- a/radis/lbl/base.py
+++ b/radis/lbl/base.py
@@ -80,6 +80,7 @@ from radis.misc.debug import printdbg
 from radis.misc.log import printwarn
 from radis.misc.printer import printg
 from radis.misc.warning import OutOfBoundError
+from radis.misc.utils import Default
 
 # TODO: rename in get_molecule_name
 from radis.io.hitran import get_molecule, get_molecule_identifier
@@ -3239,54 +3240,116 @@ class BaseFactory(DatabankLoader):
 
 
 def get_waverange(
+    wmin=None,
+    wmax=None,
+    wunit="cm-1_by_default",
     wavenum_min=None,
     wavenum_max=None,
     wavelength_min=None,
     wavelength_max=None,
     medium="air",
 ):
-    """ Returns wavenumber based on whatever input was given: either ν_min, ν_max 
-    directly, or λ_min, λ_max  in the given propagation ``medium``.
+    represent_wavelength = False
+    represent_wavenum = False
 
-    Parameters
-    ----------
+    # user did not pass wunit
+    if wunit == "cm-1_by_default":
+        # user did not pass wmin or wmax
+        if wmin is None or wmax is None:
+            assert wmin is None and wmax is None
+        else:
+            assert wmin is not None
+            assert wmax is not None
+            # user passes wmin/wmax with unit
+            if isinstance(wmin, u.Quantity) or isinstance(wmax, u.Quantity):
+                assert wmin.unit.is_equivalent(u.m) or wmin.unit.is_equivalent(1 / u.m)
+                assert wmax.unit.is_equivalent(u.m) or wmax.unit.is_equivalent(1 / u.m)
+                assert wmin.unit.is_equivalent(wmax.unit)
+                if wmin.unit.is_equivalent(u.m):
+                    represent_wavelength = True
+                else:
+                    represent_wavenum = True
+            # user did not pass wmin/wmax with unit
+            else:
+                wunit = wunit[:-11]
+                assert not isinstance(wmin, u.Quantity)
+                assert not isinstance(wmax, u.Quantity)
+                wmin = wmin * u.Unit(wunit)
+                wmax = wmax * u.Unit(wunit)
+                represent_wavenum = True
 
-    medium: ``'air'``, ``'vacuum'``
-        propagation medium
+    # user passed wunit explicitly
+    else:
+        # user did not pass wmin/wmax
+        if wmin is None or wmax is None:
+            raise ValueError("Please enter value of wmin and wmax when passing wunit")
+        # user passes wmin/wmax along with wunit
+        else:
+            assert wmin is not None
+            assert wmax is not None
+            # wmin/wmax values come with unit
+            if isinstance(wmin, u.Quantity) or isinstance(wmax, u.Quantity):
+                assert wmin.unit.is_equivalent(u.m) or wmin.unit.is_equivalent(1 / u.m)
+                assert wmax.unit.is_equivalent(u.m) or wmax.unit.is_equivalent(1 / u.m)
+                assert wmin.unit.is_equivalent(wmax.unit)
+                # wmin and wunit are the same units
+                if wmin.unit == u.Unit(wunit):
+                    if wmin.unit.is_equivalent(u.m):
+                        represent_wavelength = True
+                    else:
+                        represent_wavenum = True
+                # wmin and wunit have different units
+                else:
+                    raise ValueError(
+                        "Conflicting units entered for wmin/wmax and wunit"
+                    )
+            # wmin/wmax do not have units
+            else:
+                wmin = wmin * u.Unit(wunit)
+                wmax = wmax * u.Unit(wunit)
+                if wmin.unit.is_equivalent(u.m):
+                    represent_wavelength = True
+                else:
+                    represent_wavenum = True
 
-    wavenum_min, wavenum_max: float, or ~astropy.units.quantity.Quantity or ``None``
-        wavenumbers
+    if represent_wavenum:
+        wmin = convert_and_strip_units(wmin, 1 / u.cm)
+        wmax = convert_and_strip_units(wmax, 1 / u.cm)
+    if represent_wavelength:
+        wmin = convert_and_strip_units(wmin, u.nm)
+        wmax = convert_and_strip_units(wmax, u.nm)
 
-    wavelength_min, wavelength_max: float, or ~astropy.units.quantity.Quantity or ``None``
-        wavelengths in given ``medium``
-
-    Returns
-    -------
-
-    wavenum_min, wavenum_max,: float
-        wavenumbers
-
-    """
-
-    # Check input
     wavelength_min = convert_and_strip_units(wavelength_min, u.nm)
     wavelength_max = convert_and_strip_units(wavelength_max, u.nm)
     wavenum_min = convert_and_strip_units(wavenum_min, 1 / u.cm)
     wavenum_max = convert_and_strip_units(wavenum_max, 1 / u.cm)
     if (
-        wavelength_min is None
+        wmin is None
+        and wmax is None
+        and wavelength_min is None
         and wavelength_max is None
         and wavenum_min is None
         and wavenum_max is None
     ):
         raise ValueError("Give wavenumber or wavelength")
-    if (wavelength_min is not None or wavelength_max is not None) and (
-        wavenum_min is not None or wavenum_max is not None
-    ):
-        raise ValueError("Cant give both wavenumber and wavelength as input")
+
+    # check if only one pair of values has been passed
+    w_present = wmin is not None or wmax is not None
+    wavenum_present = wavenum_min is not None or wavenum_max is not None
+    wavelength_present = wavelength_min is not None or wavelength_max is not None
+    if w_present + wavenum_present + wavelength_present >= 2:
+        raise ValueError("Cannot pass more than one set of values as input")
     assert medium in ["air", "vacuum"]
 
-    # Get all waveranges
+    # user has passed valid wmin/wmax with units
+    if represent_wavelength:
+        assert wmin < wmax
+        wavelength_min = wmin
+        wavelength_max = wmax
+    elif represent_wavenum:
+        assert wmin < wmax
+        wavenum_min = wmin
+        wavenum_max = wmax
 
     # ... Input is in wavelength:
     if wavenum_min is None and wavenum_max is None:
--- a/radis/lbl/base.py
+++ b/radis/lbl/base.py
@@ -3249,80 +3249,8 @@ def get_waverange(
     wavelength_max=None,
     medium="air",
 ):
-    represent_wavelength = False
-    represent_wavenum = False
-
-    # user did not pass wunit
-    if isinstance(wunit, Default):
-        wunit = wunit.value
-        # user did not pass wmin or wmax
-        if wmin is None or wmax is None:
-            assert wmin is None and wmax is None
-        else:
-            assert wmin is not None
-            assert wmax is not None
-            # user passes wmin/wmax with unit
-            if isinstance(wmin, u.Quantity) or isinstance(wmax, u.Quantity):
-                assert wmin.unit.is_equivalent(u.m) or wmin.unit.is_equivalent(1 / u.m)
-                assert wmax.unit.is_equivalent(u.m) or wmax.unit.is_equivalent(1 / u.m)
-                assert wmin.unit.is_equivalent(wmax.unit)
-                if wmin.unit.is_equivalent(u.m):
-                    represent_wavelength = True
-                else:
-                    represent_wavenum = True
-            # user did not pass wmin/wmax with unit
-            else:
-                assert not isinstance(wmin, u.Quantity)
-                assert not isinstance(wmax, u.Quantity)
-                wmin = wmin * u.Unit(wunit)
-                wmax = wmax * u.Unit(wunit)
-                represent_wavenum = True
-
-    # user passed wunit explicitly
-    else:
-        # user did not pass wmin/wmax
-        if wmin is None or wmax is None:
-            raise ValueError("Please enter value of wmin and wmax when passing wunit")
-        # user passes wmin/wmax along with wunit
-        else:
-            assert wmin is not None
-            assert wmax is not None
-            # wmin/wmax values come with unit
-            if isinstance(wmin, u.Quantity) or isinstance(wmax, u.Quantity):
-                assert wmin.unit.is_equivalent(u.m) or wmin.unit.is_equivalent(1 / u.m)
-                assert wmax.unit.is_equivalent(u.m) or wmax.unit.is_equivalent(1 / u.m)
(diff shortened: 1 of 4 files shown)


radis/lbl/base.py
```diff
@@ -3269,25 +3272,73 @@ def get_waverange(
     """
 
     # Check input
+    if (
+        isinstance(wmin, u.Quantity) or isinstance(wmin, u.Quantity)
```
pkj-m, line 3276: Fix `wmin` to `wmax`

erwanp:
Hello. I added some extra tests but I am not able to push them. Did you accept "Allow edits from Maintainers"? 

Anyway:

1. we will have the same problem when generalizing units to Spectrum methods #57 . Methods such as get() have a default "nm" argument. We don't want users to be able to use `100 / u.cm, wunit="nm"`. However, `wunit` is not `None` by default. I see the following solutions:

- **a)** `wunit` default becomes None, and we write in the Docstrings that if it is None and undimensioned units are given, we choose "nm" (for instance). > looks dangerous to me. 

- **b)** we allow dimensioned units to override `wunit`. > looks dangerous to me too. 
- **c)** i looked up [how to check whether optional functino parameter is set](https://stackoverflow.com/questions/14749328/how-to-check-whether-optional-function-parameter-is-set). The answer is no. The best idea i could find there was to create a `Default` class. For instance, we would have `Default("nm")` class. 
- **d)** alternatively, we can have our own default value, for instance a string that says "nm_by_default", and is replaced by "nm" at runtime if wmin is not dimensioned (no user confusion possible). 

Any preference? 

2. I agree that `calc_spectrum(wmin=1200, wmax=1400,...)` is vague, but it also quite efficient. I see two reasons why we cannot drop supporting this syntax at this point: 

- backward compatibility: for instance, the very first example of the documentation uses this syntax:
```
                  molecule='CO',
                  isotope='1,2,3',
                  pressure=1.01325,   # bar
                  Tgas=700,           # K
                  mole_fraction=0.1,
                  path_length=1,      # cm
                  )
s.plot('radiance')
```
- Also, your remark also applies to all of the other arguments in the function above. The unit of pressure, or path_length, are not defined neither. We don't really want to force users to write them down all the time.

However, the additional uncertaintity for wmin, wmax is that not only there is a default unit, but also it can be wavelengths or wavenumbers. I see different ways to implement the defaults:
- **a.** keep `wavenum_min, wavenum_max` as the first two arguments, instead of `wmin, wmax`. So the historical syntax would work. And at the same time we force `wmin,wmax` to be dimensioned or `wunit` to be defined.
- **b.** `wmin, wmax` are the first two arguments, but `wunit` has a default value. There will be a problem if `wmin, wmax` are dimensioned and in contradiction with `wunit` default value. This is exactly the Problem 1. We will apply the same solution as discussed there.

pkj-m:
Hi @erwanp, sorry about that. I've checked the allow maintainer edits button now. Can you try pushing your code once again?

Regarding problem 1, I do like the clarity that option (d) provides without burdening the end-user with too many inputs. However, I am not sure how it can resolve the original problem, where both `wunit` and `wmin/wmax` have user-supplied contradicting units. 

It will however, resolve the second problem. Infact, for the second problem, where neither `wmin/wmax` nor `wunit` have a user-provided value, having our own `nm-by-default` kind of setup will be a neat way of handling this situation. 

I personally do not think having `wavenum_min` before `wmin` would make sense in the long run if our ultimate goal is to phase out these parameters and operate purely using `wmin/wmax`.

I can start working on implementing the `nm_by_default` setup for `wunit`, but we still need to decide on the right way to handle problem 1.

erwanp:
The push worked, thanks!

Option (d)  can solve 1. because now we can differenciate between cases where contradictory wunit was set by the user, and was the default:
```
calc_spectrum(wmin = 100*u.nm)   # where wunit is "cm-1" by default, but not a user input.
calc_spectrum(wmin = 100*u.nm, wunit="cm-1")  # where wunit is also "cm-1", but set that the user  
```
With (d), the first one will return wunit in "cm-1_by_default" and we know the user didn't mean anything wrong. So we can safely override it with the user input "u.nm".  
2nd line: something is definitly wrong, so we raise an (appropriate!) error. 

Is that ok for you for 1. ?  

---

(d) or (c) i'm not sure yet. (c) looks good too. The "Default" is never entered by the user, but appears in the fonction signature. 
For instance [calc_spectrum](https://radis.readthedocs.io/en/latest/source/radis.lbl.calc.html#radis.lbl.calc.calc_spectrum)  would look like: 

>  calc_spectrum(wmin=None, wmax=None, wunit=Default("cm-1"), ... )

Or 

>  calc_spectrum(wmin=None, wmax=None, wunit="cm-1_by_default")

Do you wanna ask on the Chat what's best? I can forward to my Lab people to have input. We can also make a Poll here. 


---

Ok for 2.b! So we give a default value to "wunit", with either option (c) or (d).

pkj-m:
Thanks for the elaborate explanation! I understand it perfectly now. So we can go ahead with using a default value for `wunit`. Between (c) and (d), I really don't see any particular advantage of either one, so I guess we should wait for a poll to decide how to move forward with that?

erwanp:
Do you wanna do the poll? I used [Github-Polls](https://github.com/apps/polls) in #63  if you want an example

pkj-m:
## Problem
After introducing new parameters `wmin`, `wmax` and `wunit` in the SpectrumFactory (which will eventually replace `wavelength_min/max` and `wavenum_min/max` parameters completely), we face a new problem of deciding the methods' behavior when the user enters `wmin/wmax` without any accompanying units *and* does not provide `wunit` either.

For instance, a call like `calc_spectrum(1200, 1400, ...)` will lead us to the above-described problem.

## Suggestion
To resolve the problem, we can specify a default value of `wunit` which will allow us to differentiate between any contradicting function calls made by the user (and raise an appropriate error) as well as handle the case where the user does not specify a unit either with `wmin/wmax` or with `wunit` easily.  The only thing left for further discussion is how do we implement the default value of the parameter. 

## Options

- Option 1: we create a default class, such as `Default`, and then set `wunit = Default("cm-1")` in the function definition.

`calc_spectrum(wmin=None, wmax=None, wunit=Default("cm-1"), ... )`

- Option 2: we hardcode a custom string, for instance, `nm_by_default` as the default value for `wunit`, which will again allow us to identify whether the value has been passed by the user (in which case it'll be `nm`, etc. to begin with), or it is the default value (in which case we internally convert it to `nm` before processing, and also realize that the units will now be supplied with the value of `wmin/wmax` directly.

`calc_spectrum(wmin=None, wmax=None, wunit="cm-1_by_default")`

I think one small advantage for option 1 might be the fact that having a `Default` class might help keep the code cleaner in case problems similar to this (with contradicting or absence of user-specified units) arise in the future.

## Votes

/polls "Option 1" "Option 2"

erwanp:
@pkj-m i couldn't get the /polls to work, so I replaced with 2 emojis. Asking for user input now.

pkj-m:
@erwanp Now that it seems all the votes have come in, should I start working towards option 1 or should we give it some more time?

erwanp:
Let's do! 
We should create a default class shared everywhere in Radis. That could be used to check default in other pull requests too #74 

There is a `Function arguments` part in `misc/utils.py`, maybe in there?

radis/misc/utils.py
```diff
@@ -96,6 +96,15 @@ def __del__(self):
...
 
+class Default:
+    """ Returns argument `unit` with an appended `_by_default` string
+    """
+
+    def __new__(cls, unit):
+        cls.unit = unit + "_by_default"
+        return cls.unit
+
+
```
erwanp, line 107: In this version you'd still keep the "_by_default" string, which I find a bit verbosy.

What do you think of this:

```suggestion
class Default:
    """ Contains a value
    
    Examples
    --------
    
    ::
    
        a = Default("42")
        isinstance(a, Default)
        >>> True
        
        a.value
        >>> 42
    """

    def __init__(self, value):
        self.value= value

```

Which also has the advantage to work for anything (i.e. : `Default(1.mbar)` if we ever need it ) 

Could event implement the "__eq__" method so you can get "a == 42 >>> True" 

And then, when entering the function we can evaluate all the Default arguments as `a = a.value`

pkj-m:
@erwanp I've introduced a new `Default` class according to the discussion we had, and also rewrote the `get_waverange` function to handle all the possible syntax that the user might use while calling the method.  Let me know if you think I should implement it differently.

Also, I had made a jupyter notebook consisting of all the different ways in which get_waverange function could be called to test my method. If you like, I can share the link to the notebook and we could use that to come up with new and more robust tests for this new implementation!

radis/misc/utils.py
pkj-m: Ah I see. The main reason why I implemented it with the `__new__` magic method was to avoid 
> when entering the function we can evaluate all the Default arguments as a = a.value

But I think you're right about the multi-functionality that your implementation has to offer! I'll change it right away.

pkj-m:
@erwanp I have made all the changes you suggested. If this looks good to you, I can make the docstring changes. Also, let me know what you think about [these tests ](https://github.com/pkj-m/radis-lab/blob/master/Testing%20the%20new%20get_waverange%20method.ipynb) for this PR. If they look good to you then I can add them to the test suite as well.

erwanp:
Hello! I've looked at the test suite and they're very complete! 

To test the ones that should raise an error, you can use:
```
with pytest.raises(ValueError):
    # code that should raise ValueError  for this Test to pass
```

radis/lbl/base.py
```diff
@@ -3239,54 +3240,116 @@ def plot_populations(self, what="vib", isotope=None, nfig=None):
...
-        wavenum_min is not None or wavenum_max is not None
-    ):
-        raise ValueError("Cant give both wavenumber and wavelength as input")
+
+    # check if only one pair of values has been passed
+    w_present = wmin is not None or wmax is not None
+    wavenum_present = wavenum_min is not None or wavenum_max is not None
+    wavelength_present = wavelength_min is not None or wavelength_max is not None
+    if w_present + wavenum_present + wavelength_present >= 2:
+        raise ValueError("Cannot pass more than one set of values as input")
```
erwanp, line 3341:
```suggestion
        raise ValueError("Cannot pass more than one set of values as input: choose either wmin/wmax (with astropy.units), wavenum_min/wavenum_max, or wavelength_min/wavelength_max")
```

radis/lbl/base.py
```diff
@@ -3239,54 +3240,116 @@ def plot_populations(self, what="vib", isotope=None, nfig=None):
...
+                    represent_wavelength = True
+                else:
+                    represent_wavenum = True
+            # user did not pass wmin/wmax with unit
+            else:
+                assert not isinstance(wmin, u.Quantity)
+                assert not isinstance(wmax, u.Quantity)
+                wmin = wmin * u.Unit(wunit)
+                wmax = wmax * u.Unit(wunit)
+                represent_wavenum = True
```
erwanp, line 3279: it works, but are you sure we cannot simplify all these if/else?

pkj-m: I think there must be a few simplifications we can make. I was under the impression that keeping the code structure like this would make it easier to follow and debug, especially now that we have quite a few ways of calling the `get_waverange` function.

radis/lbl/base.py
erwanp: @pkj-m you can commit this directly btw.

radis/lbl/base.py
erwanp: We are starting with an `if` conditional on `wunit`, although it will be irrelevant if wmin, wmax are not given. 

In general, we are mixing (in the new implementation, but the old one too!) unit conversions to base units (nm / cm-1), final conversion to cm-1 (which requires to check the propagation medium: air or vacuum), and tests of input consistency (_anything with assertion tests, or raise errors_). It's a little hard to understand.

What about refactoring to check these sequentially:
```
# 1. Check input consistency
- Check that at least a pair of inputs is given 
- Check that not too types of inputs are given at the same time
- If wavenum or wavelengh, or wmin is dimensionned, check that wunit is consistent (if not default)
- [we can maybe leave the check that Medium is given in the wavelength case for later]

# 2. Conversion to base units (strip dimensioned units, convert to nm or cm )
# Now no need to assert / raise anything. 
wmin, wmax given
 --> if not dimensioned: convert to nm or cm-1 based on wunit  (no need to assert, we did it in 1.). Then write wavelength_min/max , or wavenum_min/max
 --> if dimensioned, convert to nm, or cm-1. Then write wavelength_min/max, or wavenum_min/max
wavenum_min, wavenum_max given
 --> convert to cm-1
wavelength_min, wavelength_max given:
--> convert to nm

# now we only have wavenum_min/max, or wavelen_min/max, in standard units (cm-1, or nm). Last thing to check: converting nm > cm-1 depends on the propagation medium (air/vacuum)
# 3. Convert to cm-1 
if wavenum_min/max: 
--> nothing to do 
if wavelength_min/max
--> check that medium is given
--> convert to cm-1 with nm2cm or nm_air2cm  (handy function in convert.py) 

Return wavenum_min/max
```

pkj-m: Got it! I was a little confused and started thinking along the lines of simplifying the code from the logic point of view. I'll start working on this right away!

erwanp: Any suggestion welcome, it's more of a design than a coding job ! :)

radis/test/lbl/test_base.py
```diff
@@ -400,12 +402,161 @@ def test_optically_thick_limit_2iso(verbose=True, plot=True, *args, **kwargs):
         radis.DEBUG_MODE = DEBUG_MODE
 
 
+def test_get_waverange(*args, **kwargs):
```
erwanp, line 405: Great test suite!

I would also report here the comments of your [lab notebook](https://github.com/pkj-m/radis-lab/blob/master/lbl/Testing%20the%20new%20get_waverange%20method.ipynb), adding the "expected behavior". That will greatly improve the future readability of your test suite, if someone ever needs to update it:

Ex:
```
# wunit is none 
# ... wmin/wmax is none
# ...... wavenumber is passed  > should return the same wavenumbers

# ...... wavelength is passed > should convert and return wavenumbers

# ...... both wavelengths & wavenumbers are passed > should raise an Error

File: radis/lbl/base.py [hunk 6]
```diff
@@ -3239,64 +3240,108 @@ def plot_populations(self, what="vib", isotope=None, nfig=None):
...
-
-    wavelength_min, wavelength_max: float, or ~astropy.units.quantity.Quantity or ``None``
-        wavelengths in given ``medium``
-
-    Returns
-    -------
-
-    wavenum_min, wavenum_max,: float
-        wavenumbers
-
```
erwanp, line 3268: Keep the docstrings here 

(also add wmin, wmax, and wunit)

radis/lbl/base.py
```diff
@@ -3239,64 +3240,108 @@ def plot_populations(self, what="vib", isotope=None, nfig=None):
...
-    """
-
-    # Check input
```
erwanp, line 3271: Here, can you also keep a comment to explain what's happening ?  Like:
```
# 1. Checking consistency of all inputs 
```

radis/lbl/base.py
```diff
@@ -3239,64 +3240,108 @@ def plot_populations(self, what="vib", isotope=None, nfig=None):
...
+                wavenum_max = wmax * u.Unit(wunit.value)
+            else:
+                if u.Unit(wunit).is_equivalent(u.m):
+                    wavelength_min = wmin * u.Unit(wunit)
+                    wavelength_max = wmax * u.Unit(wunit)
+                else:
+                    wavenum_min = wmin * u.Unit(wunit)
+                    wavenum_max = wmax * u.Unit(wunit)
+
+    if wavenum_min is not None or wavenum_max is not None:
```
erwanp, line 3337: Normally we already checked that if one is given the other one too. This can simply be:
```suggestion
    if wavenum_min:
```

radis/lbl/base.py
```diff
@@ -3239,64 +3240,108 @@ def plot_populations(self, what="vib", isotope=None, nfig=None):
...
+                wavenum_min = wmin * u.Unit(wunit.value)
+                wavenum_max = wmax * u.Unit(wunit.value)
+            else:
+                if u.Unit(wunit).is_equivalent(u.m):
+                    wavelength_min = wmin * u.Unit(wunit)
+                    wavelength_max = wmax * u.Unit(wunit)
+                else:
+                    wavenum_min = wmin * u.Unit(wunit)
+                    wavenum_max = wmax * u.Unit(wunit)
+
```
erwanp, line 3336: Comment needed:
```suggestion
    # We now have wavenum_min/max, or wavelength_min/max defined. Let's convert these to cm-1 (warning: propagating medium is required if we start from wavelengths!)
```

radis/lbl/base.py
```diff
@@ -3239,64 +3240,108 @@ def plot_populations(self, what="vib", isotope=None, nfig=None):
...
+        wavenum_min = convert_and_strip_units(wavenum_min, 1 / u.cm)
+        wavenum_max = convert_and_strip_units(wavenum_max, 1 / u.cm)
+
+    if wavelength_min is not None or wavelength_max is not None:
+        assert medium in ["air", "vacuum"]
+        wavelength_min = convert_and_strip_units(wavelength_min, u.nm)
+        wavelength_max = convert_and_strip_units(wavelength_max, u.nm)
         if medium == "air":
             wavelength_min_vac = air2vacuum(wavelength_min)
             wavelength_max_vac = air2vacuum(wavelength_max)
```
erwanp, line 3347: Or, if you find it nicer (no need to create this new _vac variables) :

```suggestion
        if medium == "air":
            wavenum_min = nm_air2cm(wavelength_max)
            wavenum_max = nm_air2cm(wavelength_min)
        else: # medium == 'vacuum':
            wavenum_min = nm2cm(wavelength_max)
            wavenum_max = nm2cm(wavelength_min)
```

pkj-m: That definitely looks much cleaner! Thanks :D

PR #97, opened 2020-04-13
PR to address #59 (Read all datafiles)

### LINKED ISSUE #59: Read all datafiles in a folder
When loading a line database, [SpectrumFactory.load_databank](https://radis.readthedocs.io/en/latest/source/radis.lbl.loader.html#radis.lbl.loader.DatabankLoader.load_databank) can currently read a file, or a list of file. We should allow it to read an entire folder, or (better) paths with pattern matching. 

Examples:

```
path=PATH_TO_FOLDER/cdsd_*
path=PATH_TO_FOLDER/*
path=PATH_TO_FOLDER/*.par
```

Modifications would have to be done in the [lbl/loader.py](https://github.com/radis/radis/blob/develop/radis/lbl/loader.py) file.   For instance line [1283](https://github.com/radis/radis/blob/develop/radis/lbl/loader.py#L1283)

We could use the [fnmatch](https://docs.python.org/2/library/fnmatch.html#module-fnmatch) library 

- [ ] Implement the pattern
- [ ] Ensures it works as an argument in load_databank
- [ ] Ensures it works for paths given in the ~/.radis config file
- [ ] Add tests
- [ ] Add examples in documentation

adityabharadwaj198:
### Description

This pull request is to address ... reading all files in a folder, using regular expression.
example: path = PATH_TO_FOLDER/*.par

will read all the files having .par extension

Changes made in this pr work for all the regular expressions, they are case insensitive.

Regular expressions like `*co2*.par` will match files containing CO2 and co2 in their names.

Fixes #59

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -5,6 +5,7 @@ include radis/phys/units.txt
 include radis/test/files/*.txt
 include radis/test/files/*.spec
 include radis/test/files/*.par
+include radis/test/files/*.levels
 include radis/db/molparam.txt
 include radis/config.json
 recursive-include radis/db *.json


radis/misc/utils.py
```diff
@@ -12,9 +12,11 @@
 
 import os
 import sys
-from os.path import dirname
+from os.path import dirname, basename
 import importlib
 import inspect
+import fnmatch
+from radis.test.utils import getTestFile
```
erwanp, line 19: i don't think you need this one here, and i actually believe it creates a circular reference that breaks things

radis/test/lbl/test_loader.py
```diff
@@ -86,9 +87,37 @@ def test_retrieve_from_database(
         rmtree(temp_database_name)
 
 
+def test_files_from_regex(plot=True, verbose=True, warnings=True, *args, **kwargs):
```
erwanp, line 90: good to have a full test, but you should also have a small test in `misc/test_utils.py` [new file probably] that imports an tests the `get_files_from_regex()` function alone

radis/misc/utils.py
adityabharadwaj198: Understood and corrected, thanks!

radis/test/lbl/test_loader.py
adityabharadwaj198: Added the test in a new file.

radis/misc/utils.py
erwanp: imports can be a tedious thing!

radis/test/misc/test_utils.py
```diff
@@ -0,0 +1,23 @@
+from __future__ import print_function, absolute_import, division, unicode_literals
+import pytest
+
+
+def test_files_from_regex(*args, **kwargs):
+    from radis.misc.utils import get_files_from_regex
+    from radis.test.utils import getTestFile
+
+    path = "/home/aditya/radis/radis/test/files/*.par"
```
erwanp: this should use getTestFile ?

radis/lbl/loader.py
```diff
@@ -1287,6 +1288,9 @@ def _check_database_params(
         # Check input types are correct
         if isinstance(path, string_types):  # make it a list
             path = get_files_from_regex(path)
+            for i in range(len(path)):
```
erwanp, line 1291: you shouldn't use getTestFile (or any test related function) in the main package. Users may have stored their database files somewhere else.

radis/test/misc/test_utils.py
```diff
@@ -0,0 +1,23 @@
...
+
+    path = get_files_from_regex(path)
+
+    for files in path:
+        print(getTestFile(files))
+
```
erwanp, line 16: what are you testing here? You should add assertions.  For instance that you correctly found all the *.par files of the test files.

radis/test/misc/test_utils.py
adityabharadwaj198: This is just so that I can quickly generate a path to the .par files (as an test/example) anywhere this code runs.

radis/lbl/loader.py
adityabharadwaj198: Made changes like you suggested.

radis/test/misc/test_utils.py
adityabharadwaj198: Added asserts

radis/lbl/loader.py
```diff
@@ -104,6 +104,9 @@
 import gc
 from uuid import uuid1
 from six.moves import range
+import fnmatch
+from radis.misc.utils import get_files_from_regex
+from radis.test.utils import getTestFile
```
erwanp, line 109: unecessary import here (getTestFile)

radis/test/misc/test_utils.py
```diff
@@ -0,0 +1,67 @@
...
+    ]
+    path_actual.sort()
+
+    for i in range(
+        len(path_test)
+    ):  # finding basename because the function `get_file_from_regex` returns the absolute path
+        path_test[i] = basename(path_test[i])
+
+    path_test.sort()
+    assert all([x == y for x, y in zip(path_test, path_actual)])
```
erwanp, line 35: Although it's nice to have a single-line code, having a single assert for all test cases can make it harder to debug if the assertion test fails.  Can you do that in the loop above?

radis/test/misc/test_utils.py
```diff
@@ -0,0 +1,67 @@
...
+
+    # path2 is for all the files with regular expression cdsd_*.par
+    path2 = getTestFile("cdsd_*")
```
erwanp, line 38: can you change the test to "cdsd_*.txt" ?  (just because we want to exclude the .h5 cache files)

radis/test/misc/test_utils.py
adityabharadwaj198: I have another assertion below this one, for the second test case.

radis/lbl/loader.py
adityabharadwaj198: Removed.

radis/test/misc/test_utils.py
adityabharadwaj198: Done.

radis/test/misc/test_utils.py
adityabharadwaj198: Understood and done!

radis/test/misc/test_utils.py
```diff
@@ -0,0 +1,63 @@
...
+    # This list will have to be updated if more files matching regular expression cdsd_* are added/removed in/from the folder
+    path_actual2 = [
+        "hitran_CO2_fragment.par",
+        "hitran_co2_626_bandhead_4165_4200nm.par",
+    ]
+    path_actual2.sort()
+    path_test2.sort()
+
+    for i in range(
+        len(path_test2)
```
erwanp, line 49: you should loop over `path_actual2`. Imagine `path_test2` is empty: your test wouldn't fail.

erwanp, line 49: you should loop over `path_actual2`. Imagine `path_test2` is empty: your test wouldn't fail. 

Same in path_actual1 above.

radis/test/misc/test_utils.py
```diff
@@ -0,0 +1,63 @@
...
+    path_actual.sort()
+    path_test.sort()
+
+    for i in range(
+        len(path_test)
+    ):  # finding basename because the function `get_file_from_regex` returns the absolute path
+        path_test[i] = basename(path_test[i])
+        assert path_test[i] == path_actual[i]
+
+    # path2 is for all the files with regular expression cdsd_*.par
```
erwanp, line 35: update the comment. Also, it's a nice information that explains what the function is doing: you should copy that in the function docstring

radis/test/misc/test_utils.py
```diff
@@ -0,0 +1,63 @@
...
+from os import listdir
+
+# Test to compare basenames of files whose names match the regular expression passed.
+def test_files_from_regex(*args, **kwargs):
+    from radis.misc.utils import get_files_from_regex
+    from radis.test.utils import getTestFile
+    from os.path import dirname, basename
+    import fnmatch
+
+    # path1 is for finding all the files with .par extension having regular expression *.par
```
erwanp, line 13: (see my comment below: you should copy that in the function docstring)

radis/misc/arrays.py
```diff
@@ -119,7 +119,7 @@ def array_allclose(a, b, rtol=1e-5, atol=1e-8, equal_nan=True):
 
 
 def nantrapz(I, w, dx=1.0, axis=-1):
-    """ Returns np.nan(I, w) discarding nan """
+    """ Returns np.trapz(I, w) discarding nan """
```
erwanp, line 122: 👍

radis/test/misc/test_utils.py
```diff
@@ -1,14 +1,14 @@
...
-from os.path import dirname, basename, join, expanduser
+from os.path import dirname, basename, join, expanduser, abspath
 from os import listdir
 import fnmatch
 from radis.misc.utils import get_files_from_regex
 
 
 # Test to compare basenames of files whose names match the regular expression passed.
-FOLDER_PATH_EXAMPLE = expanduser("~/radis/radis/test/files")
+TEST_FOLDER_PATH = dirname(abspath(dirname(__file__)))
```
erwanp, line 11: That works, but you can use getTestFile in case we ever move test_utils.py, or the test files folder.

```
TEST_FILES_PATH = dirname(getTestFile("."))
```
(this returns something like "~radis/radis/test/files/")

erwanp:
@CorentinGrimaldi you may be interested in the new feature.

radis/test/misc/test_utils.py
adityabharadwaj198: I think getTestFile will throw an error because, before returning the file path, it's existance is being checked.

erwanp: In the latest version yes, so you cannot use getTestFile("*.par"), but you can use getTestFile(".") which will return the folder name (which exists!)
At least it works on my local Windows version.

erwanp:
Ok we're good for the content!

As mentioned on Gitter we have some cleaning to do on the Git History:
Right now I can see many Merge: https://github.com/radis/radis/pull/97/commits
If we merge as is, it will produce a hardly readable history. 

You should Rebase instead, i.e. change the order of your new commits to put them on top. 

At this point, maybe the easiest thing to do is to:

- save your changes somewhere
- re-clone the project from https://github.com/radis/radis to start from the latest version. 
- re-apply your changes in 1 or a few commits. 
- force-push to this PR (so to `adityabharadwaj198/radis`) . 

Maybe @ritzvik can help in how to clean a history. I'm not very experienced! :)

ritzvik:
As I can see that there are only 5 changed files, I would say copy them somewhere and preserve them. Other methods are possible but risky and might lose the progress. The complexity is not worth it.
Then we can reset the current branch by : 
```
git reset --hard HEAD~30 # going back some commits behind
git remote add upstream https://github.com/radis/radis.git
git pull --rebase upstream develop
```
Upto this point the branch is even with `develop`. Then files can be manually copied back and committed. Changes would need to be force pushed `git push origin adityabharadwaj198/radis -f`

Personally I think git skills are something which one picks up along the way. Still, in this case, the other way would have been `git stash`. One may read up about this.

ritzvik:
@erwanp We have a small concise guide to refactoring commits https://github.com/einsteinpy/einsteinpy/wiki/How-to-refactor-commits 

Please feel free to refer people to this or replicate it.

Issue #62, opened 2020-01-17
Improve test coverage


erwanp:
17/01/20 : [Test coverage](https://codecov.io/gh/radis/radis/tree/develop/radis) is about 70%. 

Edit : 17/02/21 : 78%, getting there! 💪 

Edit :: 12/11/21 : down to 75% :( 

Edit : 21/03/22 : up to 77%

- [ ] Goal:: should be > 80%. 

[![coverage](https://codecov.io/gh/radis/radis/branch/develop/graph/badge.svg)](https://codecov.io/gh/radis/radis/tree/develop/radis)

That's a good issue to learn about the package!

Identify modules that need the more coverage :
- [ ] for instance a MergeSlabs with `intersect`:  https://codecov.io/gh/radis/radis/pull/91/src/radis/los/slabs.py#L464

- [ ] in base.py there are the [_add_EvibErot_CDSD_pc](https://codecov.io/gh/radis/radis/src/64d449dee4e71a95b6b0e1bd6cacdd5760fd9b0f/radis/lbl/base.py) and [_add_Evib123Erot_CDSD_pc](https://codecov.io/gh/radis/radis/src/64d449dee4e71a95b6b0e1bd6cacdd5760fd9b0f/radis/lbl/base.py#L820) function (advanced calculations for nonequilibrium CO2 in "p,c,j,n" nomenclature ) . See how the similar [_add_EvibErot_CDSD_pcN](https://codecov.io/gh/radis/radis/src/64d449dee4e71a95b6b0e1bd6cacdd5760fd9b0f/radis/lbl/base.py#L587) function was tested in #99 

Clean deprecated parts of the code ([ask on Gitter](https://gitter.im/radis-radiation/community) if needed)

Add new tests as needed

- [x] for instance move `_test()` from `misc/signals.py` in the test folder. And run it. 

- [x] maybe exclude some parts from the coverage: see https://coverage.readthedocs.io/en/v4.5.x/excluding.html#advanced-exclusion

erwanp:
Added some room for more coverage with #91 by adding files to test the different CO2 nonequilibrium models.

sagarchotalia:
**Some of the current functions in the code that aren't being covered:**

In [radis/db](https://codecov.io/gh/radis/radis/tree/develop/radis/db): 
- In [classes.py](https://app.codecov.io/gh/radis/radis/blob/develop/radis/db/classes.py)
[...]

PR #98, opened 2020-04-14
Refactoring some misc array functions

pkj-m:
### Description

This PR cleans up the `radis/misc/array.py` file by:

1. reimplementing the `find_nearest` function: the function now behaves differently from the previous implementation in the following ways:

-  if multiple elements in the `searched` array have identical nearest elements in `array`, they are repeated as many times for each element in the output list, e.g.

ORIGINAL: `find_nearest(np.array([1,5, 10]), np.array([-5, 2, 0])`  returned `array([1])`
MODIFIED: `find_nearest(np.array([1,5, 10]), np.array([-5, 2, 0])`  returns `array([1,1,1])`

-  the index of elements in the output list match the index of the element they are nearest to in the input list, e.g.

ORIGINAL: `find_nearest(np.array([1,5,10]), np.array([12,0,6])`  returned `array([1,5,10])`
MODIFIED: `find_nearest(np.array([1,5,10]), np.array([12,0,6])`  returns `array([10,1,5])`

- Removed the `argmax` parameter. 
- Made the boolean array an optional output given it's lack of utility. It'll now be returned as an output only when the new parameter `return_bool` (default: `False`) is set to `True`.
- Added new tests for the function

2. Removed the `shift_array` function after the discussion on #95 ([comment](https://github.com/radis/radis/pull/95#issuecomment-612865352))

### DIFF
--- a/radis/misc/__init__.py
+++ b/radis/misc/__init__.py
@@ -14,7 +14,6 @@ from .arrays import (
     norm,
     norm_on,
     scale_to,
-    shift_array,
     calc_diff,
     find_nearest,
     find_first,
--- a/radis/misc/arrays.py
+++ b/radis/misc/arrays.py
@@ -189,15 +189,26 @@ def find_nearest(array, searched, return_bool=False):
     ::
 
         from numpy import array
-        find_nearest(array([1,2,3,4]), array([2.1,2]))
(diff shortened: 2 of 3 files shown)


erwanp:
Nice, contributes to improving coverage #62 too!

PR #100, opened 2020-04-15
(minor) Database - 'compress = 2' 

minouHub:
Hi guys,

The (fabulous) option `compress = 2` when saving a spectrum in a database is now documented and tested. The test consist in generating a spectrum, save it, load it, and compare it to the generated spectrum.

Minou

### DIFF
--- a/radis/test/tools/test_database.py
+++ b/radis/test/tools/test_database.py
@@ -117,16 +117,15 @@ def test_plot_spec(plot=True, close_plots=True, verbose=True, *args, **kwargs):
 
 def test_save_compressed2():
     "Check if saving a spectrum with compress = 2 does not change something."
-    from radis import calc_spectrum, SpecDatabase
     import shutil
 
-    # s1 : cm-1, vacuum
-    # s2 : nm, air
+    from radis.test.utils import setup_test_line_databases
+    from radis import calc_spectrum, SpecDatabase
+
+    setup_test_line_databases()
     s = calc_spectrum(
-        1900,
-        1950,  # cm-1
-        # wavelength_min=4160,   # nm
-        # wavelength_max=4190,   # nm
+        2000,
+        2300,  # cm-1
         molecule="CO",
         isotope="1,2,3",
         pressure=1.01325,  # bar
@@ -136,17 +135,30 @@ def test_save_compressed2():
         verbose=False,
         wstep=0.01,
         medium="vacuum",
+        databank="HITRAN-CO-TEST",
     )
 
+    # load in one databse
     db = SpecDatabase(dirname(getTestFile(".")) + "/newDb/")
-    db.add(s, compress=1, if_exists_then="error")
+    db.add(s, compress=2, if_exists_then="error")
 
-    # s_bis = db.get_closest(Tgas=700)
-    s_bis = db.get_unique(Tgas=700)
+    # simulate an experimentalist who come later and load the spectrum
+    db2 = SpecDatabase(dirname(getTestFile(".")) + "/newDb/")
+    s_bis = db2.get_unique(Tgas=700)
 
+    # we clean the folders for next times
     shutil.rmtree(dirname(getTestFile(".")) + "/newDb/")
-    assert s.compare_with(s_bis)
-    assert s == s_bis
+
+    # we check the loaded spectrum contains less information than the calculated one
+    assert not s == s_bis
+    s_bis.update()
+
+    # now we check if it works
+    assert s.compare_with(s_bis, spectra_only=True, plot=False)
+
+    # assert s == s_bis
+    # print(s)
+    # print(s_bis)
 
 
 if __name__ == "__main__":
--- a/radis/test/tools/test_database.py
+++ b/radis/test/tools/test_database.py
@@ -122,11 +122,7 @@ def test_save_compressed2(verbose=True, *args, **kwargs):
     from radis.test.utils import setup_test_line_databases
     from radis import calc_spectrum, SpecDatabase
 
-    try:
-        # we want to make sure this folder is not generated from before
-        shutil.rmtree(dirname(getTestFile(".")) + "/newDb/")
-    except:
-        pass
+    shutil.rmtree(join(dirname(getTestFile(".")),newDb"), ignore_errors=True)
     
     try:
         #get the spectrum
--- a/radis/test/tools/test_database.py
+++ b/radis/test/tools/test_database.py
@@ -123,7 +123,6 @@ def test_save_compressed2(verbose=True, *args, **kwargs):
     from radis.test.utils import setup_test_line_databases
     from radis import calc_spectrum, SpecDatabase
 
-
     shutil.rmtree(join(dirname(getTestFile(".")), "newDb"), ignore_errors=True)
 
     try:
--- a/radis/test/tools/test_database.py
+++ b/radis/test/tools/test_database.py
@@ -114,56 +114,44 @@ def test_plot_spec(plot=True, close_plots=True, verbose=True, *args, **kwargs):
     if plot:
         plot_spec(getTestFile("N2C_specair_380nm.spec"))
 
+
 def test_save_compressed2():
+    "Check if saving a spectrum with compress = 2 does not change something."
     from radis import calc_spectrum, SpecDatabase
     import shutil
-    #s1 : cm-1, vacuum
-    #s2 : nm, air
-    s1 = calc_spectrum(
-                1900, 1950,         # cm-1
-                # wavelength_min=4160,   # nm
-                # wavelength_max=4190,   # nm
-                  molecule='CO',
-                  isotope='1,2,3',
-                  pressure=1.01325,   # bar
-                  Tgas=700,           # K
-                  mole_fraction=0.1,
-                  path_length=1,      # cm
-                  verbose=False,
-                  wstep=0.01,
-                  medium='vacuum',
-                  )
-    s2 = calc_spectrum(
-                1900, 1950,         # cm-1
-                # wavelength_min=4160,   # nm
-                # wavelength_max=4190,   # nm
-                  molecule='CO',
-                  isotope='1,2,3',
-                  pressure=1.01325,   # bar
-                  Tgas=700,           # K
-                  mole_fraction=0.1,
-                  path_length=1,      # cm
-                  verbose=False,
-                  wstep=0.01,
-                  medium='vacuum',
-                  )
-    
-    db = SpecDatabase(dirname(getTestFile("."))+'/newDb/')
-    for s in [s1, s2]:
-        db.add(s, compress=2, if_exists_then='replace')
-    
-    # s1_bis = db.get_unique(Tgas=700)
-    # s2_bis = db.get_unique(Tgas=1000)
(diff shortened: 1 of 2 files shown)


radis/test/tools/test_database.py
```diff
@@ -115,17 +115,43 @@ def test_plot_spec(plot=True, close_plots=True, verbose=True, *args, **kwargs):
         plot_spec(getTestFile("N2C_specair_380nm.spec"))
 
 
-def _run_testcases(plot=True, close_plots=False, verbose=True, *args, **kwargs):
-
-    test_speclist(*args, **kwargs)
-    test_database_functions(
-        plot=plot, close_plots=close_plots, verbose=verbose, *args, **kwargs
+def test_save_compressed2():
```
erwanp, line 118: if using this example, Pytest will require an internet connection to download the database. We usually mark these with
```@pytest.mark.needs_connection``` to skip these tests if there is no connection. 

However, here you could calculate a spectrum with the Test Line databases directly. The code below would work too, and save the connection:
```
from radis.test.utils import setup_test_line_databases

setup_test_line_databases()
s = calc_spectrum(
    2000,
    2300,  # cm-1
    molecule="CO",
    isotope="1,2,3",
    pressure=1.01325,  # bar
    Tgas=700,  # K
    mole_fraction=0.1,
    path_length=1,  # cm
    verbose=False,
    wstep=0.01,
    medium="vacuum",
    databank="HITRAN-CO-TEST")
```

File: radis/test/tools/test_database.py [hunk 2]
```diff
@@ -115,17 +115,43 @@ def test_plot_spec(plot=True, close_plots=True, verbose=True, *args, **kwargs):
...
+        path_length=1,  # cm
+        verbose=False,
+        wstep=0.01,
+        medium="vacuum",
     )
-    test_plot_spec(plot=plot, close_plots=close_plots, verbose=verbose, *args, **kwargs)
 
-    return True
+    db = SpecDatabase(dirname(getTestFile(".")) + "/newDb/")
+    db.add(s, compress=1, if_exists_then="error")
```
erwanp, line 142: shouldn't this be compress=2?

radis/test/tools/test_database.py
```diff
@@ -115,17 +115,43 @@ def test_plot_spec(plot=True, close_plots=True, verbose=True, *args, **kwargs):
...
+
+    # s_bis = db.get_closest(Tgas=700)
+    s_bis = db.get_unique(Tgas=700)
```
erwanp, line 145: I disagree: here we're not reading it from the disk, but directly from the SpecDatabase. In the past spectra were stored on disk & reloaded before they were added in the database, but since #84 we store them directly, no? 

So you can:
- generate a new Database db2. Same place. 
- or simply use `s.store()` and `s=load_spec(...)`. Do we really need the SpecDatabase there?

radis/test/tools/test_database.py
```diff
@@ -115,17 +115,43 @@ def test_plot_spec(plot=True, close_plots=True, verbose=True, *args, **kwargs):
...
+
+    shutil.rmtree(dirname(getTestFile(".")) + "/newDb/")
+    assert s.compare_with(s_bis)
```
erwanp, line 148: with compress==2, these should actually fail. You'll have lost all of Lines, and the redundant spectral quantites. You may want to `compare_with(... var='raidance_noslit')` for instance, and use `s.update('radiance_noslit')` to recalculate it if missing after loading.   

Actually, you should even `assert` that they are different. 
And also assert than some redundant quantites were indeed removed.

radis/tools/database.py
```diff
@@ -2015,11 +2015,14 @@ def add(self, spectrum, store_name=None, **kwargs):
             Other :py:meth:`~radis.spectrum.spectrum.Spectrum.store` parameters can be given 
             as kwargs arguments:
 
-        compress: boolean
-            if ``True``, removes all quantities that can be regenerated with ``s.update()``,
+        compress: boolean or int
```
erwanp, line 2018: or: `compress = 0, 1, 2` to it's more clear we can go up to 2 ?

radis/test/tools/test_database.py
minouHub: yes ^^

radis/test/tools/test_database.py
minouHub: alright, I'll add first an assert to check they are different (stuff have been removed)

Then, I want to get the identical spectrum. So I also need an 2nd assert after an update, and it doesn't work :(

radis/test/tools/test_database.py
```diff
@@ -115,17 +115,55 @@ def test_plot_spec(plot=True, close_plots=True, verbose=True, *args, **kwargs):
...
+
+    # simulate an experimentalist who come later and load the spectrum
+    db2 = SpecDatabase(dirname(getTestFile(".")) + "/newDb/")
+    s_bis = db2.get_unique(Tgas=700)
+
+    # we clean the folders for next times
+    shutil.rmtree(dirname(getTestFile(".")) + "/newDb/")
+
+    # we check the loaded spectrum contains less information than the calculated one
+    assert not s == s_bis
```
erwanp, line 153: Could also add:
```suggestion
    assert not s == s_bis
    assert s_bis.get_vars() == ['abscoeff']     # only this spectral quantity was stored
    assert s_bis.lines is None
    assert s_bis.conditions is not None  # we kept the metadata
```

radis/test/tools/test_database.py
erwanp: So when you updated(), you actually created more quantities that were there initially (like the emissivity() . You can see if if running `verbose=True`. I made some changes to compare only the initial quantities.

radis/test/tools/test_database.py
minouHub: That's great !!!

radis/test/tools/test_database.py
minouHub: Good changes 👍

radis/test/tools/test_database.py
```diff
@@ -115,17 +115,67 @@ def test_plot_spec(plot=True, close_plots=True, verbose=True, *args, **kwargs):
...
+        db.add(s, compress=2, if_exists_then="error")
+    
+        # simulate an experimentalist who come later and load the spectrum
+        db2 = SpecDatabase(dirname(getTestFile(".")) + "/newDb/")
+        s_bis = db2.get_unique(Tgas=700)
+        
+        # we clean the folders for next times
+        shutil.rmtree(dirname(getTestFile(".")) + "/newDb/")
+    
+    except:
```
erwanp, line 160:
```suggestion
    finally:
```

Very different from `except`

Also, you should start the `try` on line 149: we shouldn't try to clean the folders before they were created

radis/test/tools/test_database.py
```diff
@@ -115,17 +115,67 @@ def test_plot_spec(plot=True, close_plots=True, verbose=True, *args, **kwargs):
...
-    test_plot_spec(plot=plot, close_plots=close_plots, verbose=verbose, *args, **kwargs)
+    from radis.test.utils import setup_test_line_databases
+    from radis import calc_spectrum, SpecDatabase
 
-    return True
+    try:
+        # we want to make sure this folder is not generated from before
+        shutil.rmtree(dirname(getTestFile(".")) + "/newDb/")
+    except:
+        pass
```
erwanp, line 129:
```suggestion
    shutil.rmtree(join(dirname(getTestFile(".")),newDb"), ignore_errors=True)
```

radis/test/tools/test_database.py
```diff
@@ -115,17 +115,67 @@ def test_plot_spec(plot=True, close_plots=True, verbose=True, *args, **kwargs):
...
+            mole_fraction=0.1,
+            path_length=1,  # cm
+            verbose=False,
+            wstep=0.01,
+            medium="vacuum",
+            databank="HITRAN-CO-TEST",
+        )
 
+        # load in one databse
+        db = SpecDatabase(dirname(getTestFile(".")) + "/newDb/")
```
erwanp, line 150:
```suggestion
        db = SpecDatabase(join(dirname(getTestFile(".")),"newDb"))
```

radis/test/tools/test_database.py
```diff
@@ -115,17 +115,67 @@ def test_plot_spec(plot=True, close_plots=True, verbose=True, *args, **kwargs):
...
+        db.add(s, compress=2, if_exists_then="error")
+    
+        # simulate an experimentalist who come later and load the spectrum
+        db2 = SpecDatabase(dirname(getTestFile(".")) + "/newDb/")
```
erwanp, line 154:
```suggestion
        db2 = SpecDatabase(join(dirname(getTestFile(".")),"newDb"))
```

radis/test/tools/test_database.py
```diff
@@ -115,17 +115,67 @@ def test_plot_spec(plot=True, close_plots=True, verbose=True, *args, **kwargs):
...
+        s_bis = db2.get_unique(Tgas=700)
+        
+        # we clean the folders for next times
+        shutil.rmtree(dirname(getTestFile(".")) + "/newDb/")
+    
+    except:
+        # we want to make sure this folder is deleted
+        shutil.rmtree(dirname(getTestFile(".")) + "/newDb/")
```
erwanp, line 162:
```suggestion
    finally:
        # we want to make sure this folder is deleted
        shutil.rmtree(join(dirname(getTestFile(".")),"newDb"))
```

radis/test/tools/test_database.py
```diff
@@ -122,11 +122,7 @@ def test_save_compressed2(verbose=True, *args, **kwargs):
     from radis.test.utils import setup_test_line_databases
     from radis import calc_spectrum, SpecDatabase
 
-    try:
-        # we want to make sure this folder is not generated from before
-        shutil.rmtree(dirname(getTestFile(".")) + "/newDb/")
-    except:
-        pass
+    shutil.rmtree(join(dirname(getTestFile(".")),newDb"), ignore_errors=True)
```
erwanp:
```suggestion
    shutil.rmtree(join(dirname(getTestFile(".")),"newDb"), ignore_errors=True)
```

minouHub: already done

radis/test/tools/test_database.py
```diff
@@ -115,17 +115,60 @@ def test_plot_spec(plot=True, close_plots=True, verbose=True, *args, **kwargs):
...
+            isotope="1,2,3",
+            pressure=1.01325,  # bar
+            Tgas=700,  # K
+            mole_fraction=0.1,
+            path_length=1,  # cm
+            verbose=False,
+            wstep=0.01,
+            medium="vacuum",
+            databank="HITRAN-CO-TEST",
+        )
```
erwanp, line 144:
```suggestion
    # get the spectrum
    setup_test_line_databases()
    s = calc_spectrum(
        2000,
        2300,  # cm-1
        molecule="CO",
        isotope="1,2,3",
        pressure=1.01325,  # bar
        Tgas=700,  # K
        mole_fraction=0.1,
        path_length=1,  # cm
        verbose=False,
        wstep=0.01,
        medium="vacuum",
        databank="HITRAN-CO-TEST",
    )
    try:
```

PR #101, opened 2020-04-15
(MAJOR) quick fix to the unit problem in "get_residual"

minouHub:
Guys,

There is a BIG unit problem in the function `get_residual`. If the two function do not have the same unit, then the residual won't be correct. See the example below:

**Code:**
```
from radis import calc_spectrum, get_residual, experimental_spectrum
s1 = calc_spectrum(1900, 2300,
                  molecule='CO',
                  isotope='1,2,3',
                  pressure=1.01325,
                  Tgas=1000,
                  mole_fraction=0.1,
                  )
s1.apply_slit(1, 'nm')

#s2 is like s1 but at 1001 K
s2 = calc_spectrum(1900, 2300,         
                  molecule='CO',
                  isotope='1,2,3',
                  pressure=1.01325,
                  Tgas=1001, #HERE IS THE DIFFERENCE
                  mole_fraction=0.1,
                  )
s2.apply_slit(1, 'nm')
residual_ref = get_residual(s1, s2, 'radiance', ignore_nan=False, normalize=True)


w1, I1 = s1.get('radiance', copy=False, wunit=s1.get_waveunit())
#Fake experimental spectrum (identical)
s_expe_1 = experimental_spectrum(w1, I1, Iunit="mW/cm2/sr/nm", wunit="cm-1") 
#Fake experimental spectrum with a new unit (but identical)
s_expe_2 = experimental_spectrum(w1, I1*1e-3, Iunit="W/cm2/sr/nm", wunit="cm-1") 
residual1 = get_residual(s1, s_expe_1, 'radiance', ignore_nan=False, normalize=True)
residual2 = get_residual(s1, s_expe_2, 'radiance', ignore_nan=False, normalize=True)

print(residual1, residual2, residual_ref)
```

**Old Output:**
`0.0 0.8176423186518765 5.532303126258176e-07`

1. The first one is expected
2. The second one is the error !
3. The third is just to give you an idea what change will be induced by a increase of 1 K

With my fix, the difference is reduced. I'll open an issue about the remaining difference in a minute.

**New Output:**
`0.0 0.0008176423186518765 5.532303126258176e-07`

### DIFF
--- a/radis/spectrum/operations.py
+++ b/radis/spectrum/operations.py
@@ -429,7 +429,7 @@ def _get_unique_var(s, var, inplace):
     return var
 
 
-def multiply(s, coef, var=None, inplace=False):
+def multiply(s, coef, Iunit=None, var=None, inplace=False):
     """Multiply s[var] by the float 'coef'
 
     Parameters
@@ -454,6 +454,8 @@ def multiply(s, coef, var=None, inplace=False):
     """
     # Check input
     var = _get_unique_var(s, var, inplace)
+    if Iunit == None:
+        Iunit = s.units[var]
 
     if not inplace:
         s = s.copy(quantity=var)
@@ -461,7 +463,7 @@ def multiply(s, coef, var=None, inplace=False):
     #            s.name = name
 
     # Multiply inplace       ( @dev: we have copied already if needed )
-    w, I = s.get(var, wunit=s.get_waveunit(), copy=False)
+    w, I = s.get(var, wunit=s.get_waveunit(), copy=False, Iunit=Iunit)
     I *= coef  # @dev: updates the Spectrum directly because of copy=False
 
     return s
--- a/radis/spectrum/compare.py
+++ b/radis/spectrum/compare.py
@@ -353,13 +353,13 @@ def get_residual(
         if isinstance(normalize, tuple):
             wmin, wmax = normalize
             w1, I1 = s1.get(
-                var, copy=False, wunit=s1.get_waveunit()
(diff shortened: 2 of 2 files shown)


radis/spectrum/operations.py
```diff
@@ -454,14 +454,16 @@ def multiply(s, coef, var=None, inplace=False):
...
 
     if not inplace:
         s = s.copy(quantity=var)
     #        if name is not None:
     #            s.name = name
 
     # Multiply inplace       ( @dev: we have copied already if needed )
-    w, I = s.get(var, wunit=s.get_waveunit(), copy=False)
+    w, I = s.get(var, wunit=s.get_waveunit(), copy=False, Iunit=Iunit)
     I *= coef  # @dev: updates the Spectrum directly because of copy=False
```
erwanp, line 468: and here, we should correct the unit of the spectrum.

- use `astropy.units` for the conversion. 

Hard part:
- how do you make sure that "mW/cm2/sr/nm" * "1" does not return something in ~ "W/m3/sr" ? Which is strictly speaking, through, but we should make these "cm2" are a distance and "nm" are wavelengths, and shouldn't be mixed. [spectral-flux-density](https://docs.astropy.org/en/stable/units/equivalencies.html#spectral-flux-luminosity-density-units) may help. I've never done that. 
- [ ] add tests! 

- Note: after #57  `coef` may have a dimension already.

erwanp:
Closing, a better fix will be given by fixing #109

PR #106, opened 2020-04-16
Replaces pint with astropy.units

### LINKED ISSUE #104: replace pint with astropy.units
In #55  and #57 we're deploying astropy.units in most of the Radis functions and methods.

There are some functions, however, that already used `pint`. For consistency they should be replaced. 

Changes should be made:
- [ ] in the [misc/units.py](https://github.com/radis/radis/blob/develop/radis/phys/units.py) module 
- [ ] look elsewhere? 
- [ ] remove `pint` from dependencies.

pkj-m:
### Description

- This PR replaces `pint` with `astropy.units` in the `phys/units.py` file.
- The import has been removed from the file and replaced with astropy units everywhere. 
- The `DimensionalityError` offered by `pint` has been replaced with astropy's `UnitConversionError` everywhere.

A few tests were breaking as they depended on pint's `UnitRegistry` which was imported as `Q_` earlier to convert the strings to units. The same task (and name) has now been given to astropy's `units.Unit` class which acts as the equivalent.

### TODO: 
- [x]  clean the code and remove old comments after review
- [x] search for other methods using `pint` and replace them with `astropy.units`
- [x] remove `pint` from dependencies

Will fix #104

### DIFF
--- a/docs/spectrum/spectrum.rst
+++ b/docs/spectrum/spectrum.rst
@@ -85,7 +85,7 @@ From numpy arrays, use :py:meth:`~radis.spectrum.spectrum.Spectrum.from_array` :
     # w, T are two numpy arrays 
     from radis import Spectrum
     s = Spectrum.from_array(w, T, 'transmittance_noslit', 
-                               waveunit='nm', unit='1')
+                               waveunit='nm', unit='') # adimensioned
                                
               
 From a file, use :py:meth:`~radis.spectrum.spectrum.Spectrum.from_txt` ::
--- a/docs/spectrum/spectrum.rst
+++ b/docs/spectrum/spectrum.rst
@@ -85,7 +85,7 @@ From numpy arrays, use :py:meth:`~radis.spectrum.spectrum.Spectrum.from_array` :
     # w, T are two numpy arrays 
     from radis import Spectrum
     s = Spectrum.from_array(w, T, 'transmittance_noslit', 
-                               waveunit='nm', unit='I/I0')
+                               waveunit='nm', unit='1')
                                
               
 From a file, use :py:meth:`~radis.spectrum.spectrum.Spectrum.from_txt` ::
--- a/radis/lbl/base.py
+++ b/radis/lbl/base.py
@@ -102,7 +102,7 @@ class BaseFactory(DatabankLoader):
 
     # Output units
     units = {
-        "absorbance": "1",
+        "absorbance": "",
         "abscoeff": "cm-1",
         "abscoeff_continuum": "cm-1",
         # TODO: deal with case where 'cm-1' is given as input for a Spectrum
@@ -113,10 +113,10 @@ class BaseFactory(DatabankLoader):
         "radiance_noslit": "mW/cm2/sr/nm",  # it's actually a spectral radiance
         "emisscoeff": "mW/cm3/sr/nm",
         "emisscoeff_continuum": "mW/cm3/sr/nm",
-        "emissivity": "1",
-        "emissivity_noslit": "1",
-        "transmittance": "1",
-        "transmittance_noslit": "1",
+        "emissivity": "",
+        "emissivity_noslit": "",
+        "transmittance": "",
+        "transmittance_noslit": "",
     }
 
     # Calculation Conditions units
--- a/radis/lbl/base.py
+++ b/radis/lbl/base.py
@@ -103,8 +103,8 @@ class BaseFactory(DatabankLoader):
     # Output units
     units = {
         "absorbance": "-ln(I/I0)",
-        "abscoeff": "cm_1",
-        "abscoeff_continuum": "cm_1",
+        "abscoeff": "cm-1",
+        "abscoeff_continuum": "cm-1",
         # TODO: deal with case where 'cm-1' is given as input for a Spectrum
         # (write a cast_unit of some kind)
         # different in Specair (mw/cm2/sr) because slit
@@ -162,7 +162,7 @@ class BaseFactory(DatabankLoader):
         # dont change this without making sure your line database
         # is correct, and the units conversions (ex: radiance)
         # are changed accordingly
-        # Note that radiance are converted from ~ [mW/cm2/sr/cm_1]
+        # Note that radiance are converted from ~ [mW/cm2/sr/cm-1]
         # to ~ [mW/cm/sr/nm]
         assert self.params.waveunit == self.cond_units["wstep"]
 
--- a/radis/lbl/base.py
+++ b/radis/lbl/base.py
@@ -102,7 +102,7 @@ class BaseFactory(DatabankLoader):
 
     # Output units
     units = {
-        "absorbance": "-ln(I/I0)",
+        "absorbance": "1",
         "abscoeff": "cm-1",
         "abscoeff_continuum": "cm-1",
         # TODO: deal with case where 'cm-1' is given as input for a Spectrum
@@ -113,10 +113,10 @@ class BaseFactory(DatabankLoader):
         "radiance_noslit": "mW/cm2/sr/nm",  # it's actually a spectral radiance
         "emisscoeff": "mW/cm3/sr/nm",
         "emisscoeff_continuum": "mW/cm3/sr/nm",
-        "emissivity": "eps",
-        "emissivity_noslit": "eps",
-        "transmittance": "I/I0",
-        "transmittance_noslit": "I/I0",
+        "emissivity": "1",
+        "emissivity_noslit": "1",
+        "transmittance": "1",
+        "transmittance_noslit": "1",
     }
 
     # Calculation Conditions units
--- a/radis/phys/blackbody.py
+++ b/radis/phys/blackbody.py
@@ -27,7 +27,8 @@ from __future__ import absolute_import
 from numpy import exp, arange, ones_like, zeros_like, inf
 from radis.phys.constants import k_b, c, h
 from radis.phys.constants import k_b_CGS, c_CGS, h_CGS
-from radis.phys.units import conv2, Q_
+from radis.phys.units import conv2
+from radis.phys.units import Unit as Q_
 from radis.phys.air import air2vacuum
 
 
--- a/radis/phys/blackbody.py
+++ b/radis/phys/blackbody.py
@@ -72,7 +72,7 @@ def planck(lmbda, T, eps=1, unit="mW/sr/cm2/nm"):
     return iplanck
 
 
-def planck_wn(wavenum, T, eps=1, unit="mW/sr/cm2/cm_1"):
+def planck_wn(wavenum, T, eps=1, unit="mW/sr/cm2/cm-1"):
     """ Planck function for blackbody radiation, wavenumber version
 
 
@@ -89,13 +89,13 @@ def planck_wn(wavenum, T, eps=1, unit="mW/sr/cm2/cm_1"):
         default 1
 
     unit: str
-        output unit. Default 'mW/sr/cm2/cm_1'
+        output unit. Default 'mW/sr/cm2/cm-1'
 
 
     Returns
     -------
 
-    planck: np.array   default (mW/sr/cm2/cm_1)
+    planck: np.array   default (mW/sr/cm2/cm-1)
         equilibrium radiance
 
     """
@@ -110,8 +110,8 @@ def planck_wn(wavenum, T, eps=1, unit="mW/sr/cm2/cm_1"):
     # iplanck in erg/s/sr/cm2/cm-1
     iplanck *= 1e-4  # erg/s/sr/cm2/cm-1 > mW/sr/cm^2/cm-1
 
-    if Q_(unit) != Q_("mW/sr/cm2/cm_1"):
-        iplanck = conv2(iplanck, "mW/sr/cm2/cm_1", unit)
+    if Q_(unit) != Q_("mW/sr/cm2/cm-1"):
+        iplanck = conv2(iplanck, "mW/sr/cm2/cm-1", unit)
 
     return iplanck
 
@@ -204,7 +204,7 @@ def sPlanck(
     if waveunit == "cm-1":
         # generate the vector of wavenumbers (shape M)
         w = arange(wavenum_min, wavenum_max + wstep, wstep)
-        Iunit = "mW/sr/cm2/cm_1"
+        Iunit = "mW/sr/cm2/cm-1"
         I = planck_wn(w, T, eps=eps, unit=Iunit)
     else:
         # generate the vector of lengths (shape M)
@@ -230,8 +230,8 @@ def sPlanck(
         conditions=conditions,
         units={
             "radiance_noslit": Iunit,
-            "transmittance_noslit": "I/I0",
-            "absorbance": "-ln(I/I0)",
+            "transmittance_noslit": "1",
+            "absorbance": "1",
         },
         cond_units={"wstep": waveunit},
         waveunit=waveunit,
--- a/radis/phys/blackbody.py
+++ b/radis/phys/blackbody.py
@@ -228,11 +228,7 @@ def sPlanck(
             "absorbance": (w, ones_like(w) * inf),
         },
         conditions=conditions,
-        units={
-            "radiance_noslit": Iunit,
-            "transmittance_noslit": "1",
-            "absorbance": "1",
-        },
+        units={"radiance_noslit": Iunit, "transmittance_noslit": "", "absorbance": "",},
         cond_units={"wstep": waveunit},
         waveunit=waveunit,
         name="Planck {0}K, eps={1:.2g}".format(T, eps),
--- a/radis/phys/units.py
+++ b/radis/phys/units.py
@@ -158,7 +158,7 @@ def convert_emi2cm(j_nm, wavenum, Iunit0, Iunit):
     # But both plots look alright
 
     # Convert to whatever was wanted
-    j_cm = conv2(j_cm, "mW/cm**3/sr/cm-1", Iunit)
+    j_cm = conv2(j_cm, "mW/cm3/sr/cm-1", Iunit)
 
     return j_cm
 
@@ -209,7 +209,7 @@ def convert_emi2nm(j_cm, wavenum, Iunit0, Iunit):
     # But both plots look alright
 
     # Convert to whatever was wanted
-    j_nm = conv2(j_nm, "mW/cm**3/sr/nm", Iunit)
+    j_nm = conv2(j_nm, "mW/cm3/sr/nm", Iunit)
 
     return j_nm
 
@@ -326,7 +326,7 @@ def convert_rad2nm(l_cm, wavenum, Iunit0, Iunit):
     # But both plots look alright
 
     # Convert to whatever was wanted
-    l_nm = conv2(l_nm, "mW/cm**2/sr/nm", Iunit)
+    l_nm = conv2(l_nm, "mW/cm2/sr/nm", Iunit)
 
     return l_nm
 
@@ -336,8 +336,8 @@ def convert_universal(
     from_unit,
     to_unit,
     spec=None,
-    per_nm_is_like="mW/sr/cm**2/nm",
-    per_cm_is_like="mW/sr/cm**2/cm-1",
+    per_nm_is_like="mW/sr/cm2/nm",
+    per_cm_is_like="mW/sr/cm2/cm-1",
 ):
     """ Return variable var in whatever unit, and converts to to_unit
     Also deal with cases where var is in ~1/nm (per_nm_is_like) or ~1/cm-1
--- a/radis/phys/units.py
+++ b/radis/phys/units.py
@@ -1,24 +1,6 @@
 # -*- coding: utf-8 -*-
 """
 
-Notes
------
-
-Dimensioned calculation is slower:
-
-Performance test with pint:
-
-- 2 times slower than numpy on np.std(np.exp())
-
-Examples
---------
-
-Import radis.phys.units to get access to::
-
-    from radis.phys.units import Q_
-    a = Q_(np.array([5,4,2]),'cm')
-    a.to_base_units()
-
 -------------------------------------------------------------------------------
 
 
@@ -27,11 +9,41 @@ Import radis.phys.units to get access to::
 from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import unicode_literals
-import numpy as np
-from os.path import join, dirname, abspath, exists
-from astropy import units as u
+import astropy.units as u
+import warnings
 
-Q_ = u.Unit
+
+def Unit(st, *args, **kwargs):
+    """ Radis evaluation of an unit, using :py:class:`~astropy.units.Unit`
+    
+    Changes compare to Astropy standards:
+        
+    - "µm" is accepted and converted to "um"
+    - we do not raise a warning if multiple slashes, e.g. "mW/cm2/sr/nm"
+    
+    Examples
+    --------
+    
+    ::
+        
+        from radis.phys.units import Unit as u
+        a = 200 * u("mW/cm2/sr/nm")
+        a += 0.1 * u("W/cm2/sr/nm")
+    
+    """
+
+    try:
+        st = st.replace("µ", "u")
+    except AttributeError:
+        pass
+
+    with warnings.catch_warnings() as w:
+        # Ignore some warnings : Cause all warnings to always be triggered.
+        warnings.filterwarnings(
+            "ignore", ".*multiple slashes.*", category=u.UnitsWarning
+        )
+        unit = u.Unit(st, *args, **kwargs)
+    return unit
 
 
 def conv2(quantity, fromunit, tounit):
@@ -57,7 +69,7 @@ def conv2(quantity, fromunit, tounit):
     1.
     
     The output is still non dimensional. We don't transform `quantity` 
-    into a pint array (or neq.phys.uarray) because this may create a performance
+    into a dimensioned array because this may create a performance
     drop in computationaly-expensive task. Instead, we assume we know for 
     sure the units in which some of our quantities will be created, and just
     want to let the users choose another output unit 
@@ -66,8 +78,8 @@ def conv2(quantity, fromunit, tounit):
     """
 
     try:
-        a = quantity * u.Unit(fromunit)
-        a = a.to(u.Unit(tounit))
+        a = quantity * Unit(fromunit)
+        a = a.to(Unit(tounit))
 
     except u.UnitConversionError:
         raise TypeError(
@@ -78,7 +90,7 @@ def conv2(quantity, fromunit, tounit):
 
 
 def is_homogeneous(unit1, unit2):
-    """ Tells if unit1 and unit2 are homogeneous, using the Pint library
+    """ Tells if unit1 and unit2 are homogeneous, using the Astropy library
 
 
     Parameters    
@@ -90,7 +102,7 @@ def is_homogeneous(unit1, unit2):
     """
 
     try:
-        1 * u.Unit(unit1) + 1 * u.Unit(unit2)
+        1 * Unit(unit1) + 1 * Unit(unit2)
         return True
     except u.UnitConversionError:
         return False
--- a/radis/phys/units.py
+++ b/radis/phys/units.py
@@ -29,49 +29,9 @@ from __future__ import print_function
 from __future__ import unicode_literals
 import numpy as np
 from os.path import join, dirname, abspath, exists
-from pint import UnitRegistry, DimensionalityError
+from astropy import units as u
 
-ureg = UnitRegistry()
-_units_file = abspath(join(dirname(__file__), "units.txt"))
-# make sure file exists
-if not exists(_units_file):
-    raise AssertionError("Couldn't find units file in : {0}".format(_units_file))
-ureg.load_definitions(_units_file)
-Q_ = ureg.Quantity
-
-# %% Pint aware arrays
-
-
-class uarray(np.ndarray):
-    """ Unit-aware array based on Pint
-
-    Example
-    -------
-
-    >>> a = uarray(np.linspace(10, 100, 10), 'Td')
-
-    """
-
-    def __new__(cls, input_array, unit=None):
-        # Input array is an already formed ndarray instance
-        # We first cast to be our class type
-
-        obj = np.asarray(input_array).view(cls)
-        # add the new attribute to the created instance
-
-        if unit is not None:
-            obj = Q_(obj, unit)
-
-        # Finally, we must return the newly created object:
-        return obj
-
-    def __array_finalize__(self, obj):
-        # see InfoArray.__array_finalize__ for comments
-        if obj is None:
-            return
-
-
-#        self.info = getattr(obj, 'info', None)
+Q_ = u.Unit
 
 
 def conv2(quantity, fromunit, tounit):
@@ -102,46 +62,19 @@ def conv2(quantity, fromunit, tounit):
     sure the units in which some of our quantities will be created, and just
     want to let the users choose another output unit 
 
-    2. 
-    
-    because angles are dimensionless a 'mW/cm2/sr/nm' > 'mW/cm2/nm' conversion 
-    is considered  valid, while we expected the Luminance to be converted to 
-    an exitance/irradiance and thus multiplied by Pi !
-    Here we prevent this behavior by considering 
-        
-    
 
     """
 
     try:
-        a = Q_(quantity, fromunit)
-        a = a.to(tounit)
+        a = quantity * u.Unit(fromunit)
+        a = a.to(u.Unit(tounit))
 
-        # Hardcoded: 'pint' considers angles to be dimensionless (which they are)
-        # so a 'mW/cm2/sr/nm' > 'mW/cm2/nm' conversion is then considered  valid,
-        # while we expected the Luminance to be converted to an Exitance/Irradiance
-        # and thus multiplied by Pi !!
-        # Here we prevent this behavior:
+    except u.UnitConversionError:
+        raise TypeError(
+            "Cannot convert quantity to the specified unit. Please check the dimensions."
+        )
 
-        if "sr" in fromunit and "sr" not in tounit:
-            raise DimensionalityError(fromunit, tounit)
-        if "sr" in tounit and "sr" not in fromunit:
-            raise DimensionalityError(fromunit, tounit)
-
-    except TypeError:
-        if "cm-1" in fromunit or "cm-1" in tounit:
-            #            raise TypeError('Use cm_1 instead of cm-1 else it triggers errors in '+\
-            #                            'pint (symbolic unit converter)')
-            return conv2(
-                quantity,
-                fromunit.replace("cm-1", "cm_1"),
-                tounit.replace("cm-1", "cm_1"),
-            )
-
-        else:
-            raise
-
-    return a.magnitude
+    return a.value
 
 
 def is_homogeneous(unit1, unit2):
@@ -157,9 +90,9 @@ def is_homogeneous(unit1, unit2):
     """
 
     try:
-        Q_(unit1) + Q_(unit2)
+        1 * u.Unit(unit1) + 1 * u.Unit(unit2)
         return True
-    except DimensionalityError:
+    except u.UnitConversionError:
         return False
 
 
@@ -169,7 +102,7 @@ def is_homogeneous(unit1, unit2):
 def convert_emi2cm(j_nm, wavenum, Iunit0, Iunit):
     """
     Convert spectral emission density in wavelength base (typically ~mW/cm3/sr/nm) 
-    to spectral emission density in wavenumber base (~mW/cm3/sr/cm_1)
+    to spectral emission density in wavenumber base (~mW/cm3/sr/cm-1)
 
 
     Parameters    
@@ -207,7 +140,7 @@ def convert_emi2cm(j_nm, wavenum, Iunit0, Iunit):
     Validation:
 
         >>> w_nm, j_nm = s.get('emisscoeff', 'nm', 'mW/cm2/sr/nm')
-        >>> w_cm, j_cm = s.get('emisscoeff', 'cm', 'mW/cm2/sr/cm_1')
+        >>> w_cm, j_cm = s.get('emisscoeff', 'cm', 'mW/cm2/sr/cm-1')
         >>> print(trapz(y_nm, x_nm))
         >>> print(trapz(y_cm, x_cm))
 
@@ -215,12 +148,9 @@ def convert_emi2cm(j_nm, wavenum, Iunit0, Iunit):
 
     """
 
-    if Q_(Iunit0) != Q_(
-        "mW/cm3/sr/nm"
-    ):  # Q_ makes sure 'mW/sr/cm3/nm' == 'mW/cm3/sr/nm'
-        j_nm = conv2(j_nm, Iunit0, "mW/cm3/sr/nm")
+    j_nm = conv2(j_nm, Iunit0, "mW/cm**3/sr/nm")
 
-    # Convert ''mW/cm3/sr/nm' to 'mW/cm3/sr/cm_1'
+    # Convert ''mW/cm3/sr/nm' to 'mW/cm3/sr/cm-1'
     j_cm = j_nm * 1e7 / wavenum ** 2
     # note that we discard the - sign.
     # This means that one of the integrals `trapz(j_nm, wavelen)` or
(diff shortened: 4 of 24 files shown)


radis/test/test_phys.py
```diff
@@ -116,8 +117,9 @@ def test_units(verbose=True, *args, **kwargs):
...
     # Test unit-ware arrays
     # RADIS pint-aware array
-    a = uarray(np.linspace(10, 100, 10), "Td")
-    res = Q_(np.linspace(1e-16, 1e-15, 10), "V * cm^2")  # pint definition
+    Td = "1e-17 * volt * centimeter**2"
+    a = uarray(np.linspace(10, 100, 10), Td)
+    res = np.linspace(1e-16, 1e-15, 10) * u.Unit("V * cm^2")  # pint definition
 
     assert (
         np.round(np.array(a.to("V * cm^2")) - np.array(res), 5) == np.zeros_like(res)
```
erwanp, line 125:
```suggestion
```

these can be removed, they're more plasma physics than spectroscopy.

radis/phys/units.py
```diff
@@ -114,34 +106,29 @@ def conv2(quantity, fromunit, tounit):
     """
 
     try:
-        a = Q_(quantity, fromunit)
-        a = a.to(tounit)
+        a = quantity * u.Unit(fromunit)
+        a = a.to(u.Unit(tounit))
 
         # Hardcoded: 'pint' considers angles to be dimensionless (which they are)
```
erwanp, line 112: I think unlike pint, astropy handles Angles as a physical unit by default. So the comment could be removed. 

Add to the test something like 
```
assert mW/cm2/sr/nm not equivalent to mW/cm2/nm  ?  # pseudo-code
```

erwanp:
I added some changes above. 

Now I have a big remark:

### Non-standard units

Whatever Astropy.units may consider standard, we do not want to have to write
```
mW/cm**2/sr/nm
```
That's just very unreadable compared to
```
mW/cm2/sr/nm
```
There are a few like this:
```
cm**2 -> cm2
m**2 -> m2
cm**3 -> cm3
m**3 -> m3 
mm**3 -> mm3 
```

And of course the famous "cm-1" cherished by all infrared spectroscopists. 
`cm-1 -> cm**-1 `


Preventing the use of this syntax would certainly cut us from all our user base (@minouHub @gugzy first ). 


So the question is more: 

**How do we implement these in an _elegant_, and _safe_ way?**

- Astropy allows custom units. But setting `m3` as `m**3` may not be safe: `cm3` would be interpreted as 1/100 * m**3 and not (1/100*m)**3 
- Currently, were were parsing & replacing text. That may still work. They'll only be given as inputs. We want:
```
substract(s, 2, "mW/cm2/sr/nm")   # fast version
substract(s, 2* u.mW/u.cm**2/u.sr/u.nm)  # pure astropy version
```
but we don't really need
```
substract(s, 2* u.mW/u.cm2/u.sr/u.nm)  # if really you added all these u. you can afford u.cm**2 
``` 

Same for the "cm-1": 
```
substract(s, 2, "mW/cm2/sr/cm-1")   # would still work 
substract(s, 2* u.mW/u.cm**2/u.sr/(1/u.cm))  # astropy units version 
```
but good luck in parsing the next:
```
substract(s, 2* u.mW/u.cm**2/u.sr/u.cm-1)
``` 

Any better idea than parsing text?

pkj-m:
hey @erwanp , thanks for your detailed remarks. However, I believe that astropy does infact support units the way you would like to see in Radis.

So if we choose to handle the non-standard units in a format such as `mW/cm2/sr/nm`, it's definitely possible, as can be quickly checked with something like `10 * u.Unit("mW/cm2/sr/nm")` which returns us exactly what we would expect: 
image

This also includes writing centimeter inverse as `"cm-1"`.

Also, could you explain your point regarding the possibility of confusing `cm3` as `(1/100)*m^3` instead of `(1/100 * m)^3`. Is that directed at users who might get confused? Regardless, I think Astropy does a good job at handling these unit formats as we can see:
image
and 
image



The reason I changed the syntax from `cm3` to `cm**3` was because I felt it looked more mathematically accurate. But you're right in pointing out that it makes the function calls significantly uglier for users.

Does that resolve the problem or did I misinterpret something?

pkj-m:
Another thing that I found while going through the CI tests output was the use of `cm_1` in a lot of places in the code instead for `cm-1`. I think `pint` had a way of parsing `cm_1` as `cm-1` from the `Units.txt` file that was in the `phys` folder. However, I couldn't come up with a quicker and cleaner method to solve the problem than to simply replace all occurences of `cm_1` in the code base to `cm-1`. Is that alright or do you think we should handle this problem in a different manner?

erwanp:
> 
> Also, could you explain your point regarding the possibility of confusing `cm3` as `(1/100)*m^3` instead of `(1/100 * m)^3`. Is that directed at users who might get confused? 

I was concerned that by defining a unit such as m3 ~ m**3 the parser would ready cm3 as c(m3) instead of (cm)3. But as you found out, it turns out I was worried for no reason: 

> Regardless, I think Astropy does a good job at handling these unit formats

Replacing "cm_1" with "cm-1" is actually better than initially. "cm_1" was there because `pint` couldnt resolve `cm-1`. Astropy can, so let's keep it! 

Nice improvement!

erwanp:
The test crash on
```assert s.get_power(unit="W/cm2/sr") == 2631.6288408588148```
You should replace it with `np.isclose()` and that will do.

**Comment by pkj-m** [comment]
Hey @erwanp , I was trying to fix the spectrum tests that are failing. It seems that the `apply_slit` method calls `pint`'s `UnitRegistry.units` method which basically returns the unit in a string format which was used further ([exact line here](https://github.com/radis/radis/blob/develop/radis/spectrum/spectrum.py#L2400)).

After changing that line to do the exact same thing except with `astropy` units this time, I ran into another error:
```astropy.units.core.UnitConversionError: 'mW / (cm2 nm sr um)' and 'mW / (cm2 sr)' are not convertible``` thrown by the `conv2` method.

At some point, the test is trying to convert a quantity from one unit to another one with different dimensions.  On printing the exact unit of the quantity which is being handled in the `apply_slit` function, I found out it's `mW/cm2/sr/µm*nm`. However, I am not sure why this was being converted to `mW/cm2/sr`. 
I think I am confused about the logic behind the transformations, and even though you did explain about it above, I think I am still not clear on how to handle this particular problem.

**Comment by erwanp** [comment]
So that's what we're doing when applying the slit:
image
In terms of units, the spectrum after slit as the [initial unit] * [unit of convolution kernel Phi] * [unit of wavespace]

Now there are two ways to normalize the convolution kernel. Either by area: 

image

Either at maximum: 

image

- So in the first case we end up with no change in inputs. 
- In the second case we need to multiply the spectrum unit by [unit of wavespace]   (can be `nm`, `cm-1`, or whatever was used). 

This is what we're doing. 

---

What surprises me in your error message is we end up with `'mW / (cm2 nm sr um)' . While we should have had 'mW / (cm2 sr um) * nm'    (which is indeed homogeneous to  'mW / (cm2 sr)')

**Comment by pkj-m** [comment]
Hi @erwanp ,  thanks for the thorough explanation. I understand it much better now!

> What surprises me in your error message is we end up with 'mW / (cm2 nm sr um) . While we should have had mW / (cm2 sr um) * nm (which is indeed homogeneous to mW / (cm2 sr))

That problem got fixed by rearranging the units such so that it read `nm * mW / cm2 / sr / um`. 
However, I have run into some other problems now. This is the console output when I run the `test_slit.py` file locally after substituting all occurrences of `I/I0`, `-ln(I/I0)` and `eps` from the defaults' dictionary in the BaseFactory class, and also while loading them from the database loader:

```
...
>>> _test_all_slits yield correct FWHM (+- wstep) : OK

File CO_Tgas1500K_mole_fraction0.01.spec has a deprecrated structure (key medium removed in 0.9.22). Fixing this time, but regenerate database ASAP.
File CO_Tgas1500K_mole_fraction0.01.spec has a deprecrated structure (v1u in lines is now called vu). Fixed this time, but regenerate database ASAP.
/home/pankaj/radis/radis/spectrum/rescale.py:1967: UserWarning: Large rescaling from 0.01 to 1. There may be significant changes in pressure broadening coefficients.You should calculate a new spectrum instead.
  + "You should calculate a new spectrum instead."
WARNING: UnitsWarning: 'mW/cm2/sr/nm' contains multiple slashes, which is discouraged by the FITS standard [astropy.units.format.generic]
WARNING: UnitsWarning: 'mW/sr/cm2/nm' contains multiple slashes, which is discouraged by the FITS standard [astropy.units.format.generic]
CO_Tgas1500K_mole_fraction0.01.spec new quantities added: ['absorbance', 'transmittance_noslit', 'emissivity_noslit', 'radiance_noslit', 'emisscoeff']
compare radiance with rtol=0.001
... couldn't plot radiance
compare radiance with rtol=0.001
... couldn't plot radiance
compare radiance with rtol=0.001
/home/pankaj/radis/radis/spectrum/compare.py:861: RuntimeWarning: More than 20 figures have been opened. Figures created through the pyplot interface (`matplotlib.pyplot.figure`) are retained until explicitly closed and may consume too much memory. (To control this warning, see the rcParam `figure.max_open_warning`).
  fig = plt.figure(num=nfig, figsize=figsize)
compare radiance with rtol=0.001
/home/pankaj/radis/radis/spectrum/compare.py:861: RuntimeWarning: More than 20 figures have been opened. Figures created through the pyplot interface (`matplotlib.pyplot.figure`) are retained until explicitly closed and may consume too much memory. (To control this warning, see the rcParam `figure.max_open_warning`).
  fig = plt.figure(num=nfig, figsize=figsize)
File CO_Tgas1500K_mole_fraction0.5.spec has a deprecrated structure (key medium removed in 0.9.22). Fixing this time, but regenerate database ASAP.
File CO_Tgas1500K_mole_fraction0.5.spec has a deprecrated structure (v1u in lines is now called vu). Fixed this time, but regenerate database ASAP.
WARNING: UnitsWarning: 'mW/cm2/sr/nm' contains multiple slashes, which is discouraged by the FITS standard [astropy.units.format.generic]
WARNING: UnitsWarning: 'mW/sr/cm2/nm' contains multiple slashes, which is discouraged by the FITS standard [astropy.units.format.generic]
WARNING: UnitsWarning: 'nm*mW/cm2/sr/nm' contains multiple slashes, which is discouraged by the FITS standard [astropy.units.format.generic]
WARNING: UnitsWarning: 'mW/cm2/sr' contains multiple slashes, which is discouraged by the FITS standard [astropy.units.format.generic]
Test auto resampling
Using database: HITRAN-CO-TEST
HITRAN-CO-TEST 
-------
info : HITRAN 2016 database, CO, 3 main isotopes (CO-26, 36, 28), 2000-2300 cm-1
path : ['/home/pankaj/radis/radis/test/files/hitran_co_3iso_2000_2300cm.par']
format : hitran
parfuncfmt : hapi
levelsfmt : radis


Using cache file: /home/pankaj/radis/radis/test/files/hitran_co_3iso_2000_2300cm.h5
Calculating Equilibrium Spectrum
Physical Conditions
   Tgas                 300 K
   Trot                 300 K
   Tvib                 300 K
   isotope              1,2,3
   mole_fraction        0.02
   molecule             CO
   path_length          100 cm
   self_absorption      True
   state                X
   wavenum_max          2260.0000 cm-1
   wavenum_min          2230.0000 cm-1
Computation Parameters
   Tref                 296 K
   broadening_max_width  20 cm-1
   db_assumed_sorted    True
   dbformat             hitran
   dbpath               /home/pankaj/radis/radis/test/files/hitran_co_3iso_2000_2300cm.par
   dlm_res_G            0.01
   dlm_res_L            0.01
   levelsfmt            radis
   parfuncfmt           hapi
   pseudo_continuum_threshold  0
   wavenum_max_calc     2270.0000 cm-1
   wavenum_min_calc     2220.0000 cm-1
   waveunit             cm-1
   wstep                0.02 cm-1
Information
   db_use_cached        True
   export_lines         True
   load_energies        True
   lvl_use_cached       True
   parallel             False
WARNING: UnitsWarning: 'mW/cm2/sr/nm' contains multiple slashes, which is discouraged by the FITS standard [astropy.units.format.generic]
WARNING: UnitsWarning: 'mW/sr/cm2/nm' contains multiple slashes, which is discouraged by the FITS standard [astropy.units.format.generic]
Spectrum calculated in 0.03s
/home/pankaj/radis/radis/spectrum/spectrum.py:3493: UserWarning: Wavespace is not evenly spaced (0.001%) for transmittance_noslit. This may create problems when convolving with slit function
  + " This may create problems when convolving with slit function"
test_slit.py:705: RuntimeWarning: More than 20 figures have been opened. Figures created through the pyplot interface (`matplotlib.pyplot.figure`) are retained until explicitly closed and may consume too much memory. (To control this warning, see the rcParam `figure.max_open_warning`).
  fig = plt.figure(fig_prefix + "auto-resampling")
  File "/home/pankaj/anaconda3/envs/radis-dev/lib/python3.6/site-packages/pyparsing.py", line 1683, in _parseNoCache
    loc, tokens = self.parseImpl(instring, preloc, doActions)
  File "/home/pankaj/anaconda3/envs/radis-dev/lib/python3.6/site-packages/pyparsing.py", line 2897, in parseImpl
    if instring[loc] == self.firstMatchChar:
IndexError: string index out of range

During handling of the above exception, another exception occurred:

pyparsing.ParseException: Expected {snowflake | accent | symbol | c_over_c | function | group | frac | dfrac | stackrel | binom | genfrac | sqrt | overline | operatorname}, found end of text  (at char 18), (line:1, col:19)

During handling of the above exception, another exception occurred:

  File "/home/pankaj/anaconda3/envs/radis-dev/lib/python3.6/site-packages/matplotlib/mathtext.py", line 2581, in parse
    result = self._expression.parseString(s)
  File "/home/pankaj/anaconda3/envs/radis-dev/lib/python3.6/site-packages/pyparsing.py", line 1955, in parseString
    raise exc
  File "/home/pankaj/anaconda3/envs/radis-dev/lib/python3.6/site-packages/pyparsing.py", line 4065, in parseImpl
    raise ParseSyntaxException._from_exception(pe)
pyparsing.ParseSyntaxException: Expected {snowflake | accent | symbol | c_over_c | function | group | frac | dfrac | stackrel | binom | genfrac | sqrt | overline | operatorname}, found end of text  (at char 18), (line:1, col:19)

During handling of the above exception, another exception occurred:

  File "test_slit.py", line 824, in <module>
    print(("Testing slit.py: ", _run_testcases(plot=True)))
  File "test_slit.py", line 818, in _run_testcases
    test_resampling(plot=plot, verbose=verbose, *args, **kwargs)
  File "test_slit.py", line 714, in test_resampling
    label="(stored in cm-1)",
  File "/home/pankaj/radis/radis/spectrum/spectrum.py", line 1654, in plot
    fix_style(str("origin"))
  File "/home/pankaj/anaconda3/envs/radis-dev/lib/python3.6/site-packages/publib/main.py", line 195, in fix_style
    _fix_style(style, ax, **kwargs)
  File "/home/pankaj/anaconda3/envs/radis-dev/lib/python3.6/site-packages/publib/main.py", line 230, in _fix_style
    plt.tight_layout()
  File "/home/pankaj/anaconda3/envs/radis-dev/lib/python3.6/site-packages/matplotlib/pyplot.py", line 1337, in tight_layout
    gcf().tight_layout(pad=pad, h_pad=h_pad, w_pad=w_pad, rect=rect)
  File "/home/pankaj/anaconda3/envs/radis-dev/lib/python3.6/site-packages/matplotlib/cbook/deprecation.py", line 358, in wrapper
    return func(*args, **kwargs)
  File "/home/pankaj/anaconda3/envs/radis-dev/lib/python3.6/site-packages/matplotlib/figure.py", line 2493, in tight_layout
    pad=pad, h_pad=h_pad, w_pad=w_pad, rect=rect)
  File "/home/pankaj/anaconda3/envs/radis-dev/lib/python3.6/site-packages/matplotlib/tight_layout.py", line 360, in get_tight_layout_figure
    pad=pad, h_pad=h_pad, w_pad=w_pad)
  File "/home/pankaj/anaconda3/envs/radis-dev/lib/python3.6/site-packages/matplotlib/tight_layout.py", line 109, in auto_adjust_subplotpars
    tight_bbox_raw = union([ax.get_tightbbox(renderer) for ax in subplots
  File "/home/pankaj/anaconda3/envs/radis-dev/lib/python3.6/site-packages/matplotlib/tight_layout.py", line 110, in <listcomp>
    if ax.get_visible()])
  File "/home/pankaj/anaconda3/envs/radis-dev/lib/python3.6/site-packages/matplotlib/axes/_base.py", line 4317, in get_tightbbox
    bb_xaxis = self.xaxis.get_tightbbox(renderer)
  File "/home/pankaj/anaconda3/envs/radis-dev/lib/python3.6/site-packages/matplotlib/axis.py", line 1202, in get_tightbbox
    *ticklabelBoxes2,
  File "/home/pankaj/anaconda3/envs/radis-dev/lib/python3.6/site-packages/matplotlib/axis.py", line 1200, in <genexpr>
    if a.get_visible()),
  File "/home/pankaj/anaconda3/envs/radis-dev/lib/python3.6/site-packages/matplotlib/text.py", line 905, in get_window_extent
    bbox, info, descent = self._get_layout(self._renderer)
  File "/home/pankaj/anaconda3/envs/radis-dev/lib/python3.6/site-packages/matplotlib/text.py", line 300, in _get_layout
    clean_line, self._fontproperties, ismath=ismath)
  File "/home/pankaj/anaconda3/envs/radis-dev/lib/python3.6/site-packages/matplotlib/backends/backend_agg.py", line 209, in get_text_width_height_descent
    self.mathtext_parser.parse(s, self.dpi, prop)
  File "/home/pankaj/anaconda3/envs/radis-dev/lib/python3.6/site-packages/matplotlib/mathtext.py", line 3354, in parse
    box = self._parser.parse(s, font_output, fontsize, dpi)
  File "/home/pankaj/anaconda3/envs/radis-dev/lib/python3.6/site-packages/matplotlib/mathtext.py", line 2586, in parse
    str(err)]))
ValueError: 
^\mathregular{-I/I
                  ^
Expected {snowflake | accent | symbol | c_over_c | function | group | frac | dfrac | stackrel | binom | genfrac | sqrt | overline | operatorname}, found end of text  (at char 18), (line:1, col:19)

```

Apologies for the extremely lengthy output and traceback. Any ideas why the tests are failing at this point? I can push the local changes to the remote branch if you'd like to go through them as well. 
Thanks!

**Comment by erwanp** [comment]
The problem happens with the test file: `CO_Tgas1500K_mole_fraction0.01.spec`. 
Can you try to load it and plot it directly?
```
from radis import load_spec
load_spec(`CO_Tgas1500K_mole_fraction0.01.spec`).plot()
```

Anyway, you should still push and I'll have a look!

**Comment by pkj-m** [comment]
@erwanp you were right. I tried loading the spectrum directly and it gave the same error. Most probably has something to do with the changes I made while substituting the non-standard units with `1`. Also I have pushed the commits I made while debugging but it seems the merge with the updated develop branch has made it very cluttered. I think I'll clean up the history once I get the tests running or probably create another PR and close this one!

**Comment by erwanp** [comment]
Ok I'll look that up later today, thanks for pushing!

**Comment by erwanp** [comment]
Couple things: 

- it seems that astropy (unlike pint) does not recognizes "µm": we should write "um" instead. Note that there is a [make_up](https://github.com/radis/radis/blob/develop/radis/spectrum/utils.py#L145) function applied to the spectrum plots, where we could replace the "um" in "µm" at plotting time. 

- The error you saw above is created in this same make_up function. In particular I end up with:

```
'Wavenumber (cm$^\\mathregular{-I/I$_\\mathregular{0}$}$)'
```
I think it's because of the automating renaming of L161: label.replace(r"1", r"I/I$_\mathregular{0}$")

I could suggest that you re-introduce the "I/I0" as a unit, at plotting time, in the plot() function (around L1591):
```
ylabel = make_up("{0} ({1})".format(var, Iunit))
```
Maybe with:
```
if var in ["transmittance", "transmittance_noslit"] and unit=='1':
    Iunit = "I/I0"    # more explicit for the user
```

**Comment by pkj-m** [comment]
@erwanp finally got the whole thing working! Thanks for all the help! Learnt quite a few new things about the package with this issue.
 Also, do you think I should open another PR with a cleaner git history given this PR got a little messed up due to all the merging with develop branch?

**Comment by erwanp** [comment]
I think you could get rid of that with a proper rebase. Have a look at the [Einsteinpy Refactor Commits page](https://github.com/einsteinpy/einsteinpy/wiki/How-to-refactor-commits) (I referenced it on our wiki)  [ what they call `master` is our `develop`, though! ]

File: radis/phys/units.py [hunk 3]
```diff
@@ -207,35 +194,36 @@ def convert_emi2cm(j_nm, wavenum, Iunit0, Iunit):
...
+    # Convert ''mW/cm3/sr/nm' to 'mW/cm3/sr/cm-1'
     j_cm = j_nm * 1e7 / wavenum ** 2
     # note that we discard the - sign.
     # This means that one of the integrals `trapz(j_nm, wavelen)` or
     # `trapz(j_cm, wavenum)` will be < 0 as they are sorted in opposite order.
     # But both plots look alright
 
     # Convert to whatever was wanted
-    j_cm = conv2(j_cm, "mW/cm3/sr/cm_1", Iunit)
+    j_cm = conv2(j_cm, "mW/cm**3/sr/cm-1", Iunit)
```
erwanp, line 219:
```suggestion
    j_cm = conv2(j_cm, "mW/cm3/sr/cm-1", Iunit)
```

radis/phys/units.py
```diff
@@ -269,18 +257,19 @@ def convert_emi2nm(j_cm, wavenum, Iunit0, Iunit):
...
+    # Convert 'mW/cm3/sr/cm-1' to 'mW/cm3/sr/nm'
     j_nm = j_cm * 1e-7 * wavenum ** 2
     # note that we discard the - sign.
     # This means that one of the integrals `trapz(L_nm, wavelen)` or
     # `trapz(L_cm, wavenum)` will be < 0 as they are sorted in opposite order.
     # But both plots look alright
 
     # Convert to whatever was wanted
-    j_nm = conv2(j_nm, "mW/cm3/sr/nm", Iunit)
+    j_nm = conv2(j_nm, "mW/cm**3/sr/nm", Iunit)
```
erwanp, line 272:
```suggestion
    j_nm = conv2(j_nm, "mW/cm3/sr/nm", Iunit)
```

radis/phys/units.py
```diff
@@ -390,20 +380,21 @@ def convert_rad2nm(l_cm, wavenum, Iunit0, Iunit):
...
+    # Convert 'mW/cm2/sr/cm-1' to 'mW/cm2/sr/nm'
     l_nm = l_cm * 1e-7 * wavenum ** 2
     # note that we discard the - sign.
     # This means that one of the integrals `trapz(L_nm, wavelen)` or
     # `trapz(L_cm, wavenum)` will be < 0 as they are sorted in opposite order.
     # But both plots look alright
 
     # Convert to whatever was wanted
-    l_nm = conv2(l_nm, "mW/cm2/sr/nm", Iunit)
+    l_nm = conv2(l_nm, "mW/cm**2/sr/nm", Iunit)
```
erwanp, line 397:
```suggestion
    l_nm = conv2(l_nm, "mW/cm2/sr/nm", Iunit)
```

radis/phys/units.py
```diff
@@ -413,8 +404,8 @@ def convert_universal(
     from_unit,
     to_unit,
     spec=None,
-    per_nm_is_like="mW/sr/cm2/nm",
-    per_cm_is_like="mW/sr/cm2/cm_1",
+    per_nm_is_like="mW/sr/cm**2/nm",
+    per_cm_is_like="mW/sr/cm**2/cm-1",
```
erwanp, line 408:
```suggestion
    per_nm_is_like="mW/sr/cm2/nm",
    per_cm_is_like="mW/sr/cm2/cm-1",
```

radis/spectrum/compare.py
```diff
@@ -979,8 +987,9 @@ def get_wdiff_Idiff():
 
     if title:
         fig.suptitle(title)
-
+    # print("CALLED PLOT DIFF FUNCTION///...........")
```
erwanp, line 990:
```suggestion
```

radis/spectrum/compare.py
```diff
@@ -903,6 +908,9 @@ def get_wdiff_Idiff():
             label=label2
         )
 
+    if var in ["transmittance", "transmittance_noslit"] and wunit == "1":
+        Iunit = "I/I0"  # more explicit for the user
+
```
erwanp, line 913:
```suggestion
    elif var == 'abscoeff' and wunits == "1":
        Iunit = "-ln(I/I0)" # more explicit for the user
    elif var in ["emissivity_no_slit", "emissivity"] and wunits == "1":
        Iunit = "eps"  # more explicit for the user

```

radis/tools/database.py
```diff
@@ -528,7 +535,7 @@ def _json_to_spec(sload, file=""):
     """
 
     conditions = sload["conditions"]
-
+    # print("TILL LINE 535")
```
erwanp, line 538:
```suggestion
```

radis/tools/database.py
```diff
@@ -464,6 +464,7 @@ def load_spec(file, binary=False):  # , return_binary_status=False):
 
 
     """
+    # print("LOAD SPEC CALLED")
```
erwanp, line 467:
```suggestion
```

radis/tools/database.py
```diff
@@ -500,8 +501,14 @@ def _load(binary):
     sload, fixed = _fix_format(file, sload)
 
     # Generate Spectrum
+    # print("CALLING jsontospec function")
```
erwanp, line 504:
```suggestion
```

radis/tools/database.py
```diff
@@ -500,8 +501,14 @@ def _load(binary):
...
     s = _json_to_spec(sload, file)
 
+    # print("BACK FROM jsontospec")
+    for unit in s.units:
+        if s.units[unit] in ["I/I0", "-ln(I/I0)", "eps"]:
+            s.units[unit] = "1"
+    # print("INSIDE THE LOAD FUNCTION, s.units = ", s.units)
```
erwanp, line 511:
```suggestion
```

radis/tools/database.py
```diff
@@ -500,8 +501,14 @@ def _load(binary):
     sload, fixed = _fix_format(file, sload)
 
     # Generate Spectrum
+    # print("CALLING jsontospec function")
     s = _json_to_spec(sload, file)
 
+    # print("BACK FROM jsontospec")
```
erwanp, line 507:
```suggestion
```

radis/tools/database.py
```diff
@@ -559,7 +566,7 @@ def _json_to_spec(sload, file=""):
 
     # Generate spectrum:
     waveunit = sload["conditions"]["waveunit"]
-
+    # print("WAVEUNIT == ", waveunit)
```
erwanp, line 569:
```suggestion
```

radis/tools/database.py
```diff
@@ -570,13 +577,16 @@ def _json_to_spec(sload, file=""):
     else:
         slit = {}
 
+    # print("TILL 579...")
+
```
erwanp, line 581:
```suggestion
```

radis/tools/database.py
```diff
@@ -570,13 +577,16 @@ def _json_to_spec(sload, file=""):
...
     # ... load lines if exist
     if "lines" in sload:
         df = sload["lines"]
         kwargs["lines"] = df
     else:
         kwargs["lines"] = None
 
+    # print("TILL 588...")
```
erwanp, line 589:
```suggestion
```

radis/tools/database.py
```diff
@@ -595,20 +605,24 @@ def _json_to_spec(sload, file=""):
     else:
         kwargs["populations"] = None
 
+    # print("TILL 607...")
```
erwanp, line 608:
```suggestion
```

radis/tools/database.py
```diff
@@ -595,20 +605,24 @@ def _json_to_spec(sload, file=""):
...
     # ... load other properties if exist
     for attr in ["units", "cond_units", "name"]:
         try:
             kwargs[attr] = sload[attr]
         except KeyError:
             kwargs[attr] = None
 
+    # print("TILL 615...")
```
erwanp, line 616:
```suggestion
```

radis/tools/database.py
```diff
@@ -595,20 +605,24 @@ def _json_to_spec(sload, file=""):
...
     s = Spectrum(
         quantities=quantities, conditions=conditions, waveunit=waveunit, **kwargs
     )
 
+    # print("TILL 620...")
```
erwanp, line 621:
```suggestion
```

radis/tools/database.py
```diff
@@ -595,20 +605,24 @@ def _json_to_spec(sload, file=""):
...
     # ... add file
     s.file = basename(file)
 
+    # print("TILL 624...")
```
erwanp, line 625:
```suggestion
```

radis/spectrum/spectrum.py
```diff
@@ -1597,7 +1603,7 @@ def plot(
                 y /= np.nanmax(y)
             Iunit = "norm"
 
-        set_style("origin")
+        # set_style("origin")
```
pkj-m, line 1606: remove #

radis/test/spectrum/test_spectrum.py
```diff
@@ -402,17 +402,18 @@ def _run_testcases(
     # Test conversion of intensity cm-1 works
     # -------------
     test_intensity_conversion(debug=debug, verbose=verbose, *args, **kwargs)
-
+    print("LINE 405 done...")
```
pkj-m, line 405: remove comments from file

radis/spectrum/utils.py
```diff
@@ -15,7 +15,7 @@
 
 # Waverange, wavespaces
 # ... aliases:
-WAVENUM_UNITS = ["cm", "cm-1", "cm_1", "wavenumber"]
+WAVENUM_UNITS = ["cm", "cm-1", "cm-1", "wavenumber"]
```
pkj-m, line 18: remove repetition

radis/spectrum/spectrum.py
```diff
@@ -2404,15 +2404,20 @@ def apply_slit(
...
-                    self.units[qns], unit.replace("cm-1", "cm-1")
-                )
+                if self.units[qns] == "1":
+                    new_unit = "{0}*{1}".format(
+                        self.units[qns], unit.replace("cm-1", "cm-1")
+                    )
+                else:
+                    new_unit = "{0}*{1}".format(
+                        unit.replace("cm-1", "cm-1"), self.units[qns]
+                    )
```
erwanp, line 2402: Comments on the code:

- you should look up these `replace("cm-1", "cm-1")`  (they come from "cm_1" > "cm-1" ) and remove them. 

- do we really need the if/else here? 

---

There is also an underlying physical question, which requires to understand how experiments are done [ this will become a different Issue ]  : 

In my [comment](https://github.com/radis/radis/pull/106#issuecomment-615469626) I explained why the unit changes when you apply an instrumental slit function (to account for the spectrometer limited spectral resolution), if this slit function was normalized at the maximum instead of per area [ note that 99% of the people normalize at the area, we're dealing with a very particular case here, used only when calculating emission spectra to have an idea of the spectral flux]. This applied well to emission spectra.

Here, units == "1" refers to a transmittance spectra (or emissivity, but a similar reasoning applies): 

A theoretical transmittance spectrum is the ratio of an intensity I and a background I0. In practice, no-one ever measures a transmission spectrum directly. Your instrument collects an intensity I first, a background intensity I0 next, divides both and returns a transmittance spectrum I/I0. So your spectral resolution loss applies to I and I0. Your real transmittance spectrum is : `Filter(I) / Filter(I0)` instead of `Filter(I/I0)` as we are doing here. 
In practice the background is quite flat, so we have `Filter(I0)` ~ `I0` , so we end up with `Filter(I)/Filter(I0) ~ Filter(I)/I0 = Filter(I/I0)` , so applying the spectral filter to the theoretical transmittance spectrum directly is eventually a very good approximation.  But notice that  `Filter(I0)` ~ `I0`  assumed the filtering did not change the units! 

Therefore: 

- applying a slit not normalized by area to a transmittance spectrum is meaningless.  (people experienced in absorption spectroscopy like @dcmvdbekerom  will probably confirm!). 

- we want to remain numerically correct no matter what, so we should keep on correcting the unit as implemented here.  

- but we'll end up with transmittance spectra in "nm" or "cm-1" and with values that are not within [0 - 1] anymore: good luck for the user to understand that. Eventually, I think it would be appropriate to refactor apply_slit() and raise an error if we're applying a slit not normalized by area to a Spectrum that contains ratios (like Transmittance, but also Emissivity).  [ note that this is a different Issue ! ] 

What do you think?

pkj-m: Hi @erwanp , thanks for the detailed comment. That's a lot of new information to process so I'll take it up one at a time!
Regarding why the if-else is necessary, that's a thing with astropy where concatenating strings works as follows:
```
u.Unit("1*cm")    #works fine   --> parses as u.cm
u.Unit("cm*1")    #throws error

u.Unit("mW/sr*cm")    #parses it as mW / (sr cm)
u.Unit("cm*mW/sr")    #parses it as (cm mW) / sr   --> what we want in the code
```

Regarding your second point, while I did understand the motivation and procedure behind how the experiment is conducted, I could not see how that connects with the fact that applying slit without area normalization is pointless. I did some research on the internet and the most commonly brought up argument is that peak/max normalization leads to erroneous results? Is that true?
I was also wondering if while calculating the transmittance spectra, could we perhaps allow the user to pass the intensity and background intensity separately with units? Wouldn't this help us drop "I/I0" as placeholders completely?
> but we'll end up with transmittance spectra in "nm" or "cm-1" and with values that are not within [0 - 1] anymore

Can't we normalize the spectrum after passing it through the slit?

Also apologies for any naive questions! While I did take up a few physics courses in my freshman and sophomore year, they were clearly not enough!

erwanp: Regarding your first point `u.Unit("cm*1")`: it's a nice trick, but looks like a Hack! What about replacing "1" with the empty string "" ?  It seems that that's how Astropy writes u.dimensionless_unscaled units anyway: 
```
u.Unit("1").__str__()
>> ""
u.Unit("").__str__()
>> ""
u.Unit("") == u.dimensionless_unscaled
>> True
```
And no need for the Hack then! 

---

Second point:

> the most commonly brought up argument is that peak/max normalization leads to erroneous results

It's not erroneous for emission emission spectra (if you account for the change in units), and sometimes useful: if what you're interested is the maximum energy Flux, for instance in radiative transfer, you'd get a result directly in energy flux  "~W/cm2/sr", instead of spectral energy flux "~W/cm2/sr/nm" ). 
For absorption spectrum, it's erroenous as I described it in my previous comment: because the underlying quantity is a ratio, so we assume that the units of the background spectrum did not change. 

What you're describing (given intensities separately) is possible. Actually, users could do :
```
s1 = Radiance(emission_spectrum)
s2 = Radiance(background_spectrum)
transmittance = s1/s2               # I'm not sure if this work for Spectrum objects, but same is possible using s1.get(), s2.get() ? 
```
But usually your instrument, or the instrument software, will process this directly and return a transmittance spectrum directly. So the main use case takes "I/I0". 

---

>  Can't we normalize the spectrum after passing it through the slit?

We could but that's not what the user asked for, and it would become very dangerous! We better raise an error saying that applying a non-area-normalized slit to a transmittance spectrum is erroneous, and that he should do otherwise. 

We could suggest a solution in the error message: 

```
# during error handling: 
if 'radiance_noslit' in self.get_vars():
    raise ValueError("ValueError("""applying a non-area-normalized slit to a transmittance spectrum is erroneous. Use `method="area"` in apply_slit(). If you're only interested in the emission spectrum, you can extract the emission spectrum the first: `s_rad = Radiance_noslit(s).apply_slit(...)`""")
else:
    raise ValueError("ValueError("""applying a non-area-normalized slit to a transmittance spectrum is erroneous. Use `method="area"` in apply_slit()""")

```

---

Hope this is clear! 
Also whatever physics class you took does not really matter here: it's more "technicalities" (for instance, knowledge of instruments & instrument software) than physics and science at this point! :)

pkj-m: Hi Erwan, thanks for the detailed reply! I understand it much better now. 
Regarding the "hack", while we could use "" instead of "1" while creating `new_unit`, but we'd still be forced to use an if-else block (albeit less complex than the current one) since even after replacing "1" with "" will result in a string such as `cm*` which when parsed by `u.Unit` returns an error. I have made this change. Let me know if you think we could do this in a better way!

For the second point, I think I can see more clearly why you'd want the user to not call the `slit` method when using `max` normalization. While I wouldn't say it's all perfectly clear, the idea and the intuition behind the problem (and it's solution) have become a lot more solid now.
Finally, the code you shared (to give an error in case the user tries to call the `apply_slit` method incorrectly) looks good to me. I think we should do that once we're done with this.

docs/spectrum/spectrum.rst
```diff
@@ -85,7 +85,7 @@ From numpy arrays, use :py:meth:`~radis.spectrum.spectrum.Spectrum.from_array` :
     # w, T are two numpy arrays 
     from radis import Spectrum
     s = Spectrum.from_array(w, T, 'transmittance_noslit', 
-                               waveunit='nm', unit='I/I0')
+                               waveunit='nm', unit='1')
```
erwanp, line 88:
```suggestion
                               waveunit='nm', unit='')  # adimensioned
```

radis/lbl/base.py
```diff
@@ -102,9 +102,9 @@ class BaseFactory(DatabankLoader):
 
     # Output units
     units = {
-        "absorbance": "-ln(I/I0)",
-        "abscoeff": "cm_1",
-        "abscoeff_continuum": "cm_1",
+        "absorbance": "1",
```
erwanp, line 105:
```suggestion
        "absorbance": "",
```

radis/lbl/base.py
```diff
@@ -113,10 +113,10 @@ class BaseFactory(DatabankLoader):
...
         "emisscoeff": "mW/cm3/sr/nm",
         "emisscoeff_continuum": "mW/cm3/sr/nm",
-        "emissivity": "eps",
-        "emissivity_noslit": "eps",
-        "transmittance": "I/I0",
-        "transmittance_noslit": "I/I0",
+        "emissivity": "1",
+        "emissivity_noslit": "1",
+        "transmittance": "1",
+        "transmittance_noslit": "1",
```
erwanp, line 119:
```suggestion
        "emissivity": "",
        "emissivity_noslit": "",
        "transmittance": "",
        "transmittance_noslit": "",
```

radis/phys/blackbody.py
```diff
@@ -230,8 +230,8 @@ def sPlanck(
         conditions=conditions,
         units={
             "radiance_noslit": Iunit,
-            "transmittance_noslit": "I/I0",
-            "absorbance": "-ln(I/I0)",
+            "transmittance_noslit": "1",
+            "absorbance": "1",
```
erwanp, line 234:
```suggestion
            "transmittance_noslit": "",
            "absorbance": "",
```

radis/spectrum/spectrum.py
erwanp: You're right about the `*`! 
Then what about this? :
```
new_unit = (u.Unit(self.units[qns]) * u.Unit(unit)).to_string()
```

Adimensioned.to_string() returns `""` so we should replace `"1"` with `""` for consistency

minouHub:
Hi guys, 
I just read quickly the recent improvements you are implementing. I have a question regarding the unit-less quantities. It seems that @erwanp prefers `""`  instead of `"1"`. Is it technical or motivated by physics ? I looked online and it seems that `"1"` is the official way.
https://physics.stackexchange.com/questions/59978/is-there-a-symbol-for-unitless

erwanp:
As we're moving to Astropy system, we need to comply with their standard. They use:
```
u.dimensionless_unscaled.to_string()
>>> ""
u.Unit("1").to_string()
>>> ""
```

pkj-m:
@erwanp it looks like the CI server failed to procure some data for the tests due to a connection failure. Do we need to restart the build?

minouHub:
No no, the test just failed :)

See https://travis-ci.com/github/radis/radis/jobs/322715193

```
=========================== short test summary info ============================

FAILED radis/test/lbl/test_calc.py::test_calc_spectrum_multiple_molecules - r...

==== 1 failed, 138 passed, 7 deselected, 530 warnings in 336.66s (0:05:36) =====
```

minouHub:
I restarted the job at 10.21 pm :)

pkj-m:
Awesome! Thanks a lot. The build passed this time :)

PR #112, opened 2020-05-03
Github Action Support for Radis

### LINKED ISSUE #103: Install GitHub actions
## Check syntax

Commit 90b9d8f1 created an "undefined name error" in `compare.py`.
The line was not executed in the tests so the error wasn't caught. 

We should add GitHub Action to prevent this from happening again. 

- [ ] [Python Syntax Checker](https://github.com/marketplace/actions/python-syntax-checker) looks good for instance

## Code Linting

- [ ] Following suggestion from @jayanthchandra  we could also add this [Black-code-formatter ](https://github.com/marketplace/actions/black-code-formatter) Action (we already use Black since #58 )

Has anyone experience in implementing GitHub Actions ?

jayanthchandra:
Github Workflow Action for Radis

### Description
Have Included Black Formatter and Python Syntax Checker. 

Fixes #103

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,17 @@
+  name: RadisCI
+
+  # Controls when the action will run. Triggers the workflow on push or pull request
+  # events but only for the master branch
+  on: [pull_request]
+
+  # A workflow run is made up of one or more jobs that can run sequentially or in parallel
+  jobs:
+    build:
+      runs-on: ubuntu-latest
+      steps:
+      - uses: actions/checkout@v2
+      - name: Black Code Formatter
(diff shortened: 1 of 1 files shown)


jayanthchandra:
Black Formatter Action doesn't work, so will be removing that.  See the [comment here. ](https://github.community/t5/GitHub-Actions/Can-t-push-to-forked-repository-on-the-original-repository-s/m-p/35916/highlight/true#M2372)

erwanp:
Just to be clear: it would not work on PR from people that don't have write access to the repository?

jayanthchandra:
This action would trigger for any PR which is being raised in the repo irrespective of their access. 
Closing this, will Squash the commits and raise the PR.

PR #113, opened 2020-05-09
Github Workflow Action for Radis

### LINKED ISSUE #103: Install GitHub actions

jayanthchandra:
Github Workflow Action for Radis

Description
Have Included Black Formatter and Syntax Check

Fixes #103

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,33 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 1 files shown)


jayanthchandra:
The current syntax check works but some one needs to click on the details to view the errors. I would like to add [this](https://github.com/marketplace/actions/lint-action) as it shows the error in the PR itself. @erwanp : If you let me know the python version to support then i will add this.

erwanp:
I see Lint Action has Black and flake8, that's great !!

jayanthchandra:
Not sure with the current action. I tried one more workflow so that we can see the errors in the PR itself rather than going to action details. I have done the setup for my repo, you can check it [here](https://github.com/jayanthchandra/boilerPlate-flask/pull/4/files). I think that uses the flake configs.

erwanp:
Closing, as current CI on Travis does the job !

PR #114, opened 2020-07-25
Allow selection of one of the three different weights

dcmvdbekerom:
As described in the draft paper, different weights may be used for different purposes. 

We distinguish three kinds:
'ZEP': [DEFAULT] Zero Error at (center) Peak
'linear': Linear weight (= minimal calculation time)
'min-RMS': Minimal RMS error

These three kinds are added as keyword parameters to the `SpectrumFactory` class and `_apply_lineshape_DLM` method.
They can be switched by setting the `awL_kind` and/or `awG_kind` keywords to 'ZEP', 'linear', or 'min-RMS'
The kind defaults to 'ZEP' to maintain backwards compatibility, as this was the only implemented kind in previous versions.


radis/lbl/broadening.py
```diff
@@ -1740,7 +1740,7 @@ def _apply_lineshape(self, broadened_param, line_profile, shifted_wavenum):
         return wavenumber, sumoflines
 
     def _apply_lineshape_DLM(
-        self, broadened_param, line_profile_DLM, shifted_wavenum, wL, wG, wL_dat, wG_dat
+        self, broadened_param, line_profile_DLM, shifted_wavenum, wL, wG, wL_dat, wG_dat, awL_kind = 2, awG_kind = 2
```
erwanp, line 1743: I would remove the default values here for clarity (they're overwritten by the default value in SpectrumFactory() anyway )

radis/lbl/factory.py
```diff
@@ -507,6 +523,16 @@ def __init__(
             if self.verbose >= 3:
                 print("DLM used. Defaulting broadening method to FFT")
             # TODO: make it a proper parameter in self.misc or self.params
+            
+            if awL_kind is None:
+                self.misc.awL_kind = 2
+            else:
+                self.misc.awL_kind = awL_kind
```
erwanp, line 530: Should try with : 

```
self.misc['awL_kind'] = awL_kind 
```

This should fix the [current KeyError](https://travis-ci.com/github/radis/radis/jobs/365053344#L1471) in the tests.

dcmvdbekerom:
I was wondering how to do it properly, will do this in the future, thanks!

radis/lbl/factory.py
```diff
@@ -367,6 +381,8 @@ def __init__(
         pseudo_continuum_threshold=0,
         self_absorption=True,
         chunksize=None,
+        awL_kind=None,
+        awG_kind=None,
```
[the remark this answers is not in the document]
dcmvdbekerom: I tried to use the formatting of `chunksize` as much as possible but you're right setting the default right away is much preferred here

dcmvdbekerom:
I updated the weight labels (now string instead of int) to make them more descriptive:
        'ZEP':      [DEFAULT] Zero Error at (center) Peak
        'linear':   Linear weight (= minimal calculation time)
        'min-RMS':  Minimal RMS error

dcmvdbekerom:
New updates in the manuscript make this PR not relevant anymore

erwanp:
Ok !

Issue #115, opened 2020-07-27
Improve definition of width-grid resolution (res_L, res_G)


dcmvdbekerom:
With the manuscript draft being updated, we now have a much better handle on how to precisely define the errors in the width-grid for the "DLM"-approach. It would be good to propagate these updates to the RADIS code, but some executive decisions have to be made on how to implement it.

There are basically three candidates of how the user could specify the desired error:

- **`p`**                                                     -> _relative increase between two width-gridpoints_
- **`log_p`** = np.log(p)                            -> _actual step-size of the logarithmic width-grid_
- **`err_rms`** = RMS[Y_i]*np.log(p)/12    -> _contribution to RMS-error by a single line._ 
(The factor 1/12 = 1/2 x 1/6, where 1/6 comes from integrating the grid-alignment assuming uniform distribution between 0 and 1)

Each has their own advantage, namely:

- **`p`**: unambiguous what is meant, and makes sense from a human perspective (e.g. `p = 1.2` means 20% increase between gridpoints)
- **`log_p`**: equally unambiguous, but perhaps makes less sense from a human perspective (e.g. what really does `log_p = 0.2` mean?). Advantage is that when `log_p` is closer to zero, it means higher resolution, which perhaps is more intuitive.
- **`err_rms`**: Makes most sense from human point of view, and resolution also gets better as `err_rms` gets closer to zero. Problem is that choosing the RMS-error as the defining error measure is somewhat arbitrary, and depending on the kind of weight used (1,2, or 3), the same value of `err_rms` will lead to different `p` and `log_p` values, which the user may not be aware of (but perhaps equally possibly they do not care about)

Anyway so that's a decision that has to be made. My personal preference is in the presented order (with **`p`** most favorable), but of course I'd like to hear everyone's thoughts.

erwanp:
@dcmvdbekerom is this one still up-to-date? I can look at it tomorrow !

dcmvdbekerom:
It is! And while working on it more I have changed my mind, I think **log_p** should be the value used! This is because I found it's really convenient to have a parameter that gets better as it gets closer to 0, instead of when it gets closer to 1.

erwanp:
I actually like `err_rms` better, not only because it is also better as it gets closer to 0, but because I think that it gives the user a great sense of control over what is happening. 

[...]

PR #116, opened 2020-08-05
Implementing optimized weights

dcmvdbekerom:
### Description

This pull request implements optimized weights as per the updated manuscript; Optimized weights are dependent on both the Gaussian and Lorentzian grid-alignment. Previous default, which minimized the error of Gaussian and Lorentzian error in the extreme case of pure Gaussian or pure Lorentzian, was actually worse than simple weights (a=tau) whenever the Voigt had both Gaussian and Lorentzian components.

The current optimized weights have been analytically optimized and numerically shown to significantly reduce the RMS-error compared to simple weights.

A boolean keyword `optimized_weights` is added to the `SpectrumFactory`-class to toggle optimized/simple weights.

Note:

Optimized weights can go above one and below zero (negative). This feels a bit weird but there is nothing fundamentally wrong with this, since the area of the full line will by definition always stay normalized.

The previous weights were hard coded to be "minimized-RMS for pure Gaussian/Lorentzian", and have been removed since they perform worse than simple weights. This means that the default option had to be changed, and it is set to optimized weights. This also changed the hard-coded values in the test in `test_calc.py`, which have been updated. 

The tests in `test_calc.py` now use the simple weights (`optimized_weights = False`), so that the optimized weights can be replaced with _even more optimized_ weights in the future -- should they ever be found -- without compromising the test.

Fixes #<Issue Number>


dcmvdbekerom:
When we make a decision on https://github.com/radis/radis/issues/115 I can include it in this PR.

radis/lbl/factory.py
```diff
@@ -266,7 +266,12 @@ class SpectrumFactory(BandFactory):
             argument: in this case, the DLM optimization for lineshape calculation 
             is used. Broadening method is automatically set to ``'fft'``. 
             See :py:attr:`~radis.lbl.broadening.BroadenFactory._broadening_method`.
-        
+
+    optimized_weights: bool
+        Whether to use optimized weights or simple weights.
+            True:  Optimized weights [DEFAULT]
+            False: Simple weights
+
```
erwanp, line 274: Sounds great !

My only concern is that we're introducing a new parameter than will only be used if the DLM is used, which itself is decided in another parameter (currently "chunksize", although it's not related, only because it was a quick hack)

Maybe it's time for a cleaner implementation from the user-interface point of view. 

For instance : 

```suggestion

    lineshape_interpolation : ``"simple_weights"``, ``"optimized_weights"``, ``None``
        If not ``None``, the DLM optimization for lineshape calculation is used : 
        - "optimized_weights" : weights optimized according to [DLM_article]_ 
        - "simple_weights" 
        
        If using the DLM optimization, broadening method is automatically set to ``'fft'``.  
        
        If ``None``, no lineshape interpolation is performed and the lineshape of all lines is calculated. 
        
        Refer to [DLM_article]_ for more explanation on the DLM method for lineshape interpolation. 
        
        Default ``"optimized_weights"`` 
```

erwanp: In that case I would remove Line 514 : 

```
        if chunksize == "DLM":
            self._broadening_method = "fft"
            if self.verbose >= 3:
                print("DLM used. Defaulting broadening method to FFT")
            # TODO: make it a proper parameter in self.misc or self.params
```

and have a cleaner version  : 

```
        if lineshape_interpolation is not None:
            if chunksize is not None: 
                raise ValueError("chunksize is not used when using lineshape_interpolation")
            self._broadening_method = "fft"   # TODO : replace with user parameter at some point (broadening_method=), and here we would use set to FFT if the default value is given, or raise a NotImplementedError if a different value is given. 
            if self.verbose >= 3:
                print("DLM used. Defaulting broadening method to FFT")
```

erwanp: and you could either keep the internal self.misc.optimized_weight boolean parameter (and set it somewhere around here), or turn it into self.misc.lineshape_interpolation [name can be changed of course] with the `if optimized_weights:` of broadening.py becoming `if lineshape_interpolation == "optimized_weights":`

radis/test/lbl/test_calc.py
```diff
@@ -148,32 +149,39 @@ def test_calc_spectrum(verbose=True, plot=True, warnings=True, *args, **kwargs):
...
-            0.17106096,
-            0.15400039,
-            0.13374285,
-            0.11930822,
-            0.10880631,
-            0.11111394,
-            0.04582291,
-            0.00247955,
-            0.00144128,
+            0.29043204,
```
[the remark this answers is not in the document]
dcmvdbekerom: That's explained here:
>The previous weights were hard coded to be "minimized-RMS for pure Gaussian/Lorentzian", and have been removed since they perform worse than simple weights. This means that the default option had to be changed, and it is set to optimized weights. This also changed the hard-coded values in the test in test_calc.py, which have been updated.

>The tests in test_calc.py now use the simple weights (optimized_weights = False), so that the optimized weights can be replaced with even more optimized weights in the future -- should they ever be found -- without compromising the test.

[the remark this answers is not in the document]
erwanp: Sounds good, I didn't understand the difference between pure Gaussian-Lorentzian and simple weights !!

radis/lbl/factory.py
erwanp: @dcmvdbekerom if you agree you can directly commit the Suggestion from here.

dcmvdbekerom: Completely agree, nothing else to add!

dcmvdbekerom: Actually one thing...: Do we keep `chunksize` in case people use the old method? I'm guessing _yes_ but like to hear your thoughts

dcmvdbekerom:
I noticed you already had a `lineshape_optimization` parameter that encompasses both the DLM toggle and the chunksize, perhaps this is the way to go? i.e.: a single parameter that by assigning "simple" or "min-RMS" toggles DLM, and by assigning None or int or float, the legacy broadening is used.

Either way, the test gets stuck at the bands test. The spectra are quite similar but (0.02% at worst), but don't quite make the 1e-5 threshold. I'm not sure why that is; the output should not have changed compared to previous commit, which passed all tests.

erwanp:
You're right, it's weird. 

Left : [old one](https://github.com/radis/radis/pull/116/commits/ca3b21194284f28c45d6bfe2b320239f070adaa4), passing. Right : [new one](https://github.com/radis/radis/pull/116/commits/5e1be08cae1e47e99ceddd0e602bf3041973231c), failing.

image

Difference is not really obvious, but I'll keep on looking into it !

dcmvdbekerom:
https://github.com/radis/radis/pull/116/commits/8a4af32b2ad836c3ec7a5f87dee80ff839c307d7 was the first to fail the test, I'll look into it!

erwanp:
Hey, sorry I didn't think you were working on it right now. I actually had found that too, and fixed it the same way.
I did a few other changes in the process, like make broadening_method a proper argument. 
And cleaned some code. I'll fix all the conflicts locally and re-push.

erwanp:
Tests should go through. 

Interesting thing : while runnning everything locally I also ran some of the larger calculations that are not tested automatically, because they require the full line databases. For instance, this experimental image of CO2 radiation in an atmospheric air plasma torch (line-of-sight of 24 slabs). 
It was the first time i could run this with the DLM on, so I tried : 

image

Interestingly, not only it's faster than the legacy optimization (pseudo-continuum)  [not by to much, but there aren't so many lines in there. And also pseudo-continuum was indeed quite fast !]   ... but it looks also much closer to the experimental spectra, mostly in the bandhead. I think we're seeing the effect of the large bandwings, that are otherwise cut in any legacy method (

(left : psuedo-continuum ; right : DLM.  Compare the black [experimental] to the thick red [line of sight spectrum] ) 

image

Issue #102, opened 2020-04-15
Weird behavior of 'get_residual'


minouHub:
Hello guys,

We recently realized the function 'get_residual' was not working properly when two spectra with two different units are compared. I realized a quick fix in this PR: #101

**Code:**
```
from radis import calc_spectrum, get_residual, experimental_spectrum
s1 = calc_spectrum(1900, 2300,
                  molecule='CO',
                  isotope='1,2,3',
                  pressure=1.01325,
                  Tgas=1000,
                  mole_fraction=0.1,
                  )
s1.apply_slit(1, 'nm')

#s2 is like s1 but at 1001 K
s2 = calc_spectrum(1900, 2300,         
                  molecule='CO',
                  isotope='1,2,3',
                  pressure=1.01325,
                  Tgas=1001, #HERE IS THE DIFFERENCE
                  mole_fraction=0.1,
                  )
s2.apply_slit(1, 'nm')
residual_ref = get_residual(s1, s2, 'radiance', ignore_nan=False, normalize=True)


w1, I1 = s1.get('radiance', copy=False, wunit=s1.get_waveunit())
#Fake experimental spectrum (identical)
s_expe_1 = experimental_spectrum(w1, I1, Iunit="mW/cm2/sr/nm", wunit="cm-1") 
#Fake experimental spectrum with a new unit (but identical)
s_expe_2 = experimental_spectrum(w1, I1*1e-3, Iunit="W/cm2/sr/nm", wunit="cm-1") 
residual1 = get_residual(s1, s_expe_1, 'radiance', ignore_nan=False, normalize=True)
residual2 = get_residual(s1, s_expe_2, 'radiance', ignore_nan=False, normalize=True)

print(residual1, residual2, residual_ref)
```

**New Output:**
`0.0 0.0008176423186518765 5.532303126258176e-07` 

1. The first one is expected and perfect
2. The second one should be improved.
3. The third is just to give you an idea what change will be induced by a increase of 1 K

In fact, this function is a duplicate of `compare_with` where methods to plot the difference of two plots are already correctly implemented. The good fix idea is ready:

- [x] find the common functions in `plot_diff` and `get_residual` 
- [x] write the functions of the previous point (copy-paste) 
- [x] assert NANs are correctly handled in `plot_diff` and `get_residual` 
- [x] implement the previous test 

PS : I think @erwanp wrote `compare_with`, which is a sign of good quality :+1: 

Thank you for you help,
Minou

erwanp:
[...]

Issue #109, opened 2020-04-24
Add unit support in multiply and other spectrum operations


erwanp:
### Problem

As mentioned in https://github.com/radis/radis/pull/101#pullrequestreview-394154396, the problem with the normalization of #101  is that [multiply](https://github.com/radis/radis/blob/develop/radis/spectrum/operations.py#L480) does not use dimensioned units. 

> multiply does not currently handle multiplications with dimensioned coefficients. These should change the units of the output Spectrum.

### Suggested fix

(this is related to #57 , which focused on Spectrum methods, while here we have operations between 2 Spectra, or 1 Spectrum or arrays/floats) 

- [x] use astropy.units for the conversion. We now have (#106 ) a radis.phys.units.Unit based on astropy.units:
```
a = b * c 
unit_a = (Unit(unit_b) * Unit(unit_c)).to_string() 
```

- [x] add tests!
Such as: 
```
Radiance(s) -= 0.01 * Unit("W/cm2/sr/µm")
```

- [x] same for all the others: add, divide, etc.

Finally, some cleaning:

- [x] Implement a new normalize() method with the different options of get_residual() (that uses our new dimensioned multiply method)
- [ ] Replace most of the normalizing code of get_residual() with a simple call to `get_diff(s1.normalize(*args), s2.normalize(*args) )` ? 
- [ ] that should eventually fix #102. Check and close Issue.

- [ ] Also i think we can do the same in .plot() , if we know have a normalize method()

Final remark:

how do you normalize a transmittance spectrum? It does not really make sense, but is not forbidden here. Should raise a NotImplementedError or ValueError.

minouHub:
I'll begin with the easy stuff. 

- If it is not physical to normalize a transmittance, the user should be informed. A `ValueError` with the following warning should do the work "A normalized Transmittance spectrum has no physical meaning".

- @erwanp. You gave clear instruction, thanks. Can you also close all the other PR and Issues related please ? I guess you're the only one who has a broad view on this problem, and for me it would help to have one channel to follow.

erwanp:
Did, I closed #101 . 

As for your first answer:

- In general, I think that normalizing shouldn't be possible if there are multiple spectral arrays. Let's say there are absorbance (A) and absorption coefficient (k). All are tightly bound to one another : A = k * L ... you can't normalize A and k at the same time. 

[...]

PR #120, opened 2020-08-07
Fix get residual

minouHub:
Hi guys,

This is related to #102. I did what I wanted to do:
- [x] assert NANs are correctly handled in plot_diff and get_residual
- [x]  find the common functions in plot_diff and get_residual
- [x] write the functions of the previous point (copy-paste)
- [x] **quickly** implement the previous test

Please, tell me what you think of the output. I removed the option to normalize on a certain wave-range. I think it just complicates this function a bit too much. What is left:
- [x] write the help of each function
- [x] remove previous get_residual which is commented
- [x] **properly** implement the tests with nice functions
- [x] (you tell me)

### DIFF
--- a/radis/spectrum/compare.py
+++ b/radis/spectrum/compare.py
@@ -257,184 +257,250 @@ def get_distance(s1, s2, var, wunit="default", Iunit="default", resample=True):
     return curve_distance(w1, I1, w2, I2, discard_out_of_bounds=True)
 
 
-def get_residual(
-    s1,
-    s2,
-    var,
-    norm="L2",
-    ignore_nan=False,
-    diff_window=0,
-    normalize=False,
-    normalize_how="max",
-):
-    # type: (Spectrum, Spectrum, str, bool, int) -> np.array, np.array
-    """ Returns L2 norm of ``s1`` and ``s2``
+# def get_residual(
+#     s1,
+#     s2,
+#     var,
+#     norm="L2",
+#     ignore_nan=False,
+#     diff_window=0,
+#     normalize=False,
+#     normalize_how="max",
+# ):
+#     # type: (Spectrum, Spectrum, str, bool, int) -> np.array, np.array
+#     """ Returns L2 norm of ``s1`` and ``s2``
 
-    For ``I1``, ``I2``, the values of variable ``var`` in ``s1`` and ``s2``, 
-    respectively, residual is calculated as:
+#     For ``I1``, ``I2``, the values of variable ``var`` in ``s1`` and ``s2``, 
+#     respectively, residual is calculated as:
         
-    For ``L2`` norm:
+#     For ``L2`` norm:
 
-        .. math::
+#         .. math::
     
-            res = \\frac{\\sqrt{\\sum_i {(s_1[i]-s_2[i])^2}}}{N}.
+#             res = \\frac{\\sqrt{\\sum_i {(s_1[i]-s_2[i])^2}}}{N}.
 
-    For ``L1`` norm:
+#     For ``L1`` norm:
             
-        .. math::
+#         .. math::
     
-            res = \\frac{\\sqrt{\\sum_i {|s_1[i]-s_2[i]|}}}{N}.
+#             res = \\frac{\\sqrt{\\sum_i {|s_1[i]-s_2[i]|}}}{N}.
+
+
+#     Parameters    
+#     ----------
+
+#     s1, s2: :class:`~radis.spectrum.spectrum.Spectrum` objects
+#         if not on the same range, ``s2`` is resampled on ``s1``.
+
+#     var: str
+#         spectral quantity
+
+#     norm: 'L2', 'L1'
+#         which norm to use 
+
+#     Other Parameters
+#     ----------------
+
+#     ignore_nan: boolean
+#         if ``True``, ignore nan in the difference between s1 and s2 (ex: out of bound)
+#         when calculating residual. Default ``False``. Note: ``get_residual`` will still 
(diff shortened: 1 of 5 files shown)


minouHub:
> Sorry @minouHub but it's very hard to understand the changes because all lines are marked as changed : see https://github.com/radis/radis/pull/120/files.
> Can't you simply remove them instead of commenting them ? (it's in Git history anyway)
> 
> Also : I thought that #102 would be fixed by #109, in a more elegant way. Don't you think ?

I removed the previous `get_residual`. The #109 is a good idea, however we just need `get_residual` to work properly. The way it is implemented has no impact for me. If you find using the #109 idea is significantly **faster**, then it would be ok to implement it in my opinion.

radis/spectrum/compare.py
```diff
@@ -319,116 +318,54 @@ def get_residual(
         how to normalize. ``'max'`` is the default but may not be suited for very
         noisy experimental spectra. ``'area'`` will normalize the integral to 1.
         ``'mean'`` will normalize by the mean amplitude value
-
```
erwanp, line 322: is it normal to have removed this ?

radis/spectrum/compare.py
```diff
@@ -319,116 +318,54 @@ def get_residual(
...
+        w2, I2 = s2.get(var, wunit=wunit, copy=False)
+
+        if normalize_how == "max":
+            ratio = np.nanmax(I1) / np.nanmax(I2)
+        elif normalize_how == "mean":
+            ratio = np.nanmean(I1) / np.nanmean(I2)
+        elif normalize_how == "area":
+            norm1 = np.abs(nantrapz(I1, w1))
+            norm2 = np.abs(nantrapz(I2, w2))
+            ratio = norm1 / norm2
```
erwanp, line 345: @minouHub i do not understand : `normalize_how` changes the ratio, but `ratio` is not used afterwards. In every case we normalize by the maximum.

minouHub: sorry, wrong copy paste. i correct this

minouHub: also added a `Warning` because the "area" method seams to be very ineffective

radis/spectrum/compare.py
minouHub: That was a mistake. I put back the "See Also"

Do you really want to have something else than the L2-norm implemented ?

erwanp: Ok, it may be overcomplicating things.

radis/spectrum/compare.py
```diff
@@ -261,174 +261,129 @@ def get_residual(
...
+        w2, I2 = s2.get(var, wunit=wunit, copy=False)
+        if normalize_how == "max":
+            norm1 = np.nanmax(I1) 
+            norm2 = np.nanmax(I2)
+        elif normalize_how == "mean":
+            norm1 = np.nanmean(I1) 
+            norm2 = np.nanmean(I2)
+        elif normalize_how == "area":
+            norm1 = np.abs(nantrapz(I1, w1))
+            norm2 = np.abs(nantrapz(I2, w2))
```
erwanp, line 356: Since 0.9.26 I think we can now use the following, quite simple syntax :

```suggestion
        s1 = s1.normalize(inplace=False, normalize_how=normalize_how, wrange=normalize)
        s2 = s2.normalize(inplace=False, normalize_how=normalize_how, wrange=normalize)
```

What do you think ?

radis/spectrum/compare.py
```diff
@@ -261,174 +261,129 @@ def get_residual(
...
-                # norm2 = s2.get_integral(
-                #    var, wunit=s1.get_waveunit(), Iunit=s1.units[var]
-                # )
-                w1, I1 = s1.get(var, copy=False)
-                norm1 = nantrapz(I1, w1)
-                w2, I2 = s2.get(var, Iunit=s1.units[var], wunit=s1.get_waveunit())
-                norm2 = nantrapz(I2, w2)
+            raise ValueError("Unexpected `normalize_how`: {0}".format(normalize_how))
+        I1 /= norm1
+        I2 /= norm2
```
erwanp, line 360:
```suggestion
```

radis/spectrum/compare.py
```diff
@@ -261,174 +261,129 @@ def get_residual(
...
-            else:
-                raise ValueError(
-                    "Unexpected `normalize_how`: {0}".format(normalize_how)
-                )
-            # Ensure we use the same unit system!
-            s1 = multiply(s1, 1 / norm1, var=var)
-            s2 = multiply(s2, 1 / norm2, var=var)
+        if verbose:
+            ratio = norm1 / norm2
+            print(("Rescale factor: " + str(ratio)))
```
erwanp, line 364:
```suggestion
```

radis/spectrum/compare.py
```diff
@@ -555,6 +510,54 @@ def _get_defaults(
...
+                wdiff, Idiff = get_diff(
+                    s1, s2, var=var, wunit=wunit, Iunit=Iunit, diff_window=diff_window
+                )
+            elif method == "ratio":
+                wdiff, Idiff = get_ratio(s1, s2, var=var, wunit=wunit, Iunit=Iunit)
+            else:
+                raise ValueError("Unknown comparison method: {0}".format(method))
+            wdiffs.append(wdiff)
+            Idiffs.append(Idiff)
+        else:
```
erwanp, line 542: shouldn't we normalize() here ?

erwanp (changes requested):
adding new normalize() method

erwanp:
Replaced by #133, tests we reused there

PR #122, opened 2020-08-12
dimensionned units for multiply/division 

### LINKED ISSUE #109: Add unit support in multiply and other spectrum operations

erwanp:
### Description

- Add unit support for multiply / division. Fixes #109 
- Also add min(), max() methods for spectra with unique quantities 

Allows to use the following syntax : 
```
# add a baseline : 
s += 0.1 * u.Unit("W/cm2/sr/nm")   

# normalize manually : 
s /= s.max() * u.Unit("mW/cm2/sr/nm")   
assert s.units['radiance'] == "1.0" # normalized
assert s.max() == 1.0

# Calibrate :
s.units['radiance'] = 'count'
s *= 100 * u.Unit("mW/cm2/sr/nm/count")

```

Edit : 

also added a normalization method directly : 

```

    s.normalize()
    s.normalize(normalize_how="max")
    s.normalize(normalize_how="area")
    
You can also normalize only on limited range. Useful for noisy spectra ::

    s.normalize(wrange=(2250, 2500), wunit="cm-1", normalize_how="mean")    

This returns a new spectrum and does not modify the Spectrum itself. To do so use::

    s.normalize(inplace=True)
    
You can chain the commands :: 

    s.normalize().plot()
```

### DIFF
--- a/radis/spectrum/operations.py
+++ b/radis/spectrum/operations.py
@@ -55,7 +55,7 @@ Most of these functions are implemented with the standard operators. Ex::
 
 from __future__ import print_function, absolute_import, division, unicode_literals
 
-# from radis.misc.curve import curve_substract, curve_add
+import astropy.units as u
 from radis.spectrum import Spectrum
 from radis.phys.convert import (
     cm2nm,
@@ -525,8 +525,8 @@ def add_constant(s, cst, unit=None, var=None, inplace=False):
     ----------
     s: Spectrum objects
         Spectrum you want to modify
-    cst: Float
-        Constant to add.
+    cst: float
+        Constant to add. 
     unit: str
         unit for ``cst``. If ``None``, uses the default unit in ``s`` for 
         variable ``var``.
@@ -551,6 +551,9 @@ def add_constant(s, cst, unit=None, var=None, inplace=False):
     """
     # Check input
     var = _get_unique_var(s, var, inplace)
+
+    # Note: using dimensioned value for `cst` will make it an array,
+    # processed by add_array
 
     # Convert to Spectrum unit
     if unit is not None:
@@ -583,9 +586,9 @@ def add_array(s, a, unit=None, var=None, inplace=False):
     ----------
     s: Spectrum objects
         Spectrum you want to modify
-    a: numpy array
+    a: numpy array, or `~astropy.units.quantity.Quantity`
         array to add. Must have the same length as variable ``var`` in Spectrum 
-        ``s``
+        ``s``. Can be dimensioned with :py:mod:`~astropy.units`. 
     unit: str
         unit for ``a``. If ``None``, uses the default unit in ``s`` for 
         variable ``var``.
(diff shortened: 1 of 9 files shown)


radis/spectrum/operations.py
```diff
@@ -477,15 +463,18 @@ def _get_unique_var(s, var, inplace):
...
 
     Parameters
     ----------
-    s: Spectrum objects
-        The spectra to multiply.
+    s: Spectrum object
+        The spectrum to multiply.
     coef: float
         Coefficient of the multiplication.
+    unit: str.
```
erwanp, line 475:
```suggestion
    unit: str, or `~astropy.units.core.Unit`
```

erwanp: (misc) improve docs

radis/spectrum/spectrum.py
```diff
@@ -3574,6 +3575,159 @@ def __str__(self):
...
+                norm = np.abs(nantrapz(I[b], w[b]))
+                norm_unit = u.Unit(s.units[var]) * u.Unit(wunit)
+            else:
+                raise ValueError(
+                    "Unexpected `normalize_how`: {0}".format(normalize_how)
+                )
+
+            out = multiply(s, 1 / norm, unit=norm_unit, inplace=inplace)
+
+        else:
```
minouHub, line 3707: (details)
I would make clearer the if ... then ... else here


```
 if len(wrange) == 2:
             blabla
 elif isinstance(wrange, (int, float)):
             blabla
  else:
             UserWarning

```

minouHub (approved):
Except the minor comment about the `if .. else` everything is ok for me

radis/spectrum/spectrum.py
erwanp: Great, and it would also to normalize at one specific point. I forgot this one :)

erwanp:
> 
> 
> Except the minor comment about the `if .. else` everything is ok for me

no time to implement it now, i opened a new Issue #127 and I'm merging this one now !

PR #123, opened 2020-08-13
Ensure load_databank() does not read cached files

### LINKED ISSUE #121: load_databank() not handling wildcard input properly
Using a wildcard input for `load_databank()` method in the SpectrumFactory class throws an error if executed for the second time as the wildcard entry loads the cached h5 file instead of the actual databank file.

## Expected Behavior

Executing `sf.load_databank("/path/to/file*")` when the cached files are already present in the same directory should either load exclusively from the cached files or ignore them completely.

## Current Behavior

Executing `sf.load_databank('/path/to/file*')` works the way it is expected to if there is no cached file which fits the wildcard entry. Apart from loading the data, RADIS also writes a cached h5 file in the same directory with identical name (but a h5 extension).
The next time `sf.load_databank('/path/to/file*')` is executed, the h5 file is read instead of the actual database file and an error is returned.

## Possible Solution

Either identify cached files and skip them before loading, or save cached files in a sub-directory instead of the main folder.

## Steps to Reproduce (for bugs)

1. Instantiate a `SpectrumFactory` object:
```
 sf = radis.SpectrumFactory(
        wavenum_min=2000,
        wavenum_max=3000,
        pressure=p,
    )     
```
2. Load the test dataset file. 
`file_dir = radis.test.utils.getTestFile("cdsd_hitemp_09_fragment.txt")`
`test_file = file_dir[:-8] + '*'` (to get the wildcard format)
`sf.load_databank(path=test_file, format="cdsd-hitemp", parfuncfmt="hapi")`

If no error was thrown, that was probably because a corresponding h5 file did not exist in the folder. Run the line above again.

Error:
`UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc8 in position 1: unexpected end of data`

pkj-m:
Ensure that `SpectrumFactory.load_databank()` does not read .h5 cached files written in the same directory where the main dataset files are presented.

Fixes #121

### DIFF
--- a/radis/lbl/loader.py
+++ b/radis/lbl/loader.py
@@ -73,6 +73,7 @@ from radis.io.hitran import (
 )
 
 # from radis.io.hitran import hit2dfTAB
+from radis.misc.cache_files import cache_file_name
 from radis.misc.warning import EmptyDatabaseError
 from radis.io.query import fetch_astroquery
 from radis.io.tools import drop_object_format_columns, replace_PQR_with_m101
@@ -1288,9 +1289,11 @@ class DatabankLoader(object):
         # Check input types are correct
         if isinstance(path, string_types):  # make it a list
             path = get_files_from_regex(path)
-            path = [
-                p for p in path if not (p.endswith(".h5") and p[:-3] in path)
-            ]  # ignore the cached h5 files when the main dataset files are present
+            filtered_path = [fname for fname in path]
+            for fname in path:
+                if cache_file_name(fname) in path and cache_file_name(fname) != fname:
+                    filtered_path.remove(cache_file_name(fname))
+            path = filtered_path
 
         if dbformat not in KNOWN_DBFORMAT:
             # >>>>>>>>>>>
--- a/radis/lbl/loader.py
+++ b/radis/lbl/loader.py
@@ -1288,9 +1288,6 @@ class DatabankLoader(object):
         # Check input types are correct
         if isinstance(path, string_types):  # make it a list
             path = get_files_from_regex(path)
-            path = [
-                p for p in path if not (p.endswith(".h5") and p[:-3] in path)
-            ]  # ignore cached files in the directory if the original dataset is present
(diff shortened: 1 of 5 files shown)


radis/test/lbl/test_loader.py
```diff
@@ -86,9 +86,27 @@ def test_retrieve_from_database(
...
+    file_dir = getTestFile("cdsd_hitemp_09_fragment.txt")
+    test_file = file_dir[:-8] + "*"
+    sf.load_databank(path=test_file, format="cdsd-hitemp", parfuncfmt="hapi")
+
+    try:
+        sf.load_databank(path=test_file, format="cdsd-hitemp", parfuncfmt="hapi")
+    except UnicodeDecodeError:
+        loaded_files = False
+
+    assert loaded_files
```
erwanp, line 103: I would make a more explicit error  :

```suggestion
    try:
        sf.load_databank(path=test_file, format="cdsd-hitemp", parfuncfmt="hapi")
    except UnicodeDecodeError as err:
        raise UnicodeDecodeError("Couldnt load database the 2nd time. This may be due to cache files trying to be read as normal files") from err
```

radis/test/lbl/test_loader.py
```diff
@@ -86,8 +86,25 @@ def test_retrieve_from_database(
         rmtree(temp_database_name)
 
 
+def test_ignore_cached_files():
```
erwanp, line 89: add more information and reference the Issue ?

radis/lbl/loader.py
```diff
@@ -1289,6 +1289,11 @@ def _check_database_params(
         # Check input types are correct
         if isinstance(path, string_types):  # make it a list
             path = get_files_from_regex(path)
+            filtered_path = [fname for fname in path]
```
erwanp, line 1292: leave a small comment it for why we're doing that (maybe reference the Issue too)

PR #137, opened 2020-12-05
Combine coverage, flake8, pytest configs to setup.cfg

suzil:
### Description

A lot of Python configuration can be centralized to the `setup.cfg` file. 

- **coverage.py**: https://coverage.readthedocs.io/en/coverage-4.0.3/config.html#configuration-files
- **flake8**: https://flake8.pycqa.org/en/latest/user/configuration.html#configuration-locations
- **pytest**: https://docs.pytest.org/en/stable/customize.html#setup-cfg

It's nicer to consolidate configuration to one file unless there's a reason not to.


erwanp:
Good! These files were added progressively, there is no reason not to combine them. Merging !

PR #139, opened 2020-12-05
Run flake8 in CI

suzil:
### Description

There was configuration for flake8 but it wasn't running in CI.

- Enable flake8 to run in CI with `pre-commit run --all-files`
- Remove unnecessary install for lint in Travis config
- Fix flake8 errors, such as unused variables and unused imports
- Cleanup flake8 config. The file ignores weren't doing anything before

### DIFF
--- a/radis/test/tools/test_database.py
+++ b/radis/test/tools/test_database.py
@@ -64,9 +64,8 @@ def test_database_functions(
     # ... get_unique(). If we were to add new
     # ... test cases with matching conditions
     # ... let's add more criteria to keep it unique
-    # match = db.get(**s.conditions)
-    #    assert len(match) == 1
-    # TODO: not working in Python 2.7 yet
+    match = db.get(**s.conditions)
+    assert len(match) == 1
 
     # Database add method
     s2 = s.copy()
--- a/radis/test/tools/test_database.py
+++ b/radis/test/tools/test_database.py
@@ -64,7 +64,7 @@ def test_database_functions(
     # ... get_unique(). If we were to add new
     # ... test cases with matching conditions
     # ... let's add more criteria to keep it unique
-    match = db.get(**s.conditions)
+    # match = db.get(**s.conditions)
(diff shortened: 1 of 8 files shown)


radis/test/tools/test_database.py
```diff
@@ -64,7 +64,7 @@ def test_database_functions(
     # ... get_unique(). If we were to add new
     # ... test cases with matching conditions
     # ... let's add more criteria to keep it unique
-    match = db.get(**s.conditions)
+    # match = db.get(**s.conditions)
     #    assert len(match) == 1
     # TODO: not working in Python 2.7 yet
```
erwanp, line 69: Fully reactivating this test instead, adding the "assert" back (tests passed locally).  

```suggestion
    match = db.get(**s.conditions)
    assert len(match) == 1
```

setup.cfg
```diff
@@ -10,10 +10,11 @@ markers =
     needs_connection: requires an internet connection (deselect with '-m "not needs_connection"')
 
 [flake8]
-ignore = E, W, C901, radis/io/hapi.py, */__init__.py
+ignore = C901
 max-line-length = 89
 max-complexity = 18
-select = B,C,E,F,W,T4,B9
+select = B,C,F,T4,B9
+per-file-ignores = __init__.py:F401,F403
```
erwanp, line 17: @suzil  the `radis/io/hapi.py` does not look not ignored here (it's the Python interface from the HITRAN group, we include it in RADIS for comparison in the tests as they didn't make it a library installable from pip. But it does not comply with flake so should be ignored. ) 
You tests passed, though, so it is probably ignored anyway ?

erwanp (changes requested):

setup.cfg
suzil: I thought it would be OK to remove it as ignored since there are no current flake8 violations?

erwanp (approved):

PR #58, opened 2020-01-14
Quality Checks using Black

ritzvik:
As discussed over Gitter room
@erwanp 

I'll write down the detailed changes in a separate comment

PR #140, opened 2020-12-05
Add isort

suzil:
### Description

Add isort using pre-commit.

I also noticed there's a redundant flake8 `exclude` in pre-commit, but it's already in `setup.cfg` so I removed it from `pre-commit`.


erwanp:
FYI there was a previous attempt at adding isort https://github.com/radis/radis/pull/58 , which eventually failed, probably because of semi-circular references ! Solving this would probably improve the loading performance of the package, but takes some time !

suzil:
radis (1)

erwanp:
Hello ! I investigated this one. Turns out there are many bad practices in our imports that could be fixed. 
These practices, however, were used for a few reasons which we want to keep : 

1. We want to import some classes from the root package level, e.g, `from radis import Spectrum` although Spectrum is defined in `radis.spectrum.spectrum.Spectrum`. 

This was achieved by importing many objects in the `__init__.py`, which are themselves loaded from the main [__init.py__](https://github.com/radis/radis/blob/develop/radis/__init__.py#L194). I do not know if there's a better practice for this. Usually the `from .spectrum import *` kind of imports are not. 

2. We want to be able to quickly execute the tests related to a module. 

This was historically achieved by adding test_functions() directly in the module, and running the module in the iPython console of the IDE. Then tests functions were moved in dedicated files, for instance [operations.py](https://github.com/radis/radis/blob/develop/radis/spectrum/operations.py) calls `from radis.test.spectrum.test_operations import _run_testcases` on execution of operations.py.  **Problem** : this workflow requires that operations.py is run as a script, not really as a module of a package, on execution. Therefore we end up with absolute imports such as [`from radis.spectrum import Spectrum`](https://github.com/radis/radis/blob/develop/radis/spectrum/operations.py#L58) in operations.py where we should have expected some `from .spectrum import Spectrum`. I think this is the root of all evils when it comes to later processing by isort. It probably also breaks down a lot of the autocompletion.  

How can we do better ? 

1. Maybe we could simply import the classes with an "absolute" path here, i.e., `from .spectrum.spectrum import Spectrum`. 

2. How can we tell an IDE to associate a module file with the corresponding test file ?

suzil:
> How can we tell an IDE to associate a module file with the corresponding test file ?

If we use `_run_testcases()`, then we miss out on the features that a test runner like `pytest` would provide.

For me personally, I usually will run `pytest` from the integrated terminal of my IDE. Sometimes I will point to a specific module that I want to be tested (`pytest radis/test/spectrum/test_compare.py`) or will use a fuzzy-match of test names with the `-k` option (`pytest -k get_residual`).

For IDE integration, I use VSCode with the [Python Test Explorer Extension](https://marketplace.visualstudio.com/items?itemName=LittleFoxTeam.vscode-python-test-adapter) that will allow me to point-and-click on the test to run and it will run the test with `pytest`.

Screen Shot 2020-12-06 at 10 04 57 AM

Or, alternatively to the sidebar, I can point-and-click above the test I want to run:

Screen Shot 2020-12-06 at 10 13 47 AM

In terms of passing arguments to `pytest` like the ones being passed to `_run_testcases()`, [this can be done with pytest](https://docs.pytest.org/en/6.0.1/example/simple.html#pass-different-values-to-a-test-function-depending-on-command-line-options) as customized command line options (and which options are enabled by default can be set in `setup.cfg` configuration).

It could be nicer to move test files to be included in a `tests/` in the directory that has the source code to make the test files easier to find.

Overall, not really sure if there's a way to tell the IDE to associate a module file with the corresponding test file rather than just having the human pick the right test file based on its name, but from personal experience this hasn't disrupted my workflow so long as the test file names correspond with the module names.

suzil:
> Problem : this workflow requires that operations.py is run as a script, not really as a module of a package, on execution. Therefore we end up with absolute imports such as from radis.spectrum import Spectrum in operations.py where we should have expected some from .spectrum import Spectrum. I think this is the root of all evils when it comes to later processing by isort. It probably also breaks down a lot of the autocompletion

I think this can't be causing the issue though. Absolute vs. relative imports shouldn't make any difference.

I did notice that moving `Spectrum` import in `radis/spectrum/__init__.py` from the top to the bottom breaks it, so will try to dig a bit deeper to see why exactly that's the case.

```diff
diff --git a/radis/spectrum/__init__.py b/radis/spectrum/__init__.py
index e50dd9a..4241366 100644
--- a/radis/spectrum/__init__.py
+++ b/radis/spectrum/__init__.py
@@ -3,7 +3,6 @@
 Spectrum-class module
 """
 
-from .spectrum import Spectrum, is_spectrum
 from .models import calculated_spectrum, experimental_spectrum, transmittance_spectrum
 from .compare import (
     plot_diff,
@@ -22,3 +21,4 @@ from .operations import (
     get_baseline,
     sub_baseline,
 )
+from .spectrum import Spectrum, is_spectrum
```

suzil:
I was able to fix the `Spectrum` circular dependency issue by changing an `__init__.py` import to an import directly to the respective module.

Currently struggling a bit with this circular dependency:
```py
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-1-e583f5296c1f> in <module>
----> 1 import radis

~/Documents/playground/radis/radis/__init__.py in <module>
    193 
    194 
--> 195 from .db import *  # database of molecules
    196 from .io import *  # input / output
    197 from .lbl import *  # line-by-line module

~/Documents/playground/radis/radis/db/__init__.py in <module>
      9 """
     10 
---> 11 from .molecules import Molecules, getMolecule

~/Documents/playground/radis/radis/db/molecules.py in <module>
     12 """
     13 
---> 14 from radis.db.classes import ElectronicState
     15 from radis.phys.convert import eV2cm
     16 

~/Documents/playground/radis/radis/db/classes.py in <module>
     53     get_herzberg_coefficients,
     54 )
---> 55 from radis.io.hitran import (
     56     HITRAN_CLASS1,
     57     HITRAN_CLASS5,

~/Documents/playground/radis/radis/io/__init__.py in <module>
    104 MOLECULES_LIST_NONEQUILIBRIUM = (
    105     __supported_molecules_nonequilibrium__
--> 106 ) = _get_supported_molecules_nonequilibrium()
    107 """ list: molecules that can be calculated in RADIS at nonequilibrium.
    108 Spectroscopic constants to calculate energy levels are needed.

~/Documents/playground/radis/radis/io/__init__.py in _get_supported_molecules_nonequilibrium()
     29     """
     30 
---> 31     from radis.db.molecules import Molecules
     32 
     33     return list(Molecules.keys())

ImportError: cannot import name 'Molecules'
```
Screen Shot 2020-12-06 at 12 10 59 PM

erwanp:
I can fix that by moving the `get_molecule` and `HITRAN_CLASS_` identifiers in the `radis/db` section. This would be like making the HITRAN classes the official RADIS partitioning of the database (which it is, anyway !), and leave `radis/io` with only the file parsing functions. 
If you can push to your branch I'll add these changes on top !

suzil:
Fix for `Spectrum` import pushed. 

To generate the diagrams, I'm using [pydeps](https://github.com/thebjorn/pydeps): `pydeps --show-cycles radis/`

suzil:
Let me rebase first actually!

suzil:
@erwanp all good now 🙂  ready for your fixes

erwanp (approved):

PR #141, opened 2020-12-05
Remove unnecessary future imports (only needed for py2 support)

suzil:
### Description

Remove unnecessary future imports. The ones imported are only needed for Python 2 support, which is no longer supported.

Refs: https://docs.python.org/3/library/__future__.html


erwanp (approved):
no future :)

Issue #151, opened 2020-12-06
Remove hapi.py from RADIS files, update to TIPS-2017


erwanp:
HAPI can be found on PyPi at hitran-api (see https://github.com/hitranonline/hapi/issues/2) 

We should : 
- remove it from the RADIS codebase (in io/hapi)
- make it a dependency
- and only call it for the tests and for TIPS ! 

Additional, it could be nice to write and publish a standalone, object-oriented HAPI extension that would mimic RADIS's calc_spectrum function input, compute it with HAPI (both emission & absorption), and return a RADIS Spectrum object (which all the associated post-processing functions), to make it easier to compare both codes. The [test_RADIS_vs_HAPI_line_broadening.py](https://github.com/radis/radis/blob/develop/radis/test/validation/test_RADIS_vs_HAPI_line_broadening.py#L74) validation case can serve as initial work.

erwanp:
Fixed by #152

PR #152, opened 2020-12-07
Remove hapi.py from RADIS files, update to TIPS-2017

suzil:
### Description
This requires upgrading the version of HAPI that RADIS is using. Before `hapi.py` was at version 1.1.0.6 but this version is not available on PyPI. As well, the earliest versions on PyPI do not support Python 3.6, so it would be good to go ahead and use the latest HAPI version if possible.

Some tests are failing due to the upgraded HAPI version it seems like.

Refs: #151


setup.py
```diff
@@ -149,6 +149,7 @@ def clean_line(line):
     ],
     packages=find_packages(),
     install_requires=[
+        "hitran-api",
```
erwanp, line 152: I think having it under https://github.com/radis/radis/blob/develop/environment.yml is enough ?

```suggestion
```

environment.yml
```diff
@@ -5,6 +5,7 @@ channels:
 - astropy
 dependencies:
 - python=3.6
+- hitran-api
```
erwanp, line 8:
```suggestion
```

(can't modify this directly, it's out of scope, sorry !)

hitran-api is not a Conda package so you have to install it under the pip: section below (l27).

erwanp (changes requested):

environment.yml
erwanp: Test failed with 
```
ResolvePackageNotFound: 

  - hitran-api
```
so that may be the reason !

suzil: Aha gotcha!

setup.py
suzil: Removed from `environment.yml` as it's not a conda package.

environment.yml
erwanp: You still have to add it in the environment.yml, under the pip section, i.e. 

```

- pip
- pip:
  - publib>=0.3.2  # Plotting styles for Matplotlib
  - hitran-api  👈
```

erwanp:
Hello ! RADIS now builds properly, but test fails and in particular hardcoded no-regression tests : this is caused by a change in Partition Functions that may have changed with the new HAPI version. This requires further investigation, I'll have a look this weekend !

erwanp:
Indeed, RADIS currently has HAPI `1.1.0.6` which uses TIPS-2011 for partitions functions, whereas HAPI `1.1.0.8` [introduced TIPS-2017](https://github.com/hitranonline/hapi/commit/2a12552364f0ac93e3f3bdfa7b3a9701a45d446b). 

Could thing I learned is that TIPS-2017 goes up to 5000 K for most main isotopes & species (and some even higher), while we were limited to 3000 K with TIPS-2011. 

Will be a nice addition !

erwanp (approved):

PR #142, opened 2020-12-05
Add stripping of trailing whitespace to pre-commit

suzil:
### Description

Add a pre-commit hook to strip trailing whitespace. It's commit noise, and this pre-commit hook will keep code consistent by stripping it always.

PR #178, opened 2021-01-09
Paper references in manual

dcmvdbekerom:
### Description

Fixes #<Issue Number>


erwanp:
Tests fail only because of syntax. I find that installing pre-commit is the easiest way to comply with #147  #142  etc. . 
```
cd radis
pre-commit install
```  
See here: https://radis.readthedocs.io/en/latest/dev/developer.html#code-linting

dcmvdbekerom:
pre-commit fails to run on my pc because it looks for python36, I'm running python 37... Any suggestions how to solve this?

Thanks!

erwanp:
We shouldn't be forced to use a particular Python version anyway  (as long as it's 3.x) : merging !

Issue #131, opened 2020-10-23
Faster rejection of irrelevant files


CorentinGrimaldi:
### 🎯 Describe what you'd like

A faster rejection of irrelevant files when loading a database for a calcul.

### 💡 Possible implementations

The rejection of irrelevant files (irrelevant because containing no information in the specified spectral range) is slow because radis load all the file before comparing the minimal and maximal wavenumber.
The idea is to store the minimal and maximal wavenumber in the metadata of the cache_file.
Thus allowing for future calculs to only load the metadata in order to know if the file is relevant for the specified calcul or not.
This could happens after the deprecated check.

This feature is a (huge) appreciable bonus when using large database such as CDSD-4000 and using "*" wildcard.

---
<sub>*Tip : you can vote for [New Features](https://github.com/radis/radis/issues/79) on https://feathub.com/radis/radis*</sub>

Issue #166, opened 2021-01-01
Improve database file loading / Read bz2 files directly


erwanp:
### 🎯 Describe what you'd like

Database loading is now the main performance bottleneck for equilibrium calculations. 
We need an efficient way to download & unzip only the required data for large databases on the first call, on load them extremely fast on successive calls.  

How we do that is quite open for discussion. 

- HITEMP currently ships all the database in .bz2 zip, and provide a Python script to unzip only parts of the data (see https://hitran.org/hitemp/ ). Wrote to @RomanKochanov and the HITRAN team to talk about databases. 

- we currently have an automatic .h5 cache file generation. From my own experience, .h5 files are very convenient because of the metadata associated with them (basically a Python dictionary) which is very useful to check for deprecation, but they're not as fast as pure .npy binaries . 

- Following @dcmvdbekerom initial code, the first GPU implementation #117 uses .npy files, which themselves had to be generated beforehand from the database. 

---

💬 Discussion : Slack channel [#databases](https://radis-radiation.slack.com/archives/C018C2MH5AB)

---

Note : the [radis-lab #1](https://github.com/radis/radis-lab/issues/1) experiment explores pre-installed HITEMP bz2 databases. Automatic bz2 parsing would at least make HITEMP computations very easy there !

erwanp:
I did some calculations to get insights on how to improve database loading & RAM use : see [radis-benchmark/manual_benchmarks/test_database_loading.ipynb](https://github.com/radis/radis-benchmark/blob/master/manual_benchmarks/test_database_loading.ipynb)

Key output : 

- HDF5 seem to be our best option
- However we should switch to Table format which allows SQL-like queries (much one order of magnitude faster than SQL).

erwanp:
Done in #194

PR #185, opened 2021-01-14
Faster rejection of irrelevant files

CorentinGrimaldi:
### Description

This pull request is to address #131 with a better file check relevancy by storing wmin and wmax in the cache file metadata. 
So we do not need to load all the cached file for nothing. Only the metadata are loaded and if the file is relevant for the calcul the file is loaded.
This allow for a much faster rejection of irrelevant file and speeding up the loading of large database such as CDSD-4000. Change from about 1s by file to 0.01s (approximately).

Fixes #<Issue Number>


radis/lbl/loader.py
```diff
@@ -1966,36 +1927,37 @@ def load_and_concat(files, buffer):
                         cache=db_use_cached,
                         verbose=verbose,
                         drop_non_numeric=True,
+                        wavenum_min=wavenum_min,
+                        wavenum_max=wavenum_max,
                     )
+                    if isinstance(df, str):
+                        if df == "IrrelevantFile":
```
erwanp, line 1934: Could be replaced by a larger

```
try:
    df = .... 
except IrrelevantFileError:
    if db_use_cached=='force':
        raise
     else:
        continue
``` 

`df` is not rewritten.

erwanp (changes requested):

erwanp:
Will help #166 too

radis/io/cdsd.py
```diff
@@ -116,7 +117,14 @@
...
 
 def cdsd2df(
-    fname, version="hitemp", count=-1, cache=False, verbose=True, drop_non_numeric=True
+    fname,
+    version="hitemp",
+    count=-1,
+    cache=False,
+    verbose=True,
+    drop_non_numeric=True,
+    wavenum_min=0.0,
```
erwanp, line 126: It's not straightforward what wavenum_min and max do here. Can you : 
- change the name to something explicit ? Ex : `load_only_wavenum_above=` ,  `load_only_wavenum_below=`? 
- add the names in the docstrings under Other Parameters with a shortdescription. In the docs, if needed, you can also refer to other functions like load_h5_cache_file using ``` :py:func`~radis.misc.cache_files` ```

erwanp (changes requested):

radis/lbl/loader.py
```diff
@@ -1898,104 +1899,60 @@ def load_and_concat(files, buffer):
...
-                        filename,
-                        cache=db_use_cached,
-                        verbose=verbose,
-                        drop_non_numeric=True,
-                    )
+                    except IrrelevantFileWarning:
+                        if db_use_cached=='force':
+                            raise
+                        else:
+                            continue
```
erwanp, line 1952: maybe add a low-level print here ? (if verbose==2 or 3  for instance ? )

radis/io/cdsd.py
```diff
@@ -233,14 +250,19 @@ def cdsd2df(
...
+                cache,
+                metadata=metadata,
+                current_version=radis.__version__,
+                last_compatible_version=OLDEST_COMPATIBLE_VERSION,
+                verbose=verbose,
+                load_only_wavenum_above=load_only_wavenum_above,
+                load_only_wavenum_below=load_only_wavenum_below,
+            )
+        except IrrelevantFileWarning as err:
+            raise err
```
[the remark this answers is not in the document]
CorentinGrimaldi: Ah yes, no need to do that, in my mind it was a way to transmit the warning to the upper function.
Will remove it.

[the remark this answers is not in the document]
CorentinGrimaldi: As in radis/io/hitran.py

erwanp (approved):

erwanp:
Very nice, merging ! 

@BlehMaks have a look at the PR, it can be useful to you : you may need to regenerate your cache files, but then it will make the whole loading process super fast even with the whole CDSD-4000 Database !

Issue #186, opened 2021-01-20
negative absorbance / too coarsed spectral grids 


erwanp:
### 🐛 Describe the bug

This error was detected by @Anjum48 in the [RADIS atmosphere example](https://github.com/radis/radis-examples) (see https://github.com/radis/radis-examples/issues/4)

```

  File "d:\github\radis\radis\lbl\factory.py", line 798, in eq_spectrum
    transmittance_noslit = exp(-absorbance)

RuntimeWarning: overflow encountered in exp
```

The unexpected thing is that the overflow only happens with the new LDM algorithm (on CPU), and not with the historical method. @dcmvdbekerom  

### 🖥 Steps To Reproduce

Here is a minimal example with conditions taken from the atmosphere example 

```

from radis import calc_spectrum
from warnings import catch_warnings, simplefilter

# Physical parameters                  
wmin = 400                       #: wavenumber min
wmax = 900                       #: wavenumber max
x_CO2 = 400e-6     #: float: mole fraction of CO2 

# Spectral Computation parameters
wstep = 0.003                    # wavenumber step (cm-1)

# %% Get Spectral line database
with catch_warnings():
    simplefilter('error', category=RuntimeWarning)
    calc_spectrum(wavenum_min=wmin, 
                 wavenum_max=wmax,
                 molecule='CO2',
                 isotope='1,2,3',
                 verbose=3,
                 wstep=wstep,
                 broadening_max_width=3,
                 warnings={'MissingSelfBroadeningWarning':'ignore'},
                 export_lines=False,
                 Tgas=234.25,
                 pressure=0.00646122,
                 path_length=100000,   #  cm
                 # optimization=None,      # if None, no error !
                 )
```

By investigating, we find out that the LDM method sometimes produces **negative** absorbance numbers (?) :

```
! %debug
! absorbance.min()
-680251.6244835826
```

which in turn raise an overflow.

dcmvdbekerom:
Interesting... I wonder if this is an issue with the LDM algorithm, or rather with the optimized weights of the algorithm.
Can you try the keyword ```optimization = "simple"``` and see if this solves the issue? (I don't have a computer with RADIS installed near me atm)

erwanp:
More update & further investigation :  

[...]

PR #190, opened 2021-01-24
Fixing negative lineshapes

dcmvdbekerom:
### Description

This pull request is to address https://github.com/radis/radis/issues/186.

With the new folding procedure the FFT lineshapes behave the same way as the conventional lineshapes when the linewidth is approximately equal to wstep.

The negative lineshapes are still an issue when `min-RMS` is used as optimization scheme, leading me to believe that it should _not_ be used as the default.

Fixes #<Issue Number>


radis/lbl/broadening.py
```diff
@@ -621,6 +611,34 @@ def voigt_lineshape(w_centered, hwhm_lorentz, hwhm_voigt, jit=True):
     return lineshape
 
 
+def voigt_FT(w_lineshape_ft, hwhmG, hwhmL):
+    """Fourier Transform of a Lorentzian lineshape
```
erwanp, line 615:
```suggestion
    """Fourier Transform of a Voigt lineshape
```
No?

erwanp (approved):
Nice! 

Is there any particular reason you inverted the nested loops  Gaussian first,  then Lorentzian during the database filling?

dcmvdbekerom:
> Is there any particular reason you inverted the nested loops Gaussian first, then Lorentzian during the database filling?

No particular reason, other than that that's what I'm used to, and also what I used in the paper and my [lightweight code](https://github.com/dcmvdbekerom/discrete-integral-transform/blob/master/demo/discrete_integral_transform.py)

Back in the days I got confused with the order so often that I settled on G,L because it is alphabetic. I kept this order to make it easier to drop in the working code, and as an additional benefit it made it easier to see which of the RADIS code I had checked and which I hadn't.

Long story short, the G,L order was just for my own convenience, we can easily keep it L,G like it was before.

erwanp:
Both options are fine. Merging!

PR #192, opened 2021-01-25
Improve LOS performances (SerialSlabs, etc.)

erwanp:
### Description

The radiative-transfer modules uses many checks to be robust under all conditions (different wavespace, units, etc...) . Some of these were recursively called, which ended up taking a lot of time for 100+ spectra 

Many other performance improvements to the LOS modules. 

### Examples

Extreme case with 800 spectra in the line-of-sight : 

```python

from numpy import zeros
from radis import calc_spectrum, SerialSlabs
from time import time
                  molecule='CO',
                  isotope='1,2,3',
                  pressure=1.01325,   # bar
                  Tgas=700,           # K
                  mole_fraction=0.1,
                  path_length=1,      # cm
                  wstep=0.01,
                  )

s_list =  [s.copy()]*800

t0 = time()
s_los = SerialSlabs(*s_list, modify_inputs=True)  # modify_inputs=True)
print(f'SerialSlabs in {time()-t0:.2f}s')
```

Before : 
```
>>> Spectrum calculated in 0.31s
>>> SerialSlabs in 310.86s
```

After:  
```
>>> Spectrum calculated in 0.41s
>>> SerialSlabs in 0.51s
```
So 600+ times faster :) 

Note that most of the calculation time is still spent on safety checks and verifications, as can be shown by comparing to a minimal example doing only the calculation part : 
```python
t0 = time()

I = zeros(len(s))
for si in s_list:
    ti = si.get('transmittance_noslit', copy=False)[1]
    ri = si.get('radiance_noslit', copy=False)[1]
    I = I * ti + ri

print(f'Simple LOS in {time()-t0:.2f}s')
``` 

```
>>> Simple LOS in 0.06s
```
So about 9 times faster. There's room for some improvement still !


erwanp:
@Anjum48  this improvement may help in your case (with ~ 90 layers I'd expect only about 2-3s were spent on useless recursive checks. It would only get bad (tens of seconds) beyond 200 layers. 
 
Also @ulyssed you'll definitly need this if you end up using SerialSlabs for radiative transfer with your 800+ slabs :)  

Will be added in 0.9.28, get the developer version if you need it now !

PR #206, opened 2021-02-23
Add the option 'load_all' in the Database object. 

gugzy:
This option allow to not load all the spectra at the database creation but only when accessed making the database creation almost instantaneous.

If useful, this option could be the default one for large database (number of files or database size tbd)

### DIFF
--- a/radis/db/molecules.py
+++ b/radis/db/molecules.py
@@ -177,13 +177,13 @@ def getMolecule(molecule, isotope=None, electronic_state=None, verbose=True):
     # molecule
     try:
         mol = Molecules[molecule]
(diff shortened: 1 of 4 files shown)


erwanp:
@gugzy summarizing the latest changes : 
- made lazy_loading the default mode to load a database
- automatically generate the summary .csv  if it's not there
- fixed some issues with float formats (were converted to float != float64 by read_csv), and isotope (converted to int instead of str)  that failed the get() functions
- fixes old tests with lazy_loading=True

Please check it still works correctly on your side! If so we can merge.

erwanp (approved):

PR #216, opened 2021-03-28
Fixes Rendering of Radis Flow Chart Image in LBL and cdsd4000_vs_hitemp_3409K Image in Use Case Section of Features (radis.readthedocs)

anandxkumar:
Fixes the following issue -
- The `cdsd4000_vs_hitemp_3409K` image didn't render before [here](https://radis.readthedocs.io/en/latest/features/features.html#use-cases)

Before 
Before1

After
After1

- The Radis-Flow-Chart didn't render [here](https://radis.readthedocs.io/en/latest/lbl/lbl.html#flow-chart)

Before
Before2

After
After2

- Added  `docs/auto_examples/` ,`docs/gen_modules/ `in .gitignore so that anyone can work on modifying the docs without worrying about deleting the cache files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,6 +78,8 @@ coverage.xml
 # Sphinx documentation
 docs/_build/
 docs/source/
+docs/auto_examples
+docs/gen_modules
 
 # PyBuilder
 target/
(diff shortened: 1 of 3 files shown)


erwanp:
Well done,  merging!

PR #217, opened 2021-03-29
fix: wavelength range mismatch

### LINKED ISSUE #214: Wavelength range of calc_spectrum != [wavelength_min, wavelength_max]
### 🐛 Describe the bug

When using `calc_spectrum` with wavelength as inputs, we specify `wavelength_min` and `wavelength_max`.

The output wavelengths are lower than the boundaries provided to  `calc_spectrum` due to their conversion to wavenumbers and back into wavelengths.

### 🖥 Steps To Reproduce

```python
s = calc_spectrum( wavelength_min=5200, # nm
                   wavelength_max=5250, 
                   molecule='CO',
                   isotope='1,2,3',
                   pressure=1.01325,   # bar
                   Tgas=300,           # K
                   mole_fraction=0.1,
                   path_length=1,      # cm
                   verbose=False
                  )
w, I = s.get('radiance_noslit', wunit="nm", Iunit="mW/sr/cm2/nm")
print(w.min(), w.max())
```

### 🎯 Expected behavior

w.min() == wavelength_min
w.max() == wavelength_max

### 💡 Possible solutions

Make a decision if input is provided as wavelength instead of wavenumbers.

gagan-aryan:
### Description

This pull request addresses lower output wavelengths compared to the boundaries provided in `calc_spectrum` due to conversions from wavenumber to wavelengths and vice-versa during calculations. 

Following up on the discussion in #214, [fetch_databank](https://github.com/radis/radis/blob/c421c67444d45f1d95062dd0a1e2201524965c8a/radis/lbl/loader.py#L796) is called upon Spectrum factory, and it seems to be expecting the values in wavenumbers. So, we do need to convert the wavelengths to wavenumbers taking into account the medium (?) 

There still seems to be a slight shift though, since the example below 

```
s = calc_spectrum( wavelength_min=5200, # nm
                   wavelength_max=5250, 
                   molecule='CO',
                   isotope='1,2,3',
                   pressure=1.01325,   # bar
                   Tgas=300,           # K
                   mole_fraction=0.1,
                   path_length=1,      # cm
                   verbose=False
                  )
w, I = s.get('radiance_noslit', wunit="nm", Iunit="mW/sr/cm2/nm")
print(w.min(), w.max())
```
prints `5199.973016620518 5250.00000015924`

Partially Fixes #214

### DIFF
--- a/radis/lbl/base.py
+++ b/radis/lbl/base.py
@@ -3228,7 +3228,7 @@ class BaseFactory(DatabankLoader):
 def get_waverange(
     wmin=None,
     wmax=None,
-    wunit=None,
+    wunit=Default('cm-1'),
     wavenum_min=None,
     wavenum_max=None,
     wavelength_min=None,
--- a/radis/lbl/base.py
+++ b/radis/lbl/base.py
@@ -3228,7 +3228,7 @@ class BaseFactory(DatabankLoader):
 def get_waverange(
     wmin=None,
     wmax=None,
-    wunit=Default('cm-1'),
+    wunit=Default("cm-1"),
     wavenum_min=None,
     wavenum_max=None,
     wavelength_min=None,
--- a/radis/lbl/calc.py
+++ b/radis/lbl/calc.py
@@ -22,8 +22,8 @@ from os.path import exists
 
 from radis.lbl.factory import SpectrumFactory
 from radis.misc.basics import all_in
-from radis.phys.convert import nm2cm
 from radis.phys.air import air2vacuum
+from radis.phys.convert import nm2cm
 
 
 # %%
--- a/radis/lbl/calc.py
+++ b/radis/lbl/calc.py
@@ -23,6 +23,7 @@ from os.path import exists
 from radis.lbl.factory import SpectrumFactory
 from radis.misc.basics import all_in
 from radis.phys.convert import nm2cm
+from radis.phys.air import air2vacuum
 
 
 # %%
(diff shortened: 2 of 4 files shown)


erwanp:
Well done. 👏   
 The slight shift is because of the spectral grid, it is expected : the difference between what user wants and what's calculated should be < `wstep`  (0.01 by default) .

Tests passed. Only [the code linting](https://travis-ci.com/github/radis/radis/jobs/494673252) failed. You may want to have a look at [Linting in the developer section](https://radis.readthedocs.io/en/latest/dev/developer.html#code-linting) to be able to fix it. 

Ideally, we should add an automatic test so the error is never reproduced again. Maybe add a simple case in [test_calc.py](https://github.com/radis/radis/blob/develop/radis/test/lbl/test_calc.py) : 
- test would run calc_spectrum  and assert than the min() max() objects are as required  (ex : np.isclose(..,..., atol=wstep) ) ?
- use one of the Test databases so you don't have to download data at test-time : use [HITRAN-CO-TEST](https://github.com/radis/radis/blob/c421c67444d45f1d95062dd0a1e2201524965c8a/radis/test/utils.py#L173) for instance, 2000-2300 cm-1 so you can compute a spectrum somewhere within 4348 - 5000 nm.  This would be done with : 
```
setup_test_line_databases()
calc_spectrum(..., 
                       databank="HITRAN-CO-TEST")
```
- Test should fail with current RADIS version, and pass with your correction.

erwanp:
This came from the main branch, was fixed this morning, not related to yours. I'm merging yours !!

PR #220, opened 2021-04-01
Improve appy_slit calculation time

CorentinGrimaldi:
### Description

Modified the convolution method from numpy.convolve() to scipy.signal.oaconvolve().
This is much faster for the convolution of a long array (the spectrum) with a short one (slit function).
Greatly improves the calculation time.

Also update a test with hardcoded references.

### DIFF
--- a/radis/test/lbl/test_calc.py
+++ b/radis/test/lbl/test_calc.py
@@ -128,7 +128,6 @@ def test_calc_spectrum(verbose=True, plot=True, warnings=True, *args, **kwargs):
         s.plot(wunit="nm")
 
     w, I = s.get("radiance", wunit="nm")
-    w_ref = w[::100]
     # Compare against hardcoded results (neq 0.9.22, 28/06/18)
     #        I_ref = np.array([0.28694463, 0.29141711, 0.32461613, 0.32909566, 0.21939511, 0.18606445,
     #                          0.19740763, 0.16948599, 0.16780345, 0.15572173, 0.16770853, 0.14966064,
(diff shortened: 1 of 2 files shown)


erwanp:
Nice addition ! 
Can you provide a small example of the performance difference here ?

For instance 
```
# calculate a spectrum 
                  molecule='CO',
                  isotope='1,2,3',
                  pressure=1.01325,   # bar
                  Tgas=700,           # K
                  mole_fraction=0.1,
                  path_length=1,      # cm
                  databank='hitran',  # or use 'hitemp'
                  )
# fake it to be very high resolution, with 3M spectral grid points : 
import numpy as np
s.resample(np.linspace(s.get_wavenumber().min(), s.get_wavenumber().max(),int(3e6)), 'cm-1')
```
Compute apply slit time : 
```
%timeit s.apply_slit(0.5, 'nm')
``` 
Before the modification I get 
```
>>> 5.98 s ± 129 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)     
```

CorentinGrimaldi:
Yes !
For the same case, now I get :

`>>> 692 ms ± 40.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`

Which is considerably faster !

erwanp:
Definitly, well done ! @dcmvdbekerom did you know of this convolution method ? 
(merging soon)

dcmvdbekerom:
Very interesting! Great find!
I've seen overlap-add mentioned some time ago but never gave it a second thought.. Happy to see it is this much faster!

This could have implications for the FFT method as well, although there we already benefit from the fact that we only need a single FFT instead of three, but still very much worth investigating...

dcmvdbekerom (approved):

dcmvdbekerom:
One comment on this, also related to the discussion over on the [slack channel](https://app.slack.com/client/T9E5R3R9S/C01TWEPTHFV), is that `np.convolve()` is just an efficient implementation of a direct convolution, i.e., it does not use the FT and therefore scales with N**2 and not NlogN.

So to estimate how much the overlap-add will improve the current FFT implementation, we should really compare it to `scipy.signal.fftconvolve()`. This may be a bit of a semantic discussion however since of course it does not take away the fact that switching overlap-add resulted in some real gains.

erwanp:
(comment on your comment : np.convolve scales as N*M because the two arrays can have a different size,  unlike in FFT.   That's why np.convolve is more efficient for a convolution of a small M kernel over a large N array (ex : a slit function). 

 Overlap-add brings the best of both worlds!)

PR #234, opened 2021-04-10
Fast Cython implementation of add_at()

### LINKED ISSUE #233: Replace np.add.at with cython implementation
### 🎯 Todo
For a large number of lines, the broadening step is currently limited by the filling of the LDM, which itself is bottlenecked by the `numpy.add.at()` operation. While numpy is much faster than pure Python loops, it is still much slower than an optimized compiled code.

I wrote a simple `np.add.at()` equivalent in Cyhton as follows:
```
import numpy as np
cimport numpy as np
import cython

@cython.boundscheck(False) 
@cython.wraparound(False) 
@cython.nonecheck(False)
def my_add_at(np.ndarray[np.float32_t, ndim=3] arr,
              np.ndarray[np.int32_t, ndim=1] k_arr,
              np.ndarray[np.int32_t, ndim=1] l_arr,
              np.ndarray[np.int32_t, ndim=1] m_arr,
              np.ndarray[np.float32_t, ndim=1] values):
    
    cdef unsigned int N = values.size
    cdef unsigned int i
    
    for i in range(N):
        arr[k_arr[i],l_arr[i],m_arr[i]] += values[i]

    return arr
```

I compared this to Numpy's `np.add.at()` with following script:

```
import numpy as np
from time import perf_counter
from cython_add import my_add_at

Nv = 300000
NG =  4
NL = 16

Ni = int(1e8)

I = np.random.rand(Ni).astype(np.float32)
k = np.random.randint(Nv, size = Ni, dtype=np.int32)
l = np.random.randint(NG, size = Ni, dtype=np.int32)
m = np.random.randint(NL, size = Ni, dtype=np.int32)

t0 = perf_counter()
LDM1 = np.zeros((Nv, NG, NL), dtype=np.float32)
np.add.at(LDM1, (k, l, m), I)
print(perf_counter() - t0)

t0 = perf_counter()
LDM2 = np.zeros((Nv, NG, NL), dtype=np.float32)
my_add_at(LDM2, k, l, m, I)
print(perf_counter() - t0)

print(np.sum(np.abs(LDM1-LDM2)))
```

Both functions return the same LDM array, but `np.add.at()` takes **13 s**, while the cython implementation takes only **1.6 s**, an improvement of **8x**, essentially for free!

dcmvdbekerom:
### Description
+ This PR implements a fast cython add_at()
+ Increased scipy version to 1.4.0 for oaconvolve compatibility

Fixes #233 

NOTE: some checks are failing on my system, but they also fail in the radis/develop branch


dcmvdbekerom:
Note that because of the try..except statement when loading the module, it is essentially skipped completely in the tests..

erwanp:
Is `python setup.py build_ext --inplace`  almost instantaneous if the project was built already? 

Also, did you try a numba implementation with  @jit? I have this for Whiting's, very fast while still pure Python code

erwanp:
Hello ! For code clarity, do you think we can move all the add_at Cython and Python try/except in [radis/misc/arrays.py](https://github.com/radis/radis/blob/develop/radis/misc/arrays.py) ? There already are some Numpy extensions re-implementations there. 

In `broadening.py` we could have only : 
```
from radis.misc.arrays import add_at
```

erwanp (changes requested):
see https://github.com/radis/radis/pull/234#issuecomment-817978328

environment.yml
```diff
@@ -6,7 +6,7 @@ channels:
 dependencies:
 - python=3.6
 - numpy
-- scipy
+- scipy>=1.4.0
 - matplotlib
```
erwanp, line 10:
```suggestion
- matplotlib
- cython
```

Add cython directly ?

erwanp (changes requested):

environment.yml
dcmvdbekerom: I was hesitant to do this because I wasn't sure if installing the cython module also installs a c-compiler. I suppose we can give it a go and see if it causes problems?

erwanp: Yes ! We already have Numba as a requirement (for Whiting's) so there must be some compilers required there too !

dcmvdbekerom:
Travis uses GCC for compilation whereas I use MSVC, and they take different command line arguments. In a way we're lucky that me and Travis use different compilers, because now we know the two compilers need different arguments. 

Solution is relatively simple but first I need to make Cython use GCC on my system so I can reproduce the problem.

erwanp:
Only thing missing : 2 tests! 

Unitary tests that add_add works as expected, and maybe a test that Cython-enabled spectra also yield the same output as non Cython (its possible with my latest changes to deactivate Cython after imports, for the sake of the tests)

dcmvdbekerom:
Note that there were two booleans with different functions before: `sf.use_cython` to indicate which one it _should_ use, and `sf.params.use_cython` which reflects which it _actually_ used. To prevent confusion I changed the latter into `sf.params.add_at_used` which can be `"numpy"` or `"cython"`.

Also note that currently, `sf.params.add_at_used` is not listed in the printed computational params for some reason.

erwanp (approved):

erwanp:
Nice ! 

I checked why it didn't appear ; turns out it appears properly if you print the Spectrum : 
```
print(s)
```
```
... 

Computation Parameters
   Tref                 296 K
   add_at_used          numpy
```

But does not appear in the initial print of the calculation conditions by SpectrumFactory, which is done a priori, without knowing if Cython was compiled yet. 


Merging, this will be a nice addition !

Issue #118, opened 2020-08-07
Reduce Memory usage


erwanp:
Radis is quite memory hungry but there is room for improvement. 

## Still using Pandas :

Some columns do not need to be loaded: 

- [ ] Linestrengths "int" can be recomputed from Einstein coefficients at runtime, or the other way around. 
- [x] molecule `id` is not needed as we anyway compute only one molecule at a time. Same goes for isotope if only one isotope ? 
- [ ] A good starting point may be to list all the columns present in sf.df0 and sf.df1 at the end of a SpectrumFactory calculation, and remove all that are not needed. 

Also : 

- [x] do not fill a full column "Qgas" with partition functions  (an internal dictionary could be more than enough). Same for all molecular parameters (mass, etc. )  #287 

- [ ] Some column formats can be reduced (float64 > float32, int64 > uint8) with very limited impact, or formats such as "categorical" can be used. 

- [ ] the `.df0` "buffer database" can be fully removed and replaced by a .h5 file : #176 
> a very limited H5 file for each calculation could totally replace the current mechanism where the SpectrumFactory.df0 Pandas dataframe containing all lines loaded, is copied at calculation time to SpectrumFactory.df1 (where populations will be computed based on temperature, etc.). Using a .h5 file may induce a (limited ?) performance overhead, but we could also divide the memory usage by up to 2

---

## Moving away from Pandas 

Some other libraries may be more adapted to 'Big Data' (which I define here as : doesn't fit in RAM) 

- [vaex](https://towardsdatascience.com/how-to-analyse-100s-of-gbs-of-data-on-your-laptop-with-python-f83363dda94)    
- Dask

> Dask is more focused on scaling the code to compute clusters, while Vaex makes it easier to work with large datasets on a single machine. [from here](https://towardsdatascience.com/are-you-still-using-pandas-to-process-big-data-in-2021-850ab26ad919)

Let's try them ? 

---

This is discussed on [Slack](https://radis.github.io/slack-invite/) in channel [#memory_usage](https://radis-radiation.slack.com/archives/C0193Q836N4)

Anyone wanting to work on that topic is very welcome !!

---

Check memory consumption in [radis-benchmark](https://radis.github.io/radis-benchmark/)

Crash test / suggested test case : 

- read all HITEMP CH4 database in memory : 
```
31 880 412 lines,  0-13400cm-1 , 5 GB .par file,  1.9 GB h5-compressed file.
```
[...]

PR #252, opened 2021-04-19
Set load_energies=False and updated Error message

anandxkumar:
### Description

This pull request is to set the  `load_energies` parameter to False by default in Spectrum Factory and raise error to set `load_energies=True` if the user wants to calculate non_equilibrium spectrum using Spectrum Factory.

### DIFF
--- a/radis/lbl/base.py
+++ b/radis/lbl/base.py
@@ -1000,9 +1000,6 @@ class BaseFactory(DatabankLoader):
 
         # Check energy levels are here
         for iso in self._get_isotope_list():
-            print(
-                "molecule - ", molecule, "\nisotope - ", iso, "\nelec-state - ", state
-            )
             if not iso in self.get_partition_function_molecule(molecule):
                 raise AttributeError(
                     "No Partition function calculator defined for isotope {0}".format(
--- a/radis/lbl/base.py
+++ b/radis/lbl/base.py
@@ -623,17 +623,24 @@ class BaseFactory(DatabankLoader):
             t0 = time()
 
         # Check energy levels are here
-        for iso in self._get_isotope_list():
-            if not iso in self.parsum_calc["CO2"]:
-                raise AttributeError(
-                    "No Partition function calculator defined for isotope {0}".format(
-                        iso
+        try:
+            for iso in self._get_isotope_list():
+                if not iso in self.parsum_calc["CO2"]:
+                    raise AttributeError(
+                        "No Partition function calculator defined for isotope {0}".format(
+                            iso
+                        )
+                        + ". You need energies to calculate a non-equilibrium spectrum!"
+                        + " Fill the levels parameter in your database definition, "
+                        + " with energies of known format: {0}".format(KNOWN_LVLFORMAT)
+                        + ". See SpectrumFactory.load_databank() help for more details"
                     )
-                    + ". You need energies to calculate a non-equilibrium spectrum!"
-                    + " Fill the levels parameter in your database definition, "
-                    + " with energies of known format: {0}".format(KNOWN_LVLFORMAT)
-                    + ". See SpectrumFactory.load_databank() help for more details"
-                )
+        except:
+            raise KeyError(
+                "Error while calculating the non-equilibrium spectrum!"
+                + " Load the energies levels with SpectrumFactory.load_databank"
+                + "('path', load_energies=True)"
+            )
 
         def get_Evib_CDSD_pcN_1iso(df, iso):
             """Calculate Evib for a given isotope (energies are specific to a
@@ -999,17 +1006,24 @@ class BaseFactory(DatabankLoader):
             t0 = time()
 
         # Check energy levels are here
-        for iso in self._get_isotope_list():
-            if not iso in self.parsum_calc[molecule]:
-                raise AttributeError(
-                    "No Partition function calculator defined for isotope {0}".format(
-                        iso
+        try:
+            for iso in self._get_isotope_list():
+                if not iso in self.parsum_calc[molecule]:
+                    raise AttributeError(
+                        "No Partition function calculator defined for isotope {0}".format(
+                            iso
+                        )
+                        + ". You need energies to calculate a non-equilibrium spectrum!"
+                        + " Fill the levels parameter in your database definition, "
+                        + " with energies of known format: {0}".format(KNOWN_LVLFORMAT)
+                        + ". See SpectrumFactory.load_databank() help for more details"
                     )
-                    + ". You need energies to calculate a non-equilibrium spectrum!"
-                    + " Fill the levels parameter in your database definition, "
-                    + " with energies of known format: {0}".format(KNOWN_LVLFORMAT)
-                    + ". See SpectrumFactory.load_databank() help for more details"
-                )
+        except:
+            raise KeyError(
+                "Error while calculating the non-equilibrium spectrum!"
+                + " Load the energies levels with SpectrumFactory.load_databank"
+                + "('path', load_energies=True)"
+            )
 
         def get_Evib_RADIS_cls1_1iso(df, iso):
             """Calculate Evib & Erot for a given isotope.
--- a/radis/lbl/base.py
+++ b/radis/lbl/base.py
@@ -468,7 +468,7 @@ class BaseFactory(DatabankLoader):
 
         # Check energy levels are here
         for iso in self._get_isotope_list(molecule):
-            if not iso in self.parsum_calc["CO2"]:
+            if not iso in self.get_partition_function_molecule(molecule):
                 raise AttributeError(
                     "No Partition function calculator defined for isotope {0}".format(
                         iso
@@ -623,24 +623,17 @@ class BaseFactory(DatabankLoader):
             t0 = time()
 
         # Check energy levels are here
-        try:
-            for iso in self._get_isotope_list():
-                if not iso in self.parsum_calc["CO2"]:
-                    raise AttributeError(
-                        "No Partition function calculator defined for isotope {0}".format(
-                            iso
-                        )
-                        + ". You need energies to calculate a non-equilibrium spectrum!"
-                        + " Fill the levels parameter in your database definition, "
-                        + " with energies of known format: {0}".format(KNOWN_LVLFORMAT)
-                        + ". See SpectrumFactory.load_databank() help for more details"
+        for iso in self._get_isotope_list():
+            if not iso in self.get_partition_function_molecule(molecule):
+                raise AttributeError(
+                    "No Partition function calculator defined for isotope {0}".format(
+                        iso
                     )
-        except:
-            raise KeyError(
-                "Error while calculating the non-equilibrium spectrum!"
-                + " Load the energies levels with SpectrumFactory.load_databank"
-                + "('path', load_energies=True)"
-            )
+                    + ". You need energies to calculate a non-equilibrium spectrum!"
+                    + " Fill the levels parameter in your database definition, "
+                    + " with energies of known format: {0}".format(KNOWN_LVLFORMAT)
+                    + ". See SpectrumFactory.load_databank() help for more details"
+                )
 
         def get_Evib_CDSD_pcN_1iso(df, iso):
             """Calculate Evib for a given isotope (energies are specific to a
@@ -741,7 +734,7 @@ class BaseFactory(DatabankLoader):
 
         # Check energy levels are here
         for iso in self._get_isotope_list(molecule):
-            if not iso in self.parsum_calc["CO2"]:
+            if not iso in self.get_partition_function_molecule(molecule):
                 raise AttributeError(
                     "No Partition function calculator defined for isotope {0}".format(
                         iso
@@ -1006,24 +999,20 @@ class BaseFactory(DatabankLoader):
             t0 = time()
 
         # Check energy levels are here
-        try:
-            for iso in self._get_isotope_list():
-                if not iso in self.parsum_calc[molecule]:
-                    raise AttributeError(
-                        "No Partition function calculator defined for isotope {0}".format(
-                            iso
-                        )
-                        + ". You need energies to calculate a non-equilibrium spectrum!"
-                        + " Fill the levels parameter in your database definition, "
-                        + " with energies of known format: {0}".format(KNOWN_LVLFORMAT)
-                        + ". See SpectrumFactory.load_databank() help for more details"
-                    )
-        except:
-            raise KeyError(
-                "Error while calculating the non-equilibrium spectrum!"
-                + " Load the energies levels with SpectrumFactory.load_databank"
-                + "('path', load_energies=True)"
+        for iso in self._get_isotope_list():
+            print(
+                "molecule - ", molecule, "\nisotope - ", iso, "\nelec-state - ", state
             )
+            if not iso in self.get_partition_function_molecule(molecule):
+                raise AttributeError(
+                    "No Partition function calculator defined for isotope {0}".format(
+                        iso
+                    )
+                    + ". You need energies to calculate a non-equilibrium spectrum!"
+                    + " Fill the levels parameter in your database definition, "
+                    + " with energies of known format: {0}".format(KNOWN_LVLFORMAT)
+                    + ". See SpectrumFactory.load_databank() help for more details"
+                )
 
         def get_Evib_RADIS_cls1_1iso(df, iso):
             """Calculate Evib & Erot for a given isotope.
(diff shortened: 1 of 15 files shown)


anandxkumar:
@erwanp, I changed the load_energies to False when loading the database. This fixed the Spectrum Factory issue we were facing earlier when selecting molecules that didn't support non-equilibrium mode. But as the default value was set to False, the `calc_spectrum()` at non-equilibrium conditions fails when `using databank('hitemp'/'hitran) `in calc_spectrum. The issue is caused by the  `**kwargs ` parameter. It is probably because the parameter load_energies remains False throughout the whole process and previously for this case it was set to True and hence spectrum was calculated without any error. 
Now we will have to set the load_energies parameter explicitly to` True `if non-equilibrium, and I guess the parameter will be in `**kwargs`, I tried modifying it but didn't work.
When using the database from radis.json it works fine as we are explicitly setting `load_energies=not _equilibrium`.
Following is a case that is failing -
```
calc_spectrum(
        wavenum_min=2000 ,
        wavenum_max=2300 ,
        molecule="CO",
        isotope="1,2",
        Tvib = 1500,
        Trot = 300,
        databank="hitemp",  # test by fetching directly
    )
```

<details>
<summary>Traceback :</summary>
radis/test/io/test_hitemp.py:154: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
radis/lbl/calc.py:439: in calc_spectrum
    **kwargs_molecule
radis/lbl/calc.py:670: in _calc_spectrum
    name=name,
radis/lbl/factory.py:1263: in non_eq_spectrum
    self._calc_noneq_parameters(vib_distribution, singleTvibmode)
radis/lbl/base.py:1481: in _calc_noneq_parameters
    calc_Evib_harmonic_anharmonic=calc_Evib_harmonic_anharmonic,
radis/lbl/base.py:371: in _add_EvibErot
    df, calc_Evib_harmonic_anharmonic=calc_Evib_harmonic_anharmonic
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <radis.lbl.factory.SpectrumFactory object at 0x7f6725ea0f60>
df =       id  iso          wav  ...  molar_mass   ju            Eu
0      5    1  1995.038131  ...   27.994915   42  25235...  28.998270  147  63532.057312
5400   5    1  2304.961081  ...   27.994915  110  76835.461881

[5401 rows x 20 columns]
calc_Evib_harmonic_anharmonic = False

    def _add_EvibErot_RADIS_cls1(self, df, calc_Evib_harmonic_anharmonic=False):
        """Fetch Evib & Erot in dataframe for HITRAN class 1 (diatomic)
        molecules.
    
        Parameters
        ----------
    
        df: DataFrame
            list of transitions
    
        Other Parameters
        ----------------
    
        calc_Evib_harmonic_anharmonic: boolean
            if ``True``, calculate harmonic and anharmonic components of
            vibrational energies (for Treanor distributions)
        """
        if __debug__:
            printdbg(
                "called _add_EvibErot_RADIS_cls1(calc_Evib_harmonic_anharmonic={0})".format(
                    calc_Evib_harmonic_anharmonic
                )
            )
    
        if calc_Evib_harmonic_anharmonic:
            raise NotImplementedError
    
        molecule = self.input.molecule
        state = self.input.state  # electronic state
        # TODO: for multi-molecule mode: add loops on molecules and states too
    
        if self.verbose >= 2:
            t0 = time()
    
        # Check energy levels are here
        for iso in self._get_isotope_list():
>           if not iso in self.parsum_calc[molecule]:
E           KeyError: 'CO'

radis/lbl/base.py:1003: KeyError
----------------------------- Captured stdout call -----------------------------
Using existing database HITEMP-CO
Calculating Non-Equilibrium Spectrum
Physical Conditions
   Tgas                 300 K
   Trot                 300 K
   Tvib                 1500 K
   isotope              1,2
   mole_fraction        1
   molecule             CO
   path_length          1 cm
   self_absorption      True
   state                X
   wavenum_max          2300.0000 cm-1
   wavenum_min          2000.0000 cm-1
Computation Parameters
   Tref                 296 K
   broadening_max_width  10 cm-1
   broadening_method    fft
   dbformat             hitemp-radisdb
   dbpath               /home/pipebomb/.radisdb/CO-05_HITEMP2019.h5
   dlm_log_pG           0.1375350788016573
   dlm_log_pL           0.20180288881201608
   folding_thresh       1e-06
   levelsfmt            radis
   optimization         min-RMS
   parfuncfmt           hapi
   pseudo_continuum_threshold  0
   wavenum_max_calc     2305.0100 cm-1
   wavenum_min_calc     1995.0000 cm-1
   waveunit             cm-1
   zero_padding         -1
Information
   db_use_cached        True
   export_lines         False
   lvl_use_cached       True
   total_lines          5401
Fetching Evib & Erot.

</details>

erwanp:
Doesn't it work if you modify the dictionary of **conditions given here: https://github.com/radis/radis/blob/develop/radis/lbl/calc.py#L573 ?

anandxkumar:
Yes, it worked! Right now I will have to update all the non-equilibrium tests with load_energies= True, can we do one thing -
For `CO` and `CO2`, we load the energy levels and for the rest of the molecules we leave it as the default setting i.e. load_energies=False?

erwanp:
It's a idea, and you could even generalize it easily by checking if the molecule is in the list of supported molecules at nonequilibrium : [MOLECULES_LIST_NONEQUILIBRIUM ](https://github.com/radis/radis/blob/develop/radis/db/__init__.py#L99)

However, we will still be calculating (long !) and loading (takes Memory, see #118 ) energies if computing an equilibrium spectrum only. Fortunately, calc_spectrum() handles that for you, but people doing batch calculations  (for instance, to compute [atmospheres, see GSOC 2021](https://openastronomy.org/gsoc/gsoc2021/#/projects?project=atmosphere_calculations) )  will end up using SpectrumFactory, and most of them will want equilibrium calculations only. 

A better option would be to not load them by default for all molecules [unless said otherwise in load/fetch_databank, for instance for tests !], and postpone the loading of rovibrational energies (the Energy Database) when we know we need it, i.e., just before computing nonequilibrium spectra. 
[noneq_spectrum()](https://github.com/radis/radis/blob/develop/radis/lbl/factory.py#L1263) already has a `self._calc_noneq_parameters()` function, which adds rovibrational energies from the Energy Database into the Line Database. You could edit this function, make sure the Energy Database is loaded, and try to load it from there before failing. 

Note that Energy Database and Line Database are really two separate inputs (cf [Architecture](https://radis.readthedocs.io/en/latest/dev/architecture.html))

anandxkumar:
@erwan suppose for CO2 eq calculation, I use `sf.load_databank('HITEMP-CO2-TEST')`.
Now it won't load the energy data. If I calculate non-equilibrium `sf.non_eq_spectrum(Tvib=2000,Trot=300)` it will throw me an error. So if we see the easiest approach we just need to call `self.load_databank('HITEMP-CO2-TEST', load_energies = True)` in `non_eq_spectrum()` before `_calc_noneq_parameters()`. We can check the load_energies value by `self.misc.load_energies`.
If `self.misc.load_energies==True`, we proceed as expected and if `self.misc.load_energies==False` we set is as True and load the databank similarly. But I'm not sure how to replicate the same thing in `non_eq_spectrum()`. How will get to know about the databank name because it's not present in `self.get_conditions()`? For other cases, we will have to implement a similar approach as `calc_spectrum` for the non-equilibrium part. So we will be calling calc_spectrum functionality in Spectrum Factory which will kind of ruin the purpose of Spectrum Factory. If you have a better approach let me know.

erwanp:
Warning : we shouldn't be implementing calc_spectrum in SpectrumFactory : `calc_spectrum` is the high-level interface, `SpectrumFactory` is the low-level interface for advance usages, they shouldn't be mixed. 

In your example, `self._calc_noneq_parameters()` should check that energies are not loaded, and load them. Basically, [the ending lines of load/fetch_databank()](https://github.com/radis/radis/blob/41cb85bbc0b040b73ea17266c925628e2e61a52e/radis/lbl/loader.py#L1234) should **also** be added in [_calc_noneq_parameters](https://github.com/radis/radis/blob/develop/radis/lbl/base.py#L1417), in case energies were not loaded.

You'll need the values of `levels` [path] and `levelsfmt` [format]. They are stored initially by [_store_database_params()](https://github.com/radis/radis/blob/41cb85bbc0b040b73ea17266c925628e2e61a52e/radis/lbl/loader.py#L1393) under ``self.levelspath`` and ``self.params.levelsfmt``

anandxkumar:
@erwanp, I modified `factory.py` to load Energy Level Database when `load_energy=False` and we are calculating non-eq spectrum. I stored `self.misc.load_energy` in `_store_database_params` to check the status of load_energy and load Energy Level only if `self.misc.load_energy==False`. I tested it and it's working fine for all the test cases. I found the places where we are encountering errors and modified it with `try except` statement to make the user aware to set `load_energy=True`. Some tests were using non-eq parameters before calculating the non_eq spectrum like - 
`test_calc.py` -  `test_all_calc_methods_CO2pcN()`

```
    sf.load_databank(**database_kwargs)
    # Now, define Evib:
    .............
    .............
    lvl.non_eq_spectrum(Tvib=Tgas, Trot=Tgas)  
```
In this case we will have to load the energy levels, so I modified it accordingly. 
Also I found that that the SpecDatabase [link](https://radis.readthedocs.io/en/master/source/radis.tools.database.html)
Wrong code is written `from radis.lbl import SpecDatabase`, I changed it to from `radis.tools import SpecDatabase`

radis/lbl/factory.py
```diff
@@ -1258,6 +1258,11 @@ def non_eq_spectrum(
         # Check line database and parameters, reset populations and scaled line dataframe
         # ----------
         self._check_line_databank()
+        # Checks and loads Energy level database
+        if self.misc.load_energies == False:
+            self._init_rovibrational_energies(self.levels, self.params.levelsfmt)
+            self.misc.load_energies = True
+
```
erwanp, line 1265: - If I remember well, there is an (underused) non_eq_bands() function in bands.py that would require a similar treatment (it does the same thing as non_eq_spectrum, but only computes some spectral bands, before combining them). 
- same for the `SpectrumFactory.optically_thin_power()` method (also underused : instead of computing a full spectrum it only gives the integral assuming some hypothesis that make the calculation 1000x faster, i.e., that you don't need to resolve every single line) : it also has non-eq options and therefore will fail (the fact it didn't shows there aren't tests there, that's another issue ! :) )

Instead of copying the code here there, what about moving these lines in `self._calc_noneq_parameters()` , which is probably called by all these functions I mentioned ?

radis/lbl/base.py
```diff
@@ -999,17 +1006,24 @@ def _add_EvibErot_RADIS_cls1(self, df, calc_Evib_harmonic_anharmonic=False):
...
 
         # Check energy levels are here
-        for iso in self._get_isotope_list():
-            if not iso in self.parsum_calc[molecule]:
-                raise AttributeError(
-                    "No Partition function calculator defined for isotope {0}".format(
-                        iso
+        try:
+            for iso in self._get_isotope_list():
+                if not iso in self.parsum_calc[molecule]:
```
erwanp, line 1011: Works, but I find it hard to read, as we're testing two things at the same time (loading of energies, and existence for a particular isotope).

What about replacing : 
```
                if not iso in self.parsum_calc[molecule]:
```
with  : 
```
if not iso in get_partition_function_calculator(self, molecule, isotope, "X")
```
(here "X" refers to the main electronic state, it should be defined this way).

If energies haven't been loaded, we'll trigger your new error message from loader.py). No need for the try/except in base.py anymore.

radis/lbl/base.py
```diff
@@ -1089,17 +1103,24 @@ def _add_Evib123Erot_RADIS_cls5(self, df):
...
 
         # Check energy levels are here
-        for iso in self._get_isotope_list(molecule):
-            if not iso in self.parsum_calc[molecule]:
-                raise AttributeError(
-                    "No Partition function calculator defined for isotope {0}".format(
-                        iso
+        try:
+            for iso in self._get_isotope_list(molecule):
+                if not iso in self.parsum_calc[molecule]:
```
erwanp, line 1108: same as above!

erwanp (changes requested):
Lot's of changes, nice ! 

I have a couple suggestions for simplification, and maybe some potential errors to avoid (calculations ran from non_eq_bands & optically_thin_power() )

radis/lbl/factory.py
anandxkumar: Actually the test for `SpectrumFactory.optically_thin_power()`  exists in `test_factory.py`  in `test_power_integral()`. I was actually getting error, so I loaded the energy level in that particular test -
sf.load_databank("HITRAN-CO-TEST", db_use_cached=True, `load_energies=True`).
I added the code in `self._calc_noneq_parameters()` and it's working with the default setting. It's also working for `non_eq_bands()`  :+1: 
But there is one case we need to consider, at some places we are accessing the parameters like  `self.parsum_calc`, at some places/tests  like `parfunc = sf.parsum_calc["CO2"][iso]["X"]`  and also some places like [link](https://github.com/radis/radis/blob/c6f544bc07b584389871d7b1a39469cdeb47e537/radis/test/lbl/test_overp.py#L194) without executing `self._calc_noneq_parameters()`. In these cases, we will get errors. So instead of displaying the error, should we automatically run `_init_rovibrational_energies`  based on `load_energies`?

erwanp: In your latest link,  self.get_partition_function_calculator is called anyway, so it will trigger an appropriate error, quite clear for the end-user !

radis/lbl/base.py
anandxkumar: Actually, there is one unusual error I'm facing while doing this. For Test - `test_calc_hitemp_CO_noneq` we are computing - 
```
calc_spectrum(
    wavenum_min=2000 ,
    wavenum_max=2300 ,
    molecule="CO",
    isotope="1,2",
    Tvib=1500,
    Trot=300,
    databank="hitemp",  # test by fetching directly
)
```
I'm getting a Type Error - 
```
molecule = self.input.molecule
state = self.input.state  # electronic state

for iso in self._get_isotope_list():
>               if not iso in self.get_partition_function_calculator(self, molecule, iso, state):
E               TypeError: get_partition_function_calculator() takes 4 positional arguments but 5 were given

radis/lbl/base.py:1011: TypeError
```
When printing the values of all parameters we get - 
```
molecule -  CO 
isotope -  1 
elec-state -  X
```
Which seems appropriate, what can be causing the additional argument?
There is a TODO for the above - 
# TODO: for multi-molecule mode: add loops on molecules and states too. This can justify the cause of the error, but the parameters are single values only in the above example, so what can be the culprit?

erwanp: My bad, I wrongly copy-pasted the function name. Should be:  
```
if not iso in self.get_partition_function_calculator(molecule, iso, state):
```
not 
```
if not iso in self.get_partition_function_calculator(**self**, molecule, iso, state):<
```

radis/lbl/factory.py
anandxkumar: Actually, I meant that instead of raising the error we can load the energy levels instead using in the except statement by running `_init_rovibrational_energies()`, if we still encounter an error we can raise it.
The error raised by `get_partition_function_calculator()` will work but there are some cases where we are directly accessing like [link](https://github.com/radis/radis/blob/c6f544bc07b584389871d7b1a39469cdeb47e537/radis/test/levels/test_partfunc.py#L708)
Here also we will get a `Key Error` but this won't be raised by `get_partition_function_calculator()`. I solved it by loading the energy level database. But how will we trigger an appropriate error for this case?

radis/lbl/base.py
anandxkumar: This time I got the following error -
```
for iso in self._get_isotope_list():
                print("molecule - ", molecule, "\nisotope - ",iso, "\nelec-state - ", state)
>               if not iso in self.get_partition_function_calculator(molecule, iso, state):
TypeError: argument of type 'PartFunc_Dunham' is not iterable
radis/lbl/base.py:1012: TypeError
```
This may be due to the fact that we want -
if not iso in `self.parsum_calc[molecule]`

But `self.get_partition_function_calculator(molecule, iso, state)` returns `self.parsum_calc[molecule][isotope][elec_state]`. We actually need the superset of parsum ie. `self.parsum_calc[molecule]` which can't be derived from a subset `self.parsum_calc[molecule][isotope][elec_state]`, which could be true the other way around.

erwanp: You're right, let's keep it as you suggested initially, unless you see another way!

anandxkumar:
@erwanp, I implemented your suggestions. For raising error in `base.py`, instead of applying `try except` in all functions, I created a function in `loader.py` , `get_partition_function_molecule(self, molecule)` to get `self.parsum_calc[molecule]`. 
Is there anything else I need to consider?

radis/lbl/loader.py
```diff
@@ -2260,14 +2264,37 @@ def get_partition_function_calculator(self, molecule, isotope, elec_state):
...
-        parsum = self.parsum_calc[molecule][isotope][elec_state]
-
+        try:
+            parsum = self.parsum_calc[molecule][isotope][elec_state]
+        except KeyError as err:
+            raise KeyError(
+                "Error while Retrieving Partition Function Calculator!"
+                + " Load the energies levels with SpectrumFactory.load_databank"
+                + "('path', load_energies=True)"
+            ) from err
```
erwanp, line 2274: If energies haven't been loaded, it should fail on the first key error. We could use your new `get_partition_function_molecule` function to make it clearer.

```suggestion
        parsum = self.get_partition_function_molecule(molecule)[isotope][elec_state]
```
(will fail properly with the right error)

erwanp (changes requested):

radis/lbl/factory.py
erwanp: Hello ! In the above link `sf.parsum_calc["CO2"][1]["X"].df` is only accessed for test purpose, in a way that's used only by people knowledgable of the code structure. Therefore it's ok if the error is not very clear. You should just enforce load_energies=True in the test itself.

radis/lbl/loader.py
anandxkumar: Nice Suggestion. Implemented it and updated the error message. Thanks :+1: !

erwanp (approved):

PR #257, opened 2021-04-21
fix reading of single path in config file entry

CorentinGrimaldi:
### Description

Fixes #<Issue Number>

### DIFF
--- a/radis/misc/config.py
+++ b/radis/misc/config.py
@@ -787,7 +787,7 @@ def getDatabankEntries(dbname, get_extra_keys=[]):
 
     # Path correction for all tests
     if type(entries["path"]) == str:
-        entries["path"] = entries["path"].split()
+        entries["path"] = [entries["path"]]
 
     # Merge all isotope-dependant levels into one dict
     iso_list = [k for k in entries if k.startswith("levels_iso")]


erwanp:
@anandxkumar do you remember why this line was introduced with split()?  It created some problems with paths containing spaces.

anandxkumar:
@erwanp, if we look at the previous function `getDatabankEntries_configformat(dbname, get_extra_keys=[])`, we had
`entries["path"] = entries["path"].strip("\n").split("\n")`, this basically returned a path or paths in form of a list so even if we had an entry like -

```
[HITRAN-CO2-TEST]
info = HITRAN 2016 database, CO2, 1 main isotope (CO2-626), bandhead: 2380-2398 cm-1 (4165-4200 nm)
path = d:\github\radis\radis\test\files\hitran_co2_626_bandhead_4165_4200nm.par
format = hitran
parfuncfmt = hapi
levelsfmt = radis
```
`entries["path"].strip("\n").split("\n")` -> `['d:\\github\\radis\\radis\\test\\files\\hitran_co2_626_bandhead_4165_4200nm.par']`

Earlier I was returning a string instead of a list and I do remember that most of the test cases failed because they were expecting a list instead of a string. So I added the `split() `function to return a list instead of string even when there is a single path. I'm surprised that even after removing the split() function all the test case passed!

erwanp:
Ok perfect, that's what i understood. Merging!

PR #258, opened 2021-04-22
fix: #244 return wavelength in sorted order

### LINKED ISSUE #244: Make Spectrum.get() return wavelength in sorted order
### 🐛 Describe the bug

When using the Spectrum.get() method with non default wavelength unit (such as "nm" or "nm_vac"), the arrays returned are decreasing instead of increasing.

### 🖥 Steps To Reproduce

```python
spec_h2o = calc_spectrum(wavelength_min=4170,
                                         wavelength_max=4190,
                                         Tgas=300,
                                         pressure=1.01325,
                                         mole_fraction=0.1,
                                         molecule='H2O',
                                         isotope='1,2,3,4,5',
                                         path_length=1,
                                         databank='hitran',
                                         )
spec_h2o.get('transmittance', wunit='nm')
```

### 🎯 Expected behavior

Return increasing arrays instead of decreasing

### 💡 Possible solutions

```
if array[0] > array[-1]:
    np.flip(array)
```

gagan-aryan:
### Description

The wavelengths and the intensity are now generated in sorted order by Spectrum.get()

Fixes #244


erwanp:
Hello @gagan-aryan  ! 

This one is probably harder that it seems : .get() is used internally a lot, and also in tests. We'll have to make sure it doesn't break anything (worst thing that could happen : we return the intensity in reversed order ! Fortunately, there are many tests everywhere). 

Good news is you'll get to learn a lot about the code structure : good practice for the GSOC project ! 

Tips : 
- [ ] search through all the repo wherever .get() is used, store this list somewhere.
- [ ] caution with the functions in the Line-of-sight (los) module : that's where we combine spectra and where things could go wrong.
- [ ] the Spectrum.plot() function also ensures spectra are properly sorted. This is not required if you already include it in .get().  However, the function there ensures the whole grid is properly sorted, not only first/last elements. This may seem unecessary, but may have been introduced for a reason too (I can't remember: maybe lookup the Git history ?).

gagan-aryan:
I was able to resolve the tests failing in `lbl` but there are still a few tests failing in `phys`, `tools`, `spectrum` and `validation`. There seems to be a problem in the calculation of the integral for `radiance_noslit`. The tests seem to be mainly failing when the following functions are used - 

- [get_residual_integral](https://github.com/radis/radis/blob/161e97cc9da802989e51b5be6138fa9e451ab425/radis/spectrum/compare.py#L380)
- [get_integral](https://github.com/radis/radis/blob/161e97cc9da802989e51b5be6138fa9e451ab425/radis/spectrum/spectrum.py#L1231)
- [get_residual](https://github.com/radis/radis/blob/161e97cc9da802989e51b5be6138fa9e451ab425/radis/spectrum/compare.py#L255) 

[2] Yes, there is a test fail that uses MergeSlabs from the LOS module. 

I am a bit stuck at this point though since I am unable to pinpoint the exact place where things are failing. I tried listing down all the tests that fail and the corresponding functions that are used. But that didn't help much with regard to the progress of this PR other than helping me understand a few parts of the codebase.

erwanp:
If current tests are not helpful to find the problem, we'll probably need to design new tests for this specific PR. 
To be clear : 

Radis spectra can come from various sources (codes, experimental data, etc.), and be stored in various wavegrids : wavenumbers (typically cm-1, the favorite unit of IR  spectroscopists), or wavelengths (typically nm, the favorite unit of UV/VIS spectroscopists). Problem : these two units scale inversely, i.e., nm ~ 1e7 / cm-1  ; so if you convert a spectrum stored in sorted-wavenumber (let's same 1000 - 2000 cm-1) in wavelength, it is now sorted in reverse order (10000 - 5000 nm). Radis currently handles this alright, and can plot / combine spectra with different waveunits. The only 'problem' (#244) may be for the end user : the spectra returned by `s.get()` are not always sorted. In [this community poll](https://github.com/radis/radis/issues/244#issuecomment-819764643), it was decided to always return spectra in sorted waverange. 

Trouble ahead :  :) 

What may go wrong is that because of sorting, the array is now reversed where it wasn't reversed before. It's not a problem if you just want to plot it. It's a major problem if it's combined with another spectrum and this happen's under the hood, without the end user knowing about it. Imagine this scenario : 

```python
w1, I1 = s1.get('abscoeff', wunit='nm')
w2, I2 = s2.get('abscoeff', wunit='nm')           
I_tot = I1 + I2
```
Above, if `s2.get()` is now reversed but `s1.get()` is not, we're inverting `I2` but not `I1`, therefore creating a major error in physics . This may happen if, for instance, `s1` was stored in wavelengths already sorted in ascending order (typical experimental spectra, or output of a UV/VIS code like [pySpecair](https://spectralfit.gitlab.io/specair/#)), whereas `s2` was stored in wavenumbers (typical output of a IR code like RADIS) : 

- In the current version (0.9.29), sorting is not ensured. Therefore, the wavenumber to wavelength conversion only changes `w2` (which is sorted in reverse order).  `I2` is not reversed. Therefore, `I_tot = I1 + I2` remains the same.
- With the changes from this PR #258, `I2` would be reversed during the wavenumber to wavelength conversion. Therefore, `I_tot = I1 + I2` would be wrong. 

What you need to do is : 

---

1. to carefully understand the example above, then, design a tests to make sure all possible tests respond correctly. You could use the `+` operator for this, use a test spectrum ; change how the wavegrid is sorted (in wavenumber / wavelength) to create another spectrum, and check that the operator still works properly. 

2. Then, find in the codes the different tests where similar problems can happen. I'll guide you for this : 
- during Comparison of spectra, in spectrum/compare.py. Fortunately, most comparison use the same [_get_defaults()](https://github.com/radis/radis/blob/develop/radis/spectrum/compare.py#L496) function. . This function is used to get a spectral array in the same units, 
- in the [los](https://github.com/radis/radis/blob/develop/radis/los/slabs.py) module where the functions MergeSlabs, SerialSlabs are used to combine spectra.
- in [spectrum/spectrum.py](https://github.com/radis/radis/blob/develop/radis/spectrum/spectrum.py#L3554) where some operators are defined : `s1 + s2`, `s1 // s2`, `s1 > s2`, etc.


3. to make sure the implemention is error-proof, for whatever function that uses `.get()` in RADIS ... or even in other codes that rely on RADIS (fortunately, there are only 2 so far I know of : fitroom / pyspecair ; but it's good practice to ensure that the code is reliable).  For instance, we could start by adding an option (`sort_waverange=Default('True')`)  in .get(), that would raise a warning whenever it's called without the user saying explicitely `sort_waverange=True` or `sort_waverange=False` ,  and keep this warning during the next 2 or 3 versions of RADIS.

erwanp:
Closing for the moment ; we'll re-open later.

PR #259, opened 2021-04-22
Added cutoff='None' option in calc_spectrum()

### LINKED ISSUE #251: No cutoff option in calc_spectrum()
### What is currently done
Currently the default "cutoff" value is 1e-27 cm-1/(#.cm-2), so when executing the code below:
```
spec_h2o = calc_spectrum(wavelength_min=4195,
                         wavelength_max=4215,
                         Tgas=300,
                         pressure=1.01325,
                         mole_fraction=0.001,
                         molecule='H2O',
                         isotope='1,2,3,4',
                         path_length=1,
                         databank='hitran',
                         )
```
An LinestrengthCutoffWarning is raised:
"LinestrengthCutoffWarning: Estimated error after discarding lines is large: 1.75%. Consider reducing cutoff"

The option to choose "None" for cutoff in the calc_spectrum() function, thus rejecting no line.

Just test if the cutoff is None and the put it equal to 0.
```
if cutoff is None:
    cutoff = 0
``` 

---

AlphaVictorDelta:
### Description

Added a new cutoff option in calc_spectrum(), which gives the user the option to give None as a parameter for the cutoff parameter. This is an alternate to the cutoff = 0. 

Fixes #251

### DIFF
--- a/radis/lbl/factory.py
+++ b/radis/lbl/factory.py
@@ -352,7 +352,7 @@ class SpectrumFactory(BandFactory):
         folding_thresh=1e-6,
         zero_padding=-1,
         broadening_method=Default("fft"),
-        cutoff=1e-27,
+        cutoff=0,
         verbose=True,
         warnings=True,
         save_memory=False,
@@ -476,7 +476,12 @@ class SpectrumFactory(BandFactory):
         # Initialize computation variables
         self.params.wstep = wstep
         self.params.pseudo_continuum_threshold = pseudo_continuum_threshold
+
+        if cutoff is None:
+            # If None, use no cutoff : https://github.com/radis/radis/pull/259
+            cutoff = 0
         self.params.cutoff = cutoff
+
         self.params.broadening_max_width = broadening_max_width  # line broadening
         self.misc.export_lines = export_lines
         self.misc.export_populations = export_populations


erwanp:
Hello ! I think these changes would be better added in the initialization of factory.py > SpectrumFactory, and the cutoff immediatly converted to 0.

erwanp (changes requested):
As mentioned previously, cutoff should be replaced by 0 earlier. 

What about [just before this line in SpectrumFactory initialization](https://github.com/radis/radis/blob/develop/radis/lbl/factory.py#L479) ?

AlphaVictorDelta:
So should I delete these changes in base.py?
by 'cutoff should be replaced by zero earlier' did you mean: self.params.cutoff = 0

erwanp:
Yes, revert these changes. Then just set cutoff=0 in the constructor, it will store it in self.params.cutoff automatically

radis/lbl/factory.py
```diff
@@ -476,7 +476,12 @@ def __init__(
         # Initialize computation variables
         self.params.wstep = wstep
         self.params.pseudo_continuum_threshold = pseudo_continuum_threshold
-        self.params.cutoff = cutoff
+
+        if cutoff is None:
+            self.params.cutoff = 0
+        else:
+            self.params.cutoff = cutoff
+
```
erwanp, line 484:
```suggestion
        if cutoff is None:
            # If None, use no cutoff : https://github.com/radis/radis/pull/259
            cutoff = 0
        self.params.cutoff = cutoff
```

Just a suggestion that shows better we're re-writing the default option (also good practice in case "cutoff" is ever used in the same function, not to have different values for "cutoff" and "self.params.cutoff")

erwanp (changes requested):

erwanp:
Ok, merging !

erwanp (approved):

PR #260, opened 2021-04-28
Update conditions with slit_shape

CorentinGrimaldi:
Fix untimely warning when rescaling caused by no shape saved in the conditions spectrum.

### Description

Fixes #<Issue Number>

### DIFF
--- a/radis/spectrum/rescale.py
+++ b/radis/spectrum/rescale.py
@@ -1714,7 +1714,10 @@ def _recalculate(
             try:
                 shape = spec.conditions["slit_shape"]
             except KeyError:
-                warn("Recomputing with slit_shape not given. Assuming a triangular slit.", UserWarning)
+                warn(
+                    "Recomputing with slit_shape not given. Assuming a triangular slit.",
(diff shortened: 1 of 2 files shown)


radis/spectrum/rescale.py
```diff
@@ -1712,7 +1712,7 @@ def _recalculate(
             slit_unit = spec.conditions["slit_unit"]
             norm_by = spec.conditions["norm_by"]
             try:
-                shape = spec.conditions["shape"]
+                shape = spec.conditions["slit_shape"]
             except KeyError:
                 shape = "triangular"
```
erwanp, line 1717: Can we also add a UserWarning warning here ? ("Recomputing with `slit_shape` not given. Assuming a triangular slit.")

erwanp (changes requested):

radis/spectrum/rescale.py
CorentinGrimaldi: Done

erwanp (approved):

Issue #262, opened 2021-04-28
pytest : No module named 'Cython' Found in setup.py 


anandxkumar:
### 🐛 Describe the bug

After #234, `pytest` fails on the local build.
While installing dependencies using `pip install -e .[dev] ` on a new virtual environment or existing, following error occurs -
```
Obtaining file:///home/pipebomb/radis
    ERROR: Command errored out with exit status 1:
     command: /home/pipebomb/anaconda3/envs/check_a/bin/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/home/pipebomb/radis/setup.py'"'"'; __file__='"'"'/home/pipebomb/radis/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-0frumtj7
         cwd: /home/pipebomb/radis/
    Complete output (5 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/home/pipebomb/radis/setup.py", line 39, in <module>
        from Cython.Distutils import build_ext
    ModuleNotFoundError: No module named 'Cython'
    ----------------------------------------
WARNING: Discarding file:///home/pipebomb/radis. Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.

```
While running `pytest` on develop branch clone, `test_arrays.py `fails with the following traceback -
<details>
<summary>Traceback</summary>
______________________________ test_cython_add_at ______________________________

args = (), kwargs = {}

    def test_cython_add_at(*args, **kwargs):
        """
        Compare the workings of the Cython compiled add_at() function
        versus the numpy add.at() function with bogus data.
        """
    
        # First check if Python was able to import the Cython version of add at:
>       from radis_cython_extensions import add_at as cython_add_at
E       ModuleNotFoundError: No module named 'radis_cython_extensions'

radis/test/misc/test_arrays.py:165: ModuleNotFoundError
__________________________ test_cython_add_at_spectra __________________________

args = (), kwargs = {}, wstep = 0.005, wmin = 2100, wmax = 2200, T = 1200
[...]

PR #266, opened 2021-05-02
Move GPU compilations to installer

dcmvdbekerom:
### Description

This pull request is to move the on-the-fly compilation of GPU codes to the installer.

It addresses the issue where the py_cuffs folder wasn't found, raised [on the Slack channel](https://app.slack.com/client/T9E5R3R9S/C9F7X0ZAT/thread/C9F7X0ZAT-1619738768.112700)

This PR is a "hotfix": the solution is still far from perfect.
Ideally all cupy and cupyx code should be moved out of the cython code into the regular python modules, and only the accelerated CPU routines should be compiled. These could then be moved to the radis_cython_extensions module we already had, and pure-python alternatives should be offered.


radis/lbl/factory.py
```diff
@@ -993,8 +959,8 @@ def eq_spectrum_gpu(
         log_2vMm = np.array(self._get_log_2vMm(molarmass_arr), dtype=np.float32)
         S0 = np.array(self._get_S0(Ia_arr), dtype=np.float32)
 
-        NwG = 4
-        NwL = 8
+        NwG = 4  # TO-DO: these shouldn't be hardcoded
```
erwanp, line 962: Good that you reported this !

erwanp (approved):
Great ! Do you think we can also address https://github.com/radis/radis/issues/262#issuecomment-830815789 at the same time ?

dcmvdbekerom:
>Great ! Do you think we can also address #262 (comment) at the same time ?

Yeah should be relatively easy, I'll have a look!

While I had a look at the GPU code, I'm also moving things around a bit to keep only the things that absolutely have to be in cython, in cython. I don't know how long this will take but usually the largest barrier is to get started, which is already overcome :)!

dcmvdbekerom:
Finished moving the GPU code around, now only parts that _have_ to be done in cython are in cython, and they are in the same file as the add_at() function. The rest is in the new lbl/gpu.py file which is much more inviting for future development.

Still haven't looked at #262 yet!

erwanp:
Well done!

pkj-m:
This is some really clean work!! Incredible and definitely LGTM!

pkj-m (approved):

PR #271, opened 2021-05-12
Added auto method for wstep

### LINKED ISSUE #250: Automatic wstep attribution in calc_spectrum()
### What is currently done 
Currently the default "wstep" value is 0.01, so when executing the code below:
```
spec_h2o = calc_spectrum(wavelength_min=4195,
                         wavelength_max=4215,
                         Tgas=300,
                         pressure=1.01325,
                         mole_fraction=0.001,
                         molecule='H2O',
                         isotope='1,2,3,4',
                         path_length=1,
                         databank='hitran',
                         )
```
An AccuracyWarning is raised: 
"Some lines are too narrow (FWHM ~ 0.027 cm⁻¹) for the current spectral grid (wstep=0.01). Please reduce wstep to below 0.009 cm⁻¹. You can also ignore by setting `warnings={'AccuracyWarning':'ignore'}` (if you know what you're doing!)"

The option to choose "auto" for wstep in the calc_spectrum() function. (Maybe update the default behavior to it)

Since the warning already give the value that is pertinent for the current case, the wstep="auto" option should just take it (0.009 in the example).

---

anandxkumar:
### Description

This pull request is to add an auto method in wstep. This method modifies `wstep = min_width/WARN_THRESHOLD` if lines are narrow otherwise `wstep=0.01`

@erwanp can you help me understand the error calculation process using `wstep` ?

Also a test needs to be added to test proper functioning of `auto` method.

Fixes #250


anandxkumar:
`Pytest` and `xvfb-run pytest --cov=./` passing locally but failed on the travis.

erwanp (commented):
If I understand well the current implementation, wstep will be automatically reduced if needed,  but not increased. We  only also increase it to make computations faster. 

Also : changing wstep is not enough. You should also make sure the whole wavenumber_calc array is recomputed. I think this is currently done before the linewidths are computed. If so, it should  be reversed.

anandxkumar:
@erwanp for the first point of yours, in my implementation we are reducing the wstep if accuracy is low but for the rest of the cases we are sticking with` wstep=0.01` (the default value). If I'm not wrong the fastest and accurate performance that we will get is setting wstep = min_width / WARN_THRESHOLD because this will represent the highest value of wstep that will give sufficient accuracy but if we increase the wstep further though it will make computations faster it will result the same warning message thus bringing us where we started so I'm a little confused by your first point. Also wstep = 0.01 won't be the best value for every scenario performance wise even with sufficient accuracy, so may be we can stick to wstep = min_width / WARN_THRESHOLD even if there is sufficient accuracy? 

For the second point of yours, the `wavenumber_calc` is calculated when we initialize the spectrum_factory object [link](https://github.com/radis/radis/blob/03ee3e2b0ea47ce9235241c7a3c66b7f852c30a4/radis/lbl/factory.py#L415).  Now if the wstep changes we can recompute `wavenumber_calc` by calling the `_generate_wavenumber_range()` function but the issue is that there are a number of variables that are dependent on `wstep` that are -  `wavnumber_calc`, `wbroad_centered` and `v_arr`. 
If we see individual parameters and their dependency on the entire codebase which can cause problem -
- wavenumber_calc - `self._wavenum_min_calc` and `self._wavenum_max_calc` are used in `loader.py` [here](https://github.com/radis/radis/blob/03ee3e2b0ea47ce9235241c7a3c66b7f852c30a4/radis/lbl/loader.py#L1722)
- wbroad_centered - In `base.py` [here](https://github.com/radis/radis/blob/03ee3e2b0ea47ce9235241c7a3c66b7f852c30a4/radis/lbl/base.py#L2836) 
- v_arr - Used in `gpu_init` in both `factory.py` and in `gpu.py` [here](https://github.com/radis/radis/blob/03ee3e2b0ea47ce9235241c7a3c66b7f852c30a4/radis/lbl/gpu.py#L306)

All these cases can lead to accuracy error if they are being used before we calculate HWHM to check the accuracy which I'm not sure about for the above cases (the rest cases are fine because they being used for calculation of lineshape which comes after we calculate HWHM). If all these steps are executed after the calculation of HWHM then we can shift all the parameters dependent on wstep to the part we recalculate wstep (in [_check_accuracy](https://github.com/radis/radis/blob/03ee3e2b0ea47ce9235241c7a3c66b7f852c30a4/radis/lbl/broadening.py#L845) itself)

erwanp:
1. agree, `wstep = min_width / WARN_THRESHOLD` is the optimum. One subtle point, though : when combining spectra, for instance : 

```
(s_CO2 // s_CO) > s_room_absorption
```  
then all grid points should be the same, else RADIS will have to interpolate, which is done automatically, but wastes some time and looses some accuracy. Currently, assuming you compute spectra with the same wavegrid range and resolution, no interolation is triggered. If we start using `wstep = min_width / WARN_THRESHOLD`, every spectrum will have a unique wavegrid resolution, which will trigger a lot of interpolation, which is unwanted. How to prevent this ? 

One way I can think of is to collapse the ideal wstep resolution on a 1 or 2 digits number ; i.e, `"0.014276765"` would become `"0.01"` or `"0.014"`  [slightly more refined than the ideal, not to trigger the warning]. More digits : each spectrum would individually be calculated in a more optimized way. Less digits : more chances that when combining spectra no interpolation is required. 


2. `wavenumber_calc` array should only be initialized within `eq_spectrum` and `non_eq_spectrum`, etc ; and not at the SpectrumFactory initializing.   The best way to understand the different calculation steps is to look within [SpectrumFactory.eq_spectrum](https://github.com/radis/radis/blob/03ee3e2b0ea47ce9235241c7a3c66b7f852c30a4/radis/lbl/factory.py#L659)   (which slightly follows the [RADIS flow chart](https://radis.readthedocs.io/en/latest/_images/RADIS_flow_chart.svg) ).  
- The broadening widths are computed in [_calc_broadening_HWHM](https://github.com/radis/radis/blob/03ee3e2b0ea47ce9235241c7a3c66b7f852c30a4/radis/lbl/factory.py#L686). 
- The wavegrid is used in the next step, [_calc_broadening](https://github.com/radis/radis/blob/03ee3e2b0ea47ce9235241c7a3c66b7f852c30a4/radis/lbl/factory.py#L692). 
Therefore, the `wavenumber_calc` array should be initialized in between the two. You'll still need the different boundaries ("wavenum_min_calc, wavenum_max_calc") as soon as the SpectrumFactory / LoaderFactory object is initialized, to download the database, but no need to build the calculation array at this point) .

anandxkumar:
There are some issues related to the `wavenumber_calc` array.
- `wout_range` is dependent on wstep. So the array value will change after `_calc_broadening_HWHM`. Also if I remove this part in spectrum initialization then it raises error during unit conversion (`phys/units.py`) related to shape mismatch so here we will need to declare `wavenumber_calc`.
- `wbroad_centered` matrix is needed  for calculating the [_cutoff_linestrength](https://github.com/radis/radis/blob/03ee3e2b0ea47ce9235241c7a3c66b7f852c30a4/radis/lbl/base.py#L2833) which comes before the `_calc_broadening_HWHM` step. So here also we will face accuracy error.
- `self.params.wavenum_min` and `self.params.wavenum_min` are required for fetching the data (eg - `fetch_astroquery`). This will of course needed to be executed before the `_calc_broadening_HWHM` step. So here also we will face accuracy error.

All these issues will need a lot of refactoring and I'm a little doubtful that whether we can refactor the code that these things will be executed after we calculate the HWHM to set wstep to its desired value. 
After this thing is resolved we can set the precision of wstep to a degree of 2 to have accuracy as well as reduce the number of  interpolations.

erwanp:
Good findings! 

1. Can you point me to the failing line and the units problem ? 
2. The use of the special grid in cutoff linestength_cutoff is just for informative purpose, to compute an estimate of the absolute time gained. It's an old block of code, not valid anyway in the LDM method (where time does not scale linearly with number of lines as you know) . You can comment the whole paragraph. We may reimplement it "a posteriori".  https://github.com/radis/radis/blob/03ee3e2b0ea47ce9235241c7a3c66b7f852c30a4/radis/lbl/base.py#L2833
3. Fetch only requires the bounds, not the array itself.

anandxkumar:
1. I removed the `woutrange` array from  SF init and declared it straight after `_calc_broadening_HWHM()`. On using `pytest`  only 1 test failed and error is [here](https://github.com/radis/radis/blob/03ee3e2b0ea47ce9235241c7a3c66b7f852c30a4/radis/phys/units.py#L291)- 
<details>
<summary>    Traceback</summary>
_____________________________________________ test_all_calc_methods_CO2pcN ______________________________________________

verbose = True, plot = False, warnings = True, rtol = 0.001, args = (), kwargs = {}
getDatabankEntries = <function getDatabankEntries at 0x7f94f3209598>
define_Evib_as_min_of_polyad = <function define_Evib_as_min_of_polyad at 0x7f94ecea1c80>
discard_lines_with_na_levels = <function discard_lines_with_na_levels at 0x7f94ecea1bf8>
setup_test_line_databases = <function setup_test_line_databases at 0x7f94eceb4d90>, Tgas = 500, iso = 1
sf = <radis.lbl.factory.SpectrumFactory object at 0x7f95323b9a90>
database_kwargs = {'format': 'cdsd-hitemp', 'levels': {1.0: '/home/pipebomb/Desktop/radis/radis/test/files/co2_cdsd_hamiltonian_fragment.levels'}, 'levelsfmt': 'cdsd-pcN', 'parfunc': '/home/pipebomb/Desktop/radis/radis/db/CO2/partition_functions.txt', ...}
Q_calc = <radis.levels.partfunc_cdsd.PartFuncCO2_CDSDcalc object at 0x7f94dc33c4e0>

    def test_all_calc_methods_CO2pcN(
        verbose=True, plot=False, warnings=True, rtol=1e-3, *args, **kwargs
    ):
        """Test same spectrum for 3 different calculation variants (equilibrium,
        non-equilibrium, per band and recombine
    
        Uses CO2 Levels database where the energy partitioning is done as follow:
    
            2 nonequilibrium modes
            Evib is the minimum of a "p,c,N" group
            Erot = E - Evib
    
        This corresponds to the levelsfmt = 'cdsd-pcN' in
        :data:`~radis.lbl.loader.KNOWN_LVLFORMAT`
        """
    
        from radis.misc.config import getDatabankEntries
        from radis.test.utils import (
            define_Evib_as_min_of_polyad,
            discard_lines_with_na_levels,
            setup_test_line_databases,
        )
    
        if plot:  # Make sure matplotlib is interactive so that test are not stuck in pytest
            import matplotlib.pyplot as plt
    
            plt.ion()
        # %%
        Tgas = 500
    
        iso = 1
        sf = SpectrumFactory(
            wavenum_min=2284,
            wavenum_max=2285,
            broadening_max_width=5,  # TODO @EP: crashes with 0.3?
            mole_fraction=1,
            path_length=0.025,
            cutoff=1e-25,
            molecule="CO2",
            isotope=iso,
            verbose=verbose,
            export_lines=True,
        )
        sf.warnings["MissingSelfBroadeningWarning"] = "ignore"
        sf.warnings["NegativeEnergiesWarning"] = "ignore"
        sf.warnings["HighTemperatureWarning"] = "ignore"
    
        # Preparation:
    
        setup_test_line_databases()
    
        # Generate a Levels database with p,c,N energy partitioning
        # ... copy "HITEMP-CO2-HAMIL-TEST" info
        database_kwargs = getDatabankEntries("HITEMP-CO2-HAMIL-TEST")
        # ... adapt it to 'cdsd-pcN' mode
        del database_kwargs["info"]
        database_kwargs["levelsfmt"] = "cdsd-pcN"
        # ... load the new database
        sf.load_databank(**database_kwargs, load_energies=True)
    
        # Now, define Evib:
        Q_calc.df = define_Evib_as_min_of_polyad(Q_calc.df, keys=["p", "c", "N"])
    
        # With this Evib definition, clean the Lines database from where Evib is not defined
        # (because Levels does not exist in the reduced, test Level Database)
        discard_lines_with_na_levels(sf)
    
        # %%---------------------
        # Ready, let's start the tests:
    
>       s_bands = sf.non_eq_bands(Tvib=Tgas, Trot=Tgas)

radis/test/lbl/test_calc.py:495: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
radis/lbl/bands.py:658: in non_eq_bands
    radiance_noslit, wavenumber, "mW/sr/cm2/cm-1", "mW/sr/cm2/nm"
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

l_cm = array([[1.15066180e-06, 1.15645589e-06, 1.16230986e-06, 1.16822417e-06,
        1.17419992e-06, 1.18023757e-06, 1.1863...e-06,
        1.28962323e-06, 1.28243390e-06, 1.27532199e-06, 1.26828611e-06,
        1.26132559e-06, 1.25443905e-06]])
wavenum = array([2284.  , 2284.01, 2284.02, 2284.03, 2284.04, 2284.05, 2284.06,
       2284.07, 2284.08, 2284.09, 2284.1 , 2284....4.9 ,
       2284.91, 2284.92, 2284.93, 2284.94, 2284.95, 2284.96, 2284.97,
       2284.98, 2284.99, 2285.  , 2285.01])
Iunit0 = 'mW/sr/cm2/cm-1', Iunit = 'mW/sr/cm2/nm'

    def convert_rad2nm(l_cm, wavenum, Iunit0, Iunit):
        """Convert spectral radiance in wavenumber base (~1/cm-1) to spectral
        radiance in wavelength base (~1/nm)
    
        Parameters
        ----------
        l_cm: array
            spectral radiance in yunit0 unit (~ per wavenumber)
        wavenum: array (cm-1)
            wavenumber
        Iunit0: str
            unit (~ per wavenumber) to convert from
        Iunit: str
            unit (~ per wavelength) to convert to
    
        Notes
        -----
        Implementation:
    
        We use a variable substitution::
    
                dλ/dν = - 1/ν**2        [in SI]
                dλ/dν = - 10^7 / ν**2   [nm -> cm-1]
    
        We want   Lν.dν = Lλ.dλ
        so        Lλ  = - Lν * 1e-7 * ν^2
        """
    
        l_cm = conv2(l_cm, Iunit0, "mW/cm2/sr/cm-1")
    
        # Convert 'mW/cm2/sr/cm-1' to 'mW/cm2/sr/nm'
>       l_nm = l_cm * 1e-7 * wavenum ** 2
E       ValueError: operands could not be broadcast together with shapes (1,602) (102,)

radis/phys/units.py:291: ValueError

File /home/pipebomb/Desktop/radis/radis/test/files/co2_cdsd_hamiltonian_fragment.levels.h5 deprecated:
Metadata in file /home/pipebomb/Desktop/radis/radis/test/files/co2_cdsd_hamiltonian_fragment.levels.h5 dont match expected values. See comparison below:
        Expected        File
Key            Expected Got
levelsfmt      cdsd-pcN cdsd-hamil
viblvl_label   pcN      pcJN

Deleting it!
Calculating Non-Equilibrium Spectrum
Physical Conditions
   Tgas                 300 K
   Trot                 300 K
   Tvib                 300 K
   isotope              1
   mole_fraction        1
   molecule             CO2
   path_length          0.025 cm
   self_absorption      True
   state                X
   wavenum_max          2285.0000 cm-1
   wavenum_min          2284.0000 cm-1
Computation Parameters
   Tref                 296 K
   broadening_max_width  5 cm-1
   broadening_method    fft
   cutoff               1e-25 cm-1/(#.cm-2)
   dbformat             cdsd-hitemp
   dbpath               /home/pipebomb/Desktop/radis/radis/test/files/cdsd_hitemp_09_fragment.txt
   dlm_log_pG           0.1375350788016573
   dlm_log_pL           0.20180288881201608
   folding_thresh       1e-06
   levelsfmt            cdsd-pcN
   optimization         simple
   parfuncfmt           cdsd
   parfuncpath          /home/pipebomb/Desktop/radis/radis/db/CO2/partition_functions.txt
   pseudo_continuum_threshold  0
   wavenum_max_calc     2287.5100 cm-1
   wavenum_min_calc     2281.5000 cm-1
   waveunit             cm-1
   zero_padding         -1
Information
   db_use_cached        True
   export_lines         True
   load_energies        True
   lvl_use_cached       True
   total_lines          2284
Fetching Evib & Erot.
Calculating Non-Equilibrium bands
Physical Conditions
   Tgas                 500 K
   Trot                 500 K
   Tvib                 500 K
   isotope              1
   mole_fraction        1
   molecule             CO2
   path_length          0.025 cm
   self_absorption      True
   state                X
   wavenum_max          2285.0000 cm-1
   wavenum_min          2284.0000 cm-1
Computation Parameters
   Tref                 296 K
   broadening_max_width  5 cm-1
   broadening_method    fft
   cutoff               1e-25 cm-1/(#.cm-2)
   dbformat             cdsd-hitemp
   dbpath               /home/pipebomb/Desktop/radis/radis/test/files/cdsd_hitemp_09_fragment.txt
   dlm_log_pG           0.1375350788016573
   dlm_log_pL           0.20180288881201608
   folding_thresh       1e-06
   levelsfmt            cdsd-pcN
   optimization         simple
   parfuncfmt           cdsd
   parfuncpath          /home/pipebomb/Desktop/radis/radis/db/CO2/partition_functions.txt
   pseudo_continuum_threshold  0
   wavenum_max_calc     2287.5100 cm-1
   wavenum_min_calc     2281.5000 cm-1
   waveunit             cm-1
   zero_padding         -1
Information
   db_use_cached        True
   export_lines         True
   load_energies        True
   lvl_use_cached       True
   total_lines          2284
... sorting lines by vibrational bands
... lines sorted in 0.0s
... sorting lines by vibrational bands
... lines sorted in 0.0s
wstep =  0.01
Now processing lines (5)
(0s)    100.0%
Generating bands (5)
</details>

2. We can remove the `wbroad_centered` matrix from the SpectrumFactory Initialization part and straight away compute it after `_calc_broadening_HWHM()`. I removed the `expected_broadening_time_gain` part and all tests cases ran fine. So a thumbs up here!

3. Even though we need the boundaries there will some difference, for example, I was using the following code -
```
calc_spectrum(wavelength_min=4195,
                         wavelength_max=4215,
                         Tgas=300,
                         pressure=1.01325,
                         mole_fraction=0.001,
                         molecule='H2O',
                         isotope='1,2,3,4',
                         path_length=1,
                         databank='hitran',
                         wstep = 'auto',
                         verbose = 0
                         )
```
First when using `wstep=0.01`, ` wavenum_min_calc[0] = 2366.83242701` and `wavenum_max_calc[-1] = 2388.14242701`. After updating the threshold value automatically to `wstep=0.00898126503455555` , ` wavenum_min_calc[0] = 2366.82986238` and `wavenum_max_calc[-1] = 2388.15138557`. Though it is a small difference (~0.01 and also depends on `min_width` and `wstep`) I'm not sure how much it will affect the accuracy. Also `wavenumber = arange(wavenum_min, wavenum_max + wstep, wstep)` so there can be difference between the two `wavenumber[-1]` due to difference in the `default` and `min_width/ Warn Threshold` value of`wstep`. If this much difference is not significant then we can set: 
`self.params.wavenum_min_calc` = wavenumber[0] - broadening_max_width / 2
`self.params.wavenum_max_calc` = wavenumber[-1] + broadening_max_width / 2  

I tried it with above code, and every test case passed.

radis/lbl/base.py
```diff
@@ -2831,9 +2831,7 @@ def _cutoff_linestrength(self, cutoff=None):
 
         # Estimate time gained
         expected_broadening_time_gain = (
-            self._broadening_time_ruleofthumb
-            * Nlines_cutoff
-            * len(self.wbroad_centered)
+            self._broadening_time_ruleofthumb * Nlines_cutoff
```
erwanp, line 2834: as you suggest in your PR, you can remove this entirely. To add again once we have better time estimate of how long a calculation takes based on the number of lines and the method used (that's your GSOC project!). Keep a TODO here maybe?

radis/lbl/broadening.py
```diff
@@ -887,27 +887,37 @@ def _check_accuracy(self, df, wstep):
...
-                f"Some lines are too narrow (FWHM ~ {min_width:.2g} cm⁻¹) for "
-                + f"the current spectral grid (wstep={wstep}). Please reduce "
-                + f"wstep to (at least) below {min_width/ERROR_TRESHOLD:.2g} cm⁻¹ "
-                + f"or (suggested) {min_width/WARN_THRESHOLD:.2g} cm⁻¹. "
-                + "You can also ignore by setting `warnings={'AccuracyError':'ignore'}` "
-                + "(if you know what you're doing!)",
-                "AccuracyError",
-            )
+            if wstep_auto:
+                self.params.wstep = min_width / WARN_THRESHOLD
```
erwanp, line 893: It's confusing that this very important step is done in the function `_check_accuracy`. 
Can we move it further up ? For instance in the method that builds the wavegrid array.  Keep only the warnings in the `_check_accuracy` method. 

You'll need "WARN_THRESHOLD" as some kind of global parameter. Temporarily you can use [radis/params.py](https://github.com/radis/radis/blob/develop/radis/params.py) for that. We'll move them in a .json/.yaml later on.

radis/lbl/factory.py
```diff
@@ -410,23 +410,30 @@ def __init__(
             wavelength_max,
             medium,
         )
+        # Checking if 'auto' mode used for wstep
+        self.wstep_auto = False
+        if wstep == "auto":
+            self.wstep_auto = True
+            wstep = 0.01
```
erwanp, line 417: this "0.01" shouldn't be needed anymore if using auto-mode.

radis/lbl/factory.py
```diff
@@ -410,23 +410,30 @@ def __init__(
...
         woutrange = np.in1d(wavenumber_calc, wavenumber, assume_unique=True)
-        self.wbroad_centered = wbroad_centered
+        # self.wbroad_centered = wbroad_centered
         self.wavenumber = wavenumber
         self.wavenumber_calc = wavenumber_calc
         self.woutrange = woutrange
-
+        print("e ->wavenum - ", self.wavenumber)
+        print("e ->wavenum_calc - ", self.wavenumber_calc)
+        print("e ->wbroad - ", self.wbroad_centered)
```
erwanp, line 436: /!\ still some prints there.

radis/lbl/factory.py
```diff
@@ -685,6 +692,33 @@ def eq_spectrum(
...
+        self.wavenumber_calc = wavenumber_calc
+        self.woutrange = woutrange
+        self.params.wavenum_min_calc = wavenumber_calc[0]
+        self.params.wavenum_max_calc = wavenumber_calc[-1]
+        print("l ->wavenum - ", self.wavenumber)
+        print("l ->wavenum_calc - ", self.wavenumber_calc)
+        print("l ->wbroad - ", self.wbroad_centered)
+
+        #####################################################
+
```
erwanp, line 721: can you move this in a function ? It's probably the right place to compute wstep_auto too.

radis/lbl/factory.py
```diff
@@ -1273,6 +1307,33 @@ def non_eq_spectrum(
...
```
erwanp, line 1336: same as above

erwanp (changes requested):

anandxkumar:
@erwanp I made the changes. Just a small issue, earlier we were using `_wstep` for saving the `wstep` initial value. Then we were using `assert self._wstep == self.params.wstep` to check the value in [_check_inputs()](https://github.com/radis/radis/blob/03ee3e2b0ea47ce9235241c7a3c66b7f852c30a4/radis/lbl/base.py#L3047). This step comes before `_calc_broadening_HWHM()` so this will fail for `wstep = 'auto'`. I commented out this part. Any alternative to this.

Plus I believe we will need to add tests for wstep = 'auto' for both `eq_spectrum` and `non_eq_spectrum`. And if yes then where should these tests be located?

Also: I'm not sure where we need to put the `_generate_wavenumber_entities()` function in `eq_spectrum_gpu` and also it should be done before the declaration of [v_arr](https://github.com/radis/radis/blob/03ee3e2b0ea47ce9235241c7a3c66b7f852c30a4/radis/lbl/factory.py#L938) which is dependent on wstep.

erwanp:
SpectrumFactory can compute several spectra in a batch, that may have different wstep in auto mode. 

So we need two variables, one to hold the initial input and one to hold the value evaluated for each spectrum. They will diverge at some point (one remains "auto"  the other one is evaluated) so "assert equal"  in check_inputs won't be needed anymore. However we need to make sure one of the value remains "auto"  else the 2nd spectrum of the Factory won't be computed in auto mode. I think _wstep is too much of a hidden variable for this. We could replace it with  sf.misc.wstep (or simply sf.wstep), for the input (auto or value)  and sf.params.wstep for the evaluated value, used at calculation time.

anandxkumar:
`self.wstep` will represent the initial value of `wstep` where as `self.params.wstep` will be the computed value. I made the changes and for the below example I was getting the following result:

```
sf = SpectrumFactory(
    wavelength_min=4400,
    wavelength_max=4800,
    mole_fraction=0.01,
    cutoff=1e-25,
    wstep="auto",
    isotope=[1],
    db_use_cached=True,
    self_absorption=True,
    verbose=False,
)
sf.load_databank("HITRAN-CO-TEST")
```
`s1 = sf.non_eq_spectrum(1500, 1500, path_length=0.01)`
self.min_width =  0.028311323556812134
self.wstep- = auto
self.params.wstep = 0.009

`s2 = sf.eq_spectrum(300)`
self.min_width =  0.08322811834706903
self.wstep- = auto
self.params.wstep = 0.028

`s3 = sf.non_eq_spectrum(500, 600, path_length=0.05)`
self.min_width =  0.05230942894927567
self.wstep- = auto
self.params.wstep = 0.017

----------------------------------------------------------------
For wstep = 0.005, following was the result:
`s1 = sf.non_eq_spectrum(1500, 1500, path_length=0.01)`
self.min_width =  0.028311323556812134
self.wstep- = 0.005
self.params.wstep = 0.005

`s2 = sf.eq_spectrum(300)`
self.min_width =  0.08322811834706903
self.wstep- = 0.005
self.params.wstep = 0.005

`s3 = sf.non_eq_spectrum(500, 600, path_length=0.05)`
self.min_width =  0.05230942894927567
self.wstep- = 0.005
self.params.wstep = 0.005

radis/lbl/broadening.py
```diff
@@ -889,7 +889,7 @@ def _check_accuracy(self, df, wstep):
 
         ERROR_TRESHOLD = 1
 
-        if self.params.wstep == "auto":
+        if self.wstep == "auto":
```
erwanp, line 892: I think their should be no special treatment for 'auto' method. By construction it should pass the tests, but that's not a reason not to test it (what if someone somehow ends up manually resetting the wstep value after it was set by 'auto' ? )

radis/lbl/factory.py
```diff
@@ -932,8 +931,8 @@ def eq_spectrum_gpu(
         # generate the v_arr
         v_arr = np.arange(
             self.input.wavenum_min,
-            self.input.wavenum_max + self._wstep,
-            self._wstep,
+            self.input.wavenum_max + self.wstep,
+            self.wstep,
```
erwanp, line 935: shouldn't it be `self.params.wstep` here (the numerical value) ?

erwanp (changes requested):
👍

radis/__init__.py
```diff
@@ -98,7 +98,13 @@
 
 
 # %% Global constants
-from .params import AUTO_UPDATE_SPEC, DEBUG_MODE, OLDEST_COMPATIBLE_VERSION, USE_CYTHON
+from .params import (
+    AUTO_UPDATE_SPEC,
+    DEBUG_MODE,
+    OLDEST_COMPATIBLE_VERSION,
+    USE_CYTHON,
+    WARN_THRESHOLD,
```
erwanp, line 106: to keep things together you could also keep the ERROR_THRESHOLD here. And now that they're mixed with other parameters let's call them more explicitely. What do you think of :
```
GRIDPOINTS_PER_LINEWIDTH_WARN_THRESHOLD
GRIDPOINTS_PER_LINEWIDTH_ERROR_THRESHOLD
```
(long, but explicit)

radis/lbl/bands.py
```diff
@@ -551,6 +551,9 @@ def non_eq_bands(
         # ... calculate broadening  HWHM
         self._calc_broadening_HWHM()
 
+        # ... generates all wstep related entities
+        self._generate_wavenumber_entities()
```
erwanp, line 555: what do you think of `self._generate_wavenumber_arrays()`  as a name ?

radis/lbl/broadening.py
anandxkumar: Actually, this line is required to prevent `self.wstep = 'auto'`  to have an arithmetic comparison in the next line because till this point we won't have calculated value of `wstep`, so this line is necessary. So tests cases will pass but the `auto` method won't work.

radis/lbl/base.py
```diff
@@ -3044,7 +3026,10 @@ def _check_inputs(self, mole_fraction, Tmax):
         # ... that wstep and broadening_max_width were not inadvertanly changed
         # ... (would have no effect as the waverange is calculated on SpectrumFactory
         # ... initialization)
-        assert self._wstep == self.params.wstep
+
+        # Checks there if there is change in wstep value if initial wstep != "auto"
+        if self.wstep != "auto":
```
anandxkumar, line 3031: @erwanp I believe you wanted to say those [lines](https://github.com/radis/radis/pull/271#discussion_r644231439) for this part? If I remove this line then if we compute 2 Spectrum from a single SF object the `self.params.wstep` value will never be equal to `auto` and an Assertion Error will always be thrown `assert self.wstep == self.params.wstep`. If someone manually changes the wstep value then it is fine but if we generate a new spectrum from same SF object it will be wrong. Any possible solution to this? One way I can think is resetting the `self.params.wstep` value after calculating spectrum.

radis/lbl/broadening.py
erwanp: This test is run after all FWHM have been computed already, so it should be possible to move the 'auto'  evaluation right before the test, either move the accuracy test right after 'auto'  was evaluated.

anandxkumar: Earlier we were doing the `auto` evaluation in `_check_accuracy` itself. So I implemented the later suggestion. I broke the `_check_accuracy` step in 2 parts - `_calc_min_width` to get the minimum width and `_check_accuracy` to check accuracy of wstep. Both these functions are now being called in `_generate_wavenumber_arrays` function so all computation and checks will be done in a single function.

erwanp: Yes, it's a much better architecture with the two functions ; well done !

erwanp:
It was a large change but we're almost there! 

Something I see missing : 

- in [BandFactory.eq_bands()](https://github.com/radis/radis/blob/251503a155b87d49eb5bd292f0b2ad9676fc553b/radis/lbl/bands.py#L233)  there should also be the wavenumber_array generation step (it's only in non_eq_bands ; probably becasue eq_bands is not used in tests ?)

Also, what tests do you think we should implement ? I think of : 

- making sure .step ; .params.step are properly written, like in https://github.com/radis/radis/pull/271#issuecomment-853283299
- making sure using 'auto' and computing several spectra with the same Factory use a new wstep for each spectrum (varying the Pressure rather than the temperature is the best thing to change the linewidth. You could compute at 1 atm, 20 mbar and 1 mbar ; and expect very different linewidths) 
- maybe add a simple, unitary test for your `round_off` function ? (may be easier to move it into [radis/misc/basics.py](https://github.com/radis/radis/blob/develop/radis/misc/basics.py) for instance )

The [coverage of this PR](https://app.codecov.io/gh/radis/radis/compare/271) may help. 

Tests in RADIS are written in the [radis/test](https://github.com/radis/radis/tree/develop/radis/test) folder, which tries to replicate the architecture of the main code. So your `round_off` test could be added in test/misc/test_basics.py ; and high level tests like the 'auto' mode could be added in test/lbl/broadening.py or test/lbl/factory.py   . They will be run and found by `pytest` as long as their name starts with "test*". You can also mark some (the unitary tests, for instance) as 'fast' with @pytest.mark.fast,  which allows to run only the very fast tests by running `pytest -m fast` (more on the [Developer Guide](https://radis.readthedocs.io/en/latest/dev/developer.html#test))

anandxkumar:
Yes, we need to add the `_generate_wavenumber_array()` code in `eq_bands()`. There is a test for [eq_bands](https://github.com/radis/radis/blob/251503a155b87d49eb5bd292f0b2ad9676fc553b/radis/test/lbl/test_bands.py#L90). The test cases passed earlier because before executing `sf.eq_bands` we were executing `sf.non_eq_bands` which declared the global parameters thus error was not generated. After adding `_generate_wavenumber_array()` code in `eq_bands()` it works without any error.

I added 3 different tests. One simple test for `round_off` function in `test_basic.py`. 
One test `test_wstep_auto_method_sf` is for checking that when we compute various spectrum from a single SF object by varying pressure, and checking `sf.wstep == auto` and `params.wstep` for all spectrum are different when using `auto` method.

Last test `test_all_spectrum_using_wstep_auto`  is for executing all the possible wstep dependent spectrum functions that are `eq_spectrum`, `non_eq_spectrum`,  `eq_bands`, `non_eq_bands` and checking at the end `sf.wstep == "auto".

Let me know any possible improvement!

erwanp:
I think it's ready for publication! 

It's not a small change and could lead to unexpected behaviors, so we'll wait at least one version before making it the default mode. But it should make it to 0.9.30 and we should encourage the core users to try it!

Is it documented in calc_spectrum and SpectrumFactory? We could add a small "warning / new : wstep='auto' is optimized for performances while ensuring accuracy, but is still experimental in 0.9.30. Feedback welcome! "  there.

anandxkumar:
Updated the docs. Anything else?

docs/lbl/lbl.rst
```diff
@@ -132,6 +135,8 @@ Be careful to be consistent and not to give partial or contradictory inputs. ::
             verbose=verbose,
         )
 
+**NOTE:** Warning / New:  wstep = 'auto' is optimized for performances while ensuring accuracy,
```
erwanp, line 138: Hello ! Good idea to mention if directly on the doc, but maybe directly in 'calculate-multiple-molecules-spectrum' is not the right place [(see the page here)](https://radis.readthedocs.io/en/latest/lbl/lbl.html#calculate-multiple-molecules-spectrum) 

What about a dedicated section after [Advanced/Computation Parameters](https://radis.readthedocs.io/en/latest/lbl/lbl.html#computation-parameters), which could be "Choose the right wavenumber grid" explaining there is a tradeoff in between accuracy and performance. 

Have a look at the small edits I made in your function docstrings : they show how to add a 'note' , and also how to refer to the warning parameters, for instance. 

Side note : if you edit lbl.rst, on the same files some links are broken, for instance 
```
:py:func:`~radis.lbl.calc.eq_spectrum_gpu`
``` 
which should be 
```
:py:meth:`~radis.lbl.factory.SpectrumFactory.eq_spectrum_gpu`
```

anandxkumar: Added the new section `Choose the right wavenumber grid` as you mentioned. You may want to look into its content and modify it accordingly.

docs/lbl/lbl.rst
```diff
@@ -696,6 +691,27 @@ for an example, which can be run with (you will need the CDSD-HITEMP database in
...
+Based on ``wstep``, it will determine the number of gridpoints per linewidth.
+To make sure that there are enough gridpoints, Radis will raise an Accuracy Warning :py:meth:`~radis.lbl.broadening.BroadenFactory._check_accuracy`
+if number of gridpoints are less than :py:data:`~radis.params.GRIDPOINTS_PER_LINEWIDTH_WARN_THRESHOLD` and raises an Accuracy Error
+if number of gridpoints are less than :py:data:`~radis.params.GRIDPOINTS_PER_LINEWIDTH_ERROR_THRESHOLD`.
+
+From ``0.9.30`` a new mode ``wstep='auto'`` has been added which directly computes the optimum value of ``wstep``
+ensuring both performance and accuracy. It is ensured that there are slightly more or less than :py:data:`~radis.params.GRIDPOINTS_PER_LINEWIDTH_WARN_THRESHOLD`
+points for each linewidth.
+
+.. Note::
```
erwanp, line 711: I think capitalized won't work. You can directly accept the suggestion below :

```suggestion
.. note::
```

erwanp (approved):

docs/lbl/lbl.rst
erwanp: It's good and clear. I just added a commit suggestion

docs/lbl/lbl.rst
anandxkumar: `Note` also works fine.
Screenshot from 2021-06-04 17-20-03

radis/params.py
```diff
@@ -65,7 +65,7 @@
 
 See Also
 --------
-:py:func:`~radis.lbl.broadening._check_accuarcy`
+:py:meth:`~radis.lbl.broadening.BroadenFactory._check_accuracy`
```
anandxkumar, line 68: @erwanp this doesn't seem to work. `_check_accuarcy` is a method of `BroadenFactory` class. So I wonder why its not working?

erwanp: Maybe because it's a private method? Try renaming it to check_accuracy (without underscore)

erwanp: It's not very important and we can merge anyway! :)

anandxkumar: It didn't work. Maybe because of it being a private method. Sure lets merge it :)

PR #274, opened 2021-05-31
Gpu widget

dcmvdbekerom:
### Description
This PR implements two major upgrades:

1. GPU emulation -- currently, the GPU code is entirely separate from the CPU code in RADIS. This is undesirable, because the development for the GPU/CPU might diverge at some point in this way. Implementing GPU emulation allows the GPU code to be ran entirely on the CPU. To be more precise: The same c-code is compiled for GPU or CPU respecively. This is great for debugging, as discrepancies with the "OG" CPU code can now be resolved more easily -- if there is a discrepancy, at the very least, we can now rest assured that the GPU itself is not part of the problem.

But of course it would be better if there was no discrepancy between the "OG" CPU and GPU code in the first place. My hope is that at some point, "gpu emulation" would be the default way to run CPU calculations. I think that this makes the possibility of diverging codes for GPU/CPU exponentially smaller and ensures that new developments to the RADIS code will be available for both CPU and GPU.

2. Slider widgets -- The GPU code is super fast, but limited by the initialization step. Currently there is no way in RADIS to execute the gpu iteration step separately from the gpu initialization step (both are called from lbl.factory.eq_spectrum_gpu()). This PR adds a way to add slider widgets to a plot for any of the quantities that can be updated in the gpu_iterate step. E.g., you can change temperature or pressure this way, leveraging the super fast GPU calculations. Applying a slit function is now also implemented in the GPU. This might be particularly interesting for the fit room procedures.

Below is a test script to try both features:
```
import radis

sf = SpectrumFactory(1900, 2300, # cm-1
                  molecule='CO',
                  isotope='1,2,3',
                  wstep=0.002,
                  )

sf.fetch_databank('hitemp')

emulate_gpu = True

s = sf.eq_spectrum_gpu(Tgas=1200, #K 
                       pressure=0.2, #bar 
                       mole_fraction=0.1,
                       path_length= 2.0, #cm
                       emulate=emulate_gpu,
                       )

s.conditions['slit_FWHM'] = 0.0
s.plot('transmittance',
       sliders = {'slit_FWHM':(0.0,2.0),
                  'Tgas':(300,2000),
                  'pressure_mbar':(10,1200)})
```

This PR is a draft at the moment because the "OG" CPU code and the GPU/emulated GPU code still diverge. Hope to solve this soon, which should be easier now that we can emulate GPU on the CPU.


erwanp:
@CorentinGrimaldi you'll want to have a look too!!

dcmvdbekerom:
`isort` was passing locally but failing on travis. Had to run `pre-commit clean` and running `pre-commit run --all-files` solved the issue.

erwanp:
Tried. On first try after switching to your branch (no reinstall or anything) I got : 
```

  File "d:\github\radis\radis\lbl\gpu.py", line 311, in gpu_init
    from radis_cython_extensions import set_init_params

ImportError: cannot import name 'set_init_params' from 'radis_cython_extensions' (d:\github\radis\radis_cython_extensions.cp38-win_amd64.pyd)
```

Reinstalling (`pip install -e .`)  I get  : 

```
    LINK : fatal error LNK1104: impossible to open file 'D:\GitHub\radis\radis_cython_extensions.cp38-win_amd64.pyd'
    C:\Users\erwan\Anaconda3\lib\site-packages\Cython\Compiler\Main.py:369: FutureWarning: Cyth              on directive 'language_level' not set, using 2 for now (Py2). This will change in a later relea              se! File: D:\GitHub\radis\radis\cython\radis_cython_extensions.pyx
      tree = Parsing.p_module(s, pxd, full_module_name)
    error: command 'C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\BuildTools\\VC\\Too              ls\\MSVC\\14.16.27023\\bin\\HostX86\\x64\\link.exe' failed with exit status 1104
```

Is `radis\radis_cython_extensions.cp38-win_amd64.pyd` a missing file ? Note that it's also not in `radis/radis/cython` but directly in `radis/radis`

dcmvdbekerom:
Hmm.. couple of thoughts; I used `pip install -e .[dev]`. If that doesn't work you can try to run `python setup.py build` first, you may have to delete the file `radis_cython_extensions.cpp` in the `radis.cython` folder first. I don't know why the .pyd file goes directly in the RADIS folder but it does that for me as well so that shouldn't be the the source of the error

erwanp:
Did : 
```
python setup.py build 
```
works properly, ccp file is generated
Followed by : 
```
python setup.py develop  # (or pip install -e .[dev])     
``` 
still fails with the same LINK problem. 
> LINK : fatal error LNK1104: impossible to open file 'D:\GitHub\radis\radis_cython_extensions.cp38-win_amd64.pyd'
error: command 'C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\BuildTools\\VC\\Tools\\MSVC\\14.16.27023\\bin\\HostX86\\x64\\link.exe' failed with exit status 1104


-- 

Further research : 

- https://stackoverflow.com/a/16413193/5622825   
> the search path for foo.pyd is PYTHONPATH, not the same as the path that Windows uses  
- https://stackoverflow.com/questions/35213486/cython-link-fatal-error-lnk1104-cannot-open-file-atls-lib

Spent an hour trying to figure out how to fix this Link path problem on Windows, but no results at this point. Skipping for the moment. RADIS-Cython will be Linux only for some time ;)

dcmvdbekerom:
This is weird, since I'm also on Windows and had no such issues. Did the cython installation work for you earlier (the one with the fast add_at())?

erwanp:
Hello !! Fixed the error ; an __init__.py was missing. 

I got the Sliders working, the GPU code on CPU... And I also ran my first GPU code with RADIS (@pkj-m I never did during your GSOC!) 
Here is a real-time output of your example  : 

https://user-images.githubusercontent.com/16088743/121789152-e24a7300-cbd3-11eb-82b2-1a540e8493ab.mp4

It took ~30s to initialize the first calculation (loading data to GPU), but then every spectrum was computed almost instantaneously. Amazing ! 


---

Comment on the implementation : 

You've currently embedded the slider feature in the Spectrum object, with some global variables that call the Factory. I'd like if we could keep the Spectrum object itself as independant as possible from the RADIS line-by-line engine : 
- some other codes use the Spectrum class for their own post-processing ([pyspecair](https://spectralfit.gitlab.io/specair/#) for UV/VIS spectra, @gugzy or @minouHub to post-process experimental data that won't necessarily be compared to RADIS ever)
- in other words, it is meant to be as "line-by-line code agnostic" as possible
- and in the end, we may end-up making the Spectrum class itself an independant library different from the RADIS line-by-line code 

I think that this Slider feature rather belongs to a special method of the SpectrumFactory. What about : 
- adding an argument "slider" directly in SpectrumFactory.eq_spectrum_gpu() ? (you may be missing the slit parameter at this point, but you can read it from the sliders or we could also add it as a valid input parameter) 
- adding a "fitting" (or "explore_range" ?) method in the SpectrumFactory 
```
SpectrumFactory.fitting(self, sliders):
    # assuming .eq_spectrum_gpu has already been called , you'll need the output spectrum "s". We could always keep the last spectrum stored in the Factory: self.last_spectrum() ? 
    ... 

    # copy/paste most of Spectrum.plot()  code, with the update_plot() function here or as another SpectrumFactory method, 

    # init values are gotten with : self.last_spectrum.get()          # Factory calls the Spectrum functions and not the other way round
```

Anyway, this is converging to some nice fitting tools !

erwanp:
Just for fun. HITEMP-2010 CO2, 1.8 M lines in the 2000 - 2400 cm-1 range. 

```

sf = SpectrumFactory(2000, 2400, # cm-1
                  molecule='CO2',
                  isotope='1,2,3',
                  wstep=0.002,
                  verbose=3,
                  )
sf.load_databank("HITEMP-CO2-DUNHAM")   # my local database, waiting for #281 and auto-fetching of HITEMP

s = sf.eq_spectrum_gpu(Tgas=1200, #K
                       pressure=0.2, #bar
                       mole_fraction=0.1,
                       path_length= 2.0, #cm
                       )
s.conditions['slit_FWHM'] = 0.0
s.plot('transmittance',
        sliders = {'slit_FWHM':(0.0,2.0),
                  'Tgas':(300,2000),
                  'pressure_mbar':(10,1200)})
```

https://user-images.githubusercontent.com/16088743/121789972-f34ab280-cbda-11eb-99b0-d7ba019d80d5.mp4


👏  @dcmvdbekerom @pkj-m

erwanp (approved):

erwanp (approved):

radis/lbl/gpu.py
```diff
@@ -14,605 +16,731 @@ class initData(ctypes.Structure):
...
+    N_G,
+    N_L,
+    iso,
+    v0,
+    da,
+    gamma,
+    na,
+    S0,
+    El,
+    Mm_arr,
```
erwanp, line 390: @dcmvdbekerom can you easily pass `dict` as arguments with Cython ? (I hope some, all Python is based on dictionnaries). 
It may be more interesting to use 
```
iso.map(Mm_dictionary)
```
with `Mm_dictionary = {iso:M} ` than 
```
Mm_dictionary.take(iso)
```

- No need to ensure the index are the same  (in your case, if user asks for iso='3' I assume you still have to generate molar_mass & abundance arrays with [1, 2, 3].  Doesn't cost much to transport these arrays though. For partition functions I don't know. 
- it may work with non-integer names of isotopes : HITRAN started using 'A' and 'B' for CO2(11, 12).   No big deal though, we could pre-process this. 

As you see it's very minor. 
@gagan-aryan I wonder if you could try to benchmark the two approaches in terms of performance (I expect very similar results) ? For instance with [perfplot](https://github.com/nschloe/perfplot).

gagan-aryan: I will try it out.

dcmvdbekerom: Dictionaries in cython are no problem, but it has to be arrays when passing to the GPU.
But this conversion can be done inside gpu_init() so that you can pass dicts as argument to gpu_init().
As I noted in @gagan-aryan PR, we should not just do this for isotopes but also for species while we're at it.

erwanp: Ok good ! 

I'd keep it single-species for the moment ; lineshape parameters may vary a lot and a shared LDM may not be the best; we would better need a LDM "per block" which will require some changes anyway

erwanp:
(just did a big rebase of this PR on-to the latest develop branch to integrate latest changes; fix conflicts)

dcmvdbekerom:
@erwanp @minouHub @gagan-aryan for gpu update functions we need to recalculate the partition function each frame. What is the best way to do this in RADIS currently? I was hoping there is a method of the spectrum object that can do this, ideally something like spectrum.recalc_Q(T) which returns an array of partitium sums for all species/isotopologues currently in the spectrum.

Additionally, the function base.calc_S0() stopped working after recent pushes, with the following traceback:
```
  File "C:\Users\dcmvd\Documents\GitHub\radis-test\gpu-sliders\test_spectrum_gpu.py", line 18, in <module>
    emulate=emulate_gpu,
  File "c:\users\dcmvd\documents\github\radis\radis\lbl\factory.py", line 1058, in eq_spectrum_gpu
    self.calc_S0()
  File "c:\users\dcmvd\documents\github\radis\radis\lbl\base.py", line 2067, in calc_S0
    id_set = df0.id.unique()
  File "C:\Users\dcmvd\AppData\Local\Programs\Python\Python37\lib\site-packages\pandas\core\generic.py", line 5465, in __getattr__
    return object.__getattribute__(self, name)
AttributeError: 'DataFrame' object has no attribute 'id'
```

The function base.calc_S0() looks very hacky so I'm open to suggestions how to improve it.

erwanp:
Partition function [answered on Slack](https://radis-radiation.slack.com/archives/C9F7X0ZAT/p1634433478029400)

For calc_S0 and the missing `id` column; it's now as an attribute of thE DataFrame (saves one column). This requires to update this branch (based on 0.9.x) with all the new changes of 0.10.x)  ; but that's where I broke the code last time I tried the merge. I'll have another look but not before next week

PR #283, opened 2021-06-09
Updated convolve_with_slit with warnings if instrumental function has large near-zero wings

gh4ag:
updated warnings on convolve_with slit and the documentaiton of the function, modified doc of spectrum to point to the doc of convolve_with_wlit

### Description

PR in order to update the documentation and the errors of radis.tools.slit.py convolve_with_slit.

Convolution intervene in mainly 2 places:
- convolution with physical broadening of spectral features during simulation
- convolution with instrumental function to fit simulation to experimental spectra

I did my best to explain to the user what to expect out of the convolution product in the case of intrumental functions.

It is important to tell users how convolution could affect data outside of the know spectral range. 

I modified apply_slit to point to the Notes I wrote in convolve_with_slit.

I transformed the AssertError of convolve with slit with several warnings, that look both at the spectral range and the FWHM of the input experimental slit. IMO, it will better guide users to what they should do in order to improve the output of the convolution.

### DIFF
--- a/radis/tools/slit.py
+++ b/radis/tools/slit.py
@@ -661,6 +661,11 @@ def convolve_with_slit(
     w_range = w[-1] - w[0]
     w_slit_range = w_slit[-1] - w_slit[0]
     slit_FWHM = fwhm(w_slit, I_slit)
+    print(slit_FWHM)
+    w_range = abs(w[-1] - w[0])
+    w_slit_range = abs(w_slit[-1] - w_slit[0])
+    slit_FWHM = get_FWHM(w_slit, I_slit)
+    print(slit_FWHM)
 
     if np.abs(w_slit_range) >10*slit_FWHM:
         warn(
@@ -808,15 +813,27 @@ def get_FWHM(w, I, return_index=False):
     """
     # TODO: Linearly interpolate at the boundary? insignificant for large number of points
 
-    upper = np.argwhere(I >= I.max() / 2)
+    half = I.max() / 2
+    I_offset = I - half
+    upper = np.argwhere(I_offset >= 0)
 
     xmin = upper.min()
     xmax = upper.max()
 
+    def get_zero(x1, y1, x2, y2):
+        """
+        Linear interpolation on data to get zero
+        Return location of zero by solving for f(x) = (y2-y1)/(x2-x1) * (x-x1) + y1 = 0 
+        """
+        return x1 - y1*(x2-x1)/(y2-y1)
+
+    wmin = get_zero(w[xmin-1], I_offset[xmin-1], w[xmin], I_offset[xmin])
+    wmax = get_zero(w[xmax], I_offset[xmax], w[xmax+1], I_offset[xmax+1])
+
     if return_index:
-        return abs(w[xmax] - w[xmin]), xmin, xmax
+        return abs(wmin - wmax), xmin, xmax
     else:
-        return abs(w[xmax] - w[xmin])
+        return abs(wmax - wmin)
 
 
 def get_effective_FWHM(w, I):
--- a/radis/tools/slit.py
+++ b/radis/tools/slit.py
@@ -647,7 +647,6 @@ def convolve_with_slit(
     w_range = abs(w[-1] - w[0])
     w_slit_range = abs(w_slit[-1] - w_slit[0])
     slit_FWHM = get_FWHM(w_slit, I_slit)
-    print(slit_FWHM)
 
     if w_slit_range >10*slit_FWHM:
         warn(
--- a/radis/tools/slit.py
+++ b/radis/tools/slit.py
@@ -673,10 +673,10 @@ def convolve_with_slit(
     if np.abs(w_range) < np.abs(w_slit_range):
         if mode == "valid":
             raise AssertionError(
-                f"Slit spectral range > Spectrum spectral range and mode is set to valid!\n" \
-              + f"Slit function is provided with a spectral range ({w_slit_range:.1f} nm) larger than " \
-              + f"the spectral range of the spectrum ({w_range:.1f} nm). The output spectrum is therefore empty! " \
-              + f"Set mode to same in this case. However, we recommend truncating the input slit function. "
+              f"Slit function is provided with a spectral range ({w_slit_range:.1f} {waveunit}) larger than " \
+              + f"the spectral range of the spectrum ({w_range:.1f} {waveunit}) : the output spectrum will therefore be empty " \
+              + f"as boundary effects of the convolution are automatically discarded. If you still want to apply the slit "
+              + f"despite potential boundary effects, set `mode='same'`. However, we recommend truncating the input slit function if possible, or compute the spectrum on a larger spectral range. "
               + f"See radis documentation about convolve_with_slit for more information."
             )
         else:
--- a/radis/tools/slit.py
+++ b/radis/tools/slit.py
@@ -78,7 +78,7 @@ def get_slit_function(
     verbose=True,
     auto_recenter_crop=True,
     *args,
-    **kwargs
+    **kwargs,
 ):
     """Import or generate slit function in correct wavespace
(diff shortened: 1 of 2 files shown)


radis/tools/slit.py
```diff
@@ -583,16 +635,50 @@ def convolve_with_slit(
...
+        zero_crossings_i = np.where(zero_crossings)[0]
+        return [lin_interp(x, y, zero_crossings_i[0], half),
+                lin_interp(x, y, zero_crossings_i[1], half)]
+    def fwhm(x,y):
+        hmx = half_max_x(x,y)
+        return hmx[1] - hmx[0]
+
+    w_range = w[-1] - w[0]
+    w_slit_range = w_slit[-1] - w_slit[0]
+    slit_FWHM = fwhm(w_slit, I_slit)
```
erwanp, line 655: RADIS already has a [get_FWHM()](https://github.com/radis/radis/blob/develop/radis/tools/slit.py#L682) function, we could use it directly if it also works as you expect (less code = easier to maintain): 

```suggestion
    w_range = abs(w[-1] - w[0])
    w_slit_range = abs(w_slit[-1] - w_slit[0])
    slit_FWHM = get_FWHM(w_slit, I_slit)
```

(note : you can commit the suggestion directly from the GitHub interface)

radis/tools/slit.py
```diff
@@ -583,16 +635,50 @@ def convolve_with_slit(
...
         )
 
+    if np.abs(w_range) < np.abs(w_slit_range):
+        if mode == "valid":
+            raise AssertionError(
+                f"Slit spectral range > Spectrum spectral range and mode is set to valid!\n" \
+              + f"Slit function is provided with a spectral range ({w_slit_range:.1f} nm) larger than " \
+              + f"the spectral range of the spectrum ({w_range:.1f} nm). The output spectrum is therefore empty! " \
+              + f"Set mode to same in this case. However, we recommend truncating the input slit function. "
+              + f"See radis documentation about convolve_with_slit for more information."
```
erwanp, line 672:
```suggestion
              f"Slit function is provided with a spectral range ({w_slit_range:.1f} {waveunit}) larger than " \
              + f"the spectral range of the spectrum ({w_range:.1f} {waveunit}) : the output spectrum will therefore be empty " \
              + f"as boundary effects of the convolution are automatically discarded. If you still want to apply the slit "
              + f"despite potential boundary effects, set `mode='same'`. However, we recommend truncating the input slit function if possible, or compute the spectrum on a larger spectral range. "
              + f"See radis documentation about convolve_with_slit for more information."
```

Note : spectra can be in wavenumbers and not in nm. 
Slit & spectrum are always given in the same waveunit, that's ensured by the Spectrum.apply_slit method.

radis/tools/slit.py
```diff
@@ -561,12 +571,54 @@ def convolve_with_slit(
...
+    
+    ::
+    
+        I_slit  ^
+                |     |   _   |
+                |     |  / \\  |
+                |     | /   \\ |
+               0|_____|/     \\|________
+              --|-----         --------->
+                    cut       cut        w_slit
```
erwanp, line 603: 👍

radis/tools/slit.py
```diff
@@ -500,11 +500,14 @@ def convolve_with_slit(
...
+        Returns a convolved spectrum on a valid range. 
+
+        Check Notes section below for more details about its implementation.
 
         This function is called directly from a :py:class:`~radis.spectrum.spectrum.Spectrum`
         object with the :py:meth:`~radis.spectrum.spectrum.Spectrum.apply_slit` method.
-
+    
 
     .. warning::
```
erwanp, line 512: @gh4ag  note the existing warning here, which also [appears in the docs](https://radis.readthedocs.io/en/latest/source/radis.tools.slit.html#radis.tools.slit.convolve_with_slit).   It's not in [apply_slit](https://radis.readthedocs.io/en/latest/source/radis.spectrum.spectrum.html#radis.spectrum.spectrum.Spectrum.apply_slit, though. 

The fact it's not visible enough makes me wonder if the additional information you wrote should be added in the Notes of the function docstring , or elsewhere, maybe in the [Spectrum object guide](https://radis.readthedocs.io/en/latest/spectrum/spectrum.html#apply-instrumental-slit-function) ?

erwanp (changes requested):

radis/tools/slit.py
gh4ag: Regarding get_FWHM(), I saw that the interpolation was not implemented yet, but for the slit function it is a must have.

The slit function that we provide to convolve_with_slit is made of not many points, and it follows the spirit of this functions since we ask users to cut the input slit function to its minimum spectral range.

I dit a test on this slit function and I got
[slit_fig.pdf](https://github.com/radis/radis/files/6629252/slit_fig.pdf)

Initially (FWHM): 0.2343906011052468
With get_FWHM: 0.16641365720090562

A gaussian fit yields something very close to 0.2389 nm. The main difference comes from interpolation.

erwanp: What's the reference for "initial = 0.234" ? Your slit file seems to have a resolution of ~0.07 nm so 0.166 and 0.234 nm are all within the error bar. 

You can use [plot_slit()](https://radis.readthedocs.io/en/latest/auto_examples/plot_slit_function.html#sphx-glr-auto-examples-plot-slit-function-py), it's a handy function to plot the slit & show the different FWHM (uses get_FWHM internally). 

Implementation wise : 

In convolve_with_slit we cannot afford to fit a gaussian all the time ; many slit functions aren't Gaussian (see example above) ; and it can be a bit costly although performance is key in fitting procedures . 

I'd suggest a very minimal test (compare w_slit_range to w_spectrum ?) first to determine if there can be a problem ; if so, then switch to a warning /error where performance doesn't matter anymore and the goal is to be be as informative as possible for the user : in this condition, you can afford fitting a Gaussian, or whatever seems best for you.

gh4ag: I disagree with you estimation. By hand, I find 0.251 nm.

gh4ag: "cannot afford to fit a gaussian" -> I would not even recommend it

radis/tools/slit.py
gh4ag: So I replied in the wrong thread apparently.  :(

radis/tools/slit.py
gh4ag: I modified the get_FWHM to work better: see 
b9abfe1

radis/tools/slit.py
gh4ag: See 
b9abfe1

Solves the problem

radis/tools/slit.py
gh4ag: Could you raise another issue for this ? I would like to use the output of this PR as soon as possible for my work :)

radis/tools/slit.py
```diff
@@ -714,15 +787,27 @@ def get_FWHM(w, I, return_index=False):
     """
     # TODO: Linearly interpolate at the boundary? insignificant for large number of points
 
-    upper = np.argwhere(I >= I.max() / 2)
+    half = I.max() / 2
+    I_offset = I - half
```
erwanp, line 791: I do not understand why it's needed : the slit is zero on the sides.

gh4ag: This steps makes interpolation easier conceptually to find the FWHM by solving the question "when do my graph crosses the y=0 axis". The function get_FWHM assumes the slit to have only 1 peak and to go to zero monotously away from the peak. You can modify it if you want, to find the y=half - but then you have to change the helper function get_zero as well (should be easy). 

I don't think I understood your comment. This function doesn't look at the sides of the slit here. It looks at the FWHM, so close to the center of the slit.

gh4ag: I have just implemented the TODO XD

># TODO: Linearly interpolate at the boundary? insignificant for large number of points

PR #285, opened 2021-06-12
fix conda install dependencies,

erwanp:
and mention how to install conda environment on the docs


erwanp:
(test failed because they rely on HITEMP-OH ; HITRAN servers are in maintenance this weekend. Retry Monday!)

PR #286, opened 2021-06-14
Add: Profiler Class 

anandxkumar:
### Description

This pull request adds class `Profiler()` which contains dictionary `dict_time` which will contain all parameters value which is bottleneck or important for broadening step for all optimizations. Also, it will shift all print statements from broadening.py to Time class.


anandxkumar:
@erwanp, based on your suggestion on slack, I have shifted all the print statements from `_calc_lineshape()` function to `Time()` class in `loader.py`. `self.time.dict_time` will be the dictionary comprising of all the parameters that will necessary for my analysis. What are your thoughts on this? If a green flag I will implement this for all the functions in the broadening step for both Legacy and LDM.
Also: Mind the additional print statement, and variable declaration. Will take care of them later.

dcmvdbekerom:
One very quick comment: I would give it a bit longer name such as "PerformanceTime" or something (just one example). As "Time" may be ambiguous.

anandxkumar:
@dcmvdbekerom right! Maybe `CalcTime` or `PerformanceTime` as you suggested?

radis/lbl/broadening.py
```diff
@@ -1339,25 +1341,10 @@ def _calc_lineshape(self, dg):
...
-                        "... Calculated Lorentzian profile in {0:.1f}s".format(t11 - t1)
-                    )
-                    printg(
-                        "... Calculated Gaussian profile in {0:.1f}s".format(t12 - t11)
-                    )
-                    printg("... Convolved both profiles in {0:.1f}s".format(t2 - t12))
-                elif broadening_method == "fft":
-                    raise NotImplementedError("FFT")
+            _time["t2"] = time()
+            self.time._calc_lineshape_time(
```
erwanp, line 1345: Time()  should be a very general object and shouldn't have methods specific for, for instance the lineshapes. 
It could be something like : 


```
if broadening_method == "voigt":                 
   Self.time.stop(level=3, "... Calculated Voigt profile (jit={1})") 
```

Which would handle the verbose, print, and store?

anandxkumar:
@erwanp in order to print the statements in a proper manner together (rather than across the function), we will require `verbose`, `initial time`, `final time`, and the `statement` itself. Though it will print and store the data using a single method it will take a lot of space in the function itself because we will require to execute this method for the individual print statement which may take up more space than what was original. Is it an issue? In case you have a better idea do let me know. I added a save variable that will save the parameter in the `dict_time` dictionary with a rounded-off value using the same `round_off` function that I created for `wstep`. So for all the necessary parts we can save that parameter and using Spec Database these conditions can be stored for future analysis.

radis/lbl/factory.py
```diff
@@ -477,6 +478,12 @@ def __init__(
             cutoff = 0
         self.params.cutoff = cutoff
 
+        # Time Based variables
+        self.verbose = verbose
```
anandxkumar, line 482: @erwanp I'm declaring the CalcTime object here because in `loader.py` the `self.verbose` values remains True and is not updated by the SpectrumFactory value, so it will never be never satisfy `self.verbose>=verbose` condition to print the statement. Declaring here will store the updated value in self.verbose thus we don't have to send it as an argument in `start` or `stop`

radis/lbl/loader.py
```diff
@@ -496,6 +497,44 @@ def __init__(self):
...
+        # Dev: Init here to be found by autocomplete
+        self.initial = {}
+        self.dict_time = {}
+        self.verbose = verbose
+
+    def start(self, key, verbose):
+        if __debug__:
+            self.initial[key] = {"start_time": time(), "verbose": verbose}
+
+    def stop(self, key, details=""):
```
anandxkumar, line 527: Adding argument `details` here because there are many instances where we  need a particular value to print like `jit` which is declared after the `start` point and also makes easier to print statements in `if elif else` type condition by having a common starting point!

radis/lbl/factory.py
erwanp: Are you sure ? SpectrumFactory inherits from the Loader, so Loader.verbose should actually be the same as SpectrumFactory.verbose ?

radis/lbl/factory.py
```diff
@@ -745,6 +751,9 @@ def eq_spectrum(
                 "radis_version": get_version(),
             }
         )
+        time_calc = {}
+        time_calc["time"] = self.time.dict_time
+        conditions.update(time_calc)
```
erwanp, line 756: In one line and add it to the dictionary above ?

Warning : make sure to make a copy/deepcopy ; else each Spectrum will still be connected to the Factory conditions (check if not done already?).

erwanp (changes requested):

radis/lbl/factory.py
anandxkumar: If you see the line [here](https://github.com/radis/radis/blob/4ebfa27ac85dde933df1f523b401ba8130c5b389/radis/lbl/loader.py#L555) we are specify `self.verbose = True`. But if I remove this line so it should automatically take the value from `SF.__init__` , in that case on declaring `self.time = CalcTime(self.verbose)` in `DatabankLoader.__init__`, it gives the following error:

  ```
File "/home/pipebomb/Desktop/radis/radis/lbl/loader.py", line 620, in __init__
    self.time = CalcTime(self.verbose)

AttributeError: 'SpectrumFactory' object has no attribute 'verbose'
```
which is indicating self.verbose hasn't been declared in `self` object yet.

File: radis/lbl/loader.py [hunk 5]
```diff
@@ -496,6 +497,44 @@ def __init__(self):
         self.total_lines = 0  #: int : number of lines in database.
 
 
+class CalcTime(ConditionDict):
+    """A class to hold Spectrum calculation time dependent parameters, under the attribute
+    :py:attr:`~radis.lbl.loader.DatabankLoader.time.dict_time` of
```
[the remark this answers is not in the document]
anandxkumar: `profiler` seems better!

radis/lbl/factory.py
erwanp: I see! That's because CalcTimer keeps the verbose level it had when initialized. 
You could reset this at the beginning of a calculatino ; for instance [here for eq_spectrum](https://github.com/radis/radis/blob/develop/radis/lbl/factory.py#L653). It would also allow you to change the verbose level in between 2 calculations of a same Factory and immediatly see the changes. 

Another option would be to fully Reset the Timer, creating a new instance of the CalcTimer class, at the beginning of the calculation. 
- you can init it with the current verbose level of the Factory.
- no need to copy() the output when storing them in the Spectrum conditions, as I mentionned in another comment. 
- makes sure all keys are properly reset in between 2 calculations

radis/lbl/loader.py
```diff
@@ -496,6 +497,44 @@ def __init__(self):
...
+
+    def stop(self, key, details=""):
+        if __debug__:
+            items = self.initial.pop(key)
+            # print("ITEMS: ",items)
+            self.dict_time[details] = round_off(time() - items["start_time"])
+            # print("items['verbose']: ",items['verbose'])
+            # print("self.verbose: ",self.verbose)
+            if self.verbose >= items["verbose"]:
+                printg(details, "in {0:.1f}s".format(self.dict_time[details]))
```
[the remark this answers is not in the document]
anandxkumar: Ok :+1:

radis/lbl/loader.py
```diff
@@ -496,6 +497,44 @@ def __init__(self):
...
+
+    def start(self, key, verbose):
+        if __debug__:
+            self.initial[key] = {"start_time": time(), "verbose": verbose}
+
+    def stop(self, key, details=""):
+        if __debug__:
+            items = self.initial.pop(key)
+            # print("ITEMS: ",items)
+            self.dict_time[details] = round_off(time() - items["start_time"])
```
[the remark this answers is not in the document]
anandxkumar: We are using `key` as `'a', 'b', 'c'`. So it won't be helpful in understanding what the key refers to, whereas the details themselves will be easier to understand the entry. 
Never mind the round_off, removing it!

radis/lbl/factory.py
anandxkumar: What about setting `self.time.dict_time = {}` after every Spectrum calculation? This way we can keep each Spectrum calculation to itself even when multiple Spectrums are created using the same SF object?

radis/lbl/loader.py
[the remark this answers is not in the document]
erwanp: Maybe add more explicit keys then ? 
They appear in the code ; it should be quite clear what Timer is being Started / Stopped, so keys should be explicit ; but short; to build your Benchmark Dataframe. 

Details can be one line or more so I wouldn't use that for keys.

anandxkumar:
Modified `_calc_lineshape()` and `_apply_lineshape()` in `broadening.py` with new `start`, `stop` and `print` and for a normal calculation using optimization=None and verbose = 3, get the following outcome: 
```
Scaling equilibrium linestrength
Scaled equilibrium linestrength in 0.01s
Calculated lineshift in 0.00s
... Calculate broadening HWHM in 0.01s
... Calculating line broadening (45000 lines: expect ~ 9.00s on 1 CPU)
...... Initialized vectors in 0.20s
...... Calculated Voigt profile (jit=True) in 4.95s
...... Initialized vectors in 0.12s
...... Get closest matching line & fraction in 0.00s
...... Aggregate center lines in 0.63s
...... Aggregate wing lines in 0.01s
Calculated line broadening in 5.95s
Calculated other spectral quantities in 0.01s
Spectrum calculated in 6.00s (before object generation)
Generated Spectrum object in 0.01s
```
`verbose=2` with `...` , `verbose=3` with `......` and so on

erwanp:
Super nice, the implementation in the main code is so much cleaner than my original `t0 = time()` everywhere. 
For consistency, your CalcTime object can be renamed Profiler ?

I think it's ready, and can be extended to all functions !

radis/lbl/loader.py
```diff
@@ -496,6 +497,61 @@ def __init__(self):
         self.total_lines = 0  #: int : number of lines in database.
 
 
+class CalcTime(object):
```
erwanp, line 500: Detail  : the class itself could be moved outside of lbl/loader.py ; maybe in radis/misc/debug.py or a new radis/misc/profiler.py ?

anandxkumar: Indeed. A new file will be better!

radis/lbl/broadening.py
```diff
@@ -697,26 +696,48 @@ def _whiting_jit(w_centered, wl, wv):
...
     # ... 20.5.s > 16.5s) on the total eq_spectrum calculation
     # ... w_wv is typically a (10.001, 1997) array
-    # from time import time
-    a = time()
+    #   from time import time
+    #   print("w_centered : ", w_centered)
+    #   print("w_centered shape : ", np.shape(w_centered))
+    #   print("wl shape : ", np.shape(wl))
+    #   print("wv shape : ", np.shape(wv))
+    # a = time()
```
erwanp, line 704: @anandxkumar can you entirely remove the commented prints ? (you can get them back from the Git history if needs be!)

radis/misc/profiler.py
```diff
@@ -41,23 +43,22 @@ def __init__(self, verbose):
...
         if __debug__:
             items = self.initial.pop(key)
             self.dict_time[key] = time() - items["start_time"]
             if self.verbose >= items["verbose"]:
-                self._print(key, items["verbose"], items["details"])
+                self._print(key, items["verbose"], details)
 
     def _print(self, key, verbose, details):
-        if verbose == 1:
+        if verbose == 1 or verbose == True:
```
erwanp, line 61: should be uncessary : normally `True==1` is True in Python

erwanp (approved):

anandxkumar:
@erwanp should we replace statements like:
```
if self.verbose >= 2:
                    print("{0}: No rovibconstants found".format(self.get_fullname()))
```
with statement like 

`self.profiler._print(None, 2, "{0}: No rovibconstants found".format(self.get_fullname()))`

I think it will be unnecessary for such cases and maybe we can leave it as it is?

erwanp:
Hello ! Profiler should run only during the execution , so anything that is called before eq_spectrum() shouldn't be added in the Profiler class. Just keep the verbose ? 
You should only be replacing the instances of "t0 = time()" etc. And do not add the loading of the database in there ; they won't interfere with the calculatino method anyway

radis/misc/profiler.py
```diff
@@ -0,0 +1,74 @@
...
+                "verbose": verbose,
+            }
+
+    def stop(self, key, details=""):
+        if __debug__:
+            items = self.initial.pop(key)
+            self.dict_time[key] = time() - items["start_time"]
+            if self.verbose >= items["verbose"]:
+                self._print(key, items["verbose"], details)
+
```
dcmvdbekerom, line 59: Looks like you don't use self.verbose. Just a thought, but you could start with self.verbose = 0, and every time start() is called self.verbose is incremented, and when stop() is called self.verbose is decremented. This way you don't need to explicitly specify the verbosity when profiling.

dcmvdbekerom:
Hey, just dropping in to have a look.. Looks really nice! This is going to be very helpful moving forward!
Perhaps printing the time first (right after the dots) and then the details would make it easier to compare times.
Also it would be nice to have percentages that show the relative time within a given verbosity level

radis/misc/profiler.py
erwanp: I'd call the inside key "level"  rather than verbose. And have "if... Verbose>=level then print ()"

radis/misc/profiler.py
```diff
@@ -0,0 +1,126 @@
...
+            if key == None:
+                printg("..." * (verbose - 1), details)
+            else:
+                printg(
+                    "..." * (verbose - 1),
+                    "{0:.2f}s -".format(self.dict_time[key][0]),
+                    details,
+                )
+
+    def percentage_distribution(self):
```
anandxkumar, line 78: @dcmvdbekerom added percentage time distribution for steps involved in each verbose level. I'm storing it in a Spectrum object and the result will be as follow :

`spectrum.get_conditions()['profiler']` :

```
{1: [('spectrum_calculation', [100.0, 1.119952917098999])],
 2: [('spectrum_calc_before_obj', [56.97151990399073, 1.061187982559204]),
  ('calc_noneq_population', [19.74545153399409, 0.36779141426086426]),
  ('check_non_eq_param', [9.267285489209133, 0.17261838912963867]),
  ('fetch_energy_6', [7.526554291483005, 0.14019441604614258]),
  ('generate_spectrum_obj', [3.1442882049576717, 0.05856752395629883]),
  ('calc_line_broadening', [1.9036095507002235, 0.03545784950256348]),
  ('calc_weight_trans', [0.9310497852245717, 0.017342329025268555]),
  ('non_eq_linestrength', [0.14664811709423994, 0.0027315616607666016]),
  ('calc_other_spectral_quan', [0.11783560844632739, 0.0021948814392089844]),
  ('calc_emission_integral', [0.1084405034496291, 0.0020198822021484375]),
  ('calc_hwhm', [0.10745491341591554, 0.002001523971557617]),
  ('calc_lineshift', [0.029862098034464674, 0.0005562305450439453])],
 3: [('part_function', [93.1809938338774, 0.3619859218597412]),
  ('DLM_convolve', [4.159662498841589, 0.0161592960357666]),
  ('population', [1.4629428122355217, 0.005683183670043945]),
  ('corrected_population_se', [0.6323236898293653, 0.0024564266204833984]),
  ('precompute_DLM_lineshapes', [0.37977472509600235, 0.0014753341674804688]),
  ('DLM_closest_matching_line', [0.08273049926137867, 0.00032138824462890625]),
  ('DLM_Distribute_lines', [0.050571165720605354, 0.0001964569091796875]),
  ('map_part_func', [0.04965057411161376, 0.00019288063049316406]),
  ('DLM_Initialized_vectors', [0.0013502010265210168, 5.245208740234375e-06])]}
```
So its ```
{        {
            verbose_level: [
                ("key_1", [percentage_time, time]),
                ("key_2", [percentage_time, time]),
            ]
        }
```
and in sorted manner!
Your feedback!

**Comment by anandxkumar** [comment]
@erwanp, one of the test case failed, not sure why. Can you help me understand why it failed?

**Comment by anandxkumar** [comment]
Exampe:
```
sf = SpectrumFactory(
    2284.2,
    2284.6,
    wstep=0.001,  # cm-1
    pressure=20 * 1e-3,  # bar
    cutoff=0,
    path_length=0.1,
    mole_fraction=400e-6,
    molecule="CO2",
    isotope=[1],
    db_use_cached=True,
    medium="vacuum",
    broadening_max_width=10,
    export_populations="rovib",
    verbose=3,
)
sf.warnings.update(
    {
        "MissingSelfBroadeningWarning": "ignore",
        "NegativeEnergiesWarning": "ignore",
        "LinestrengthCutoffWarning": "ignore",
        "HighTemperatureWarning": "ignore",
    }
)
sf.init_databank(
    "HITEMP-CO2-TEST"
) 
s1 = sf.non_eq_spectrum(2000, 1000)
```

print ->
```
... 0.14s - Fetched energies for all 2284 transitions
... 0.02s - calculated weighted transition moment
... 0.17s - Checked nonequilibrium parameters
... sorting lines by vibrational bands
... lines sorted in 0.0s
...... 0.37s - partition functions
...... 0.01s - populations
... 0.38s - Calculated nonequilibrium populations
...... 0.00s - map partition functions
...... 0.00s - corrected for populations and stimulated emission
... 0.00s - scaled nonequilibrium linestrength
... 0.00s - calculated emission integral
... 0.00s - Calculated lineshift
... 0.00s - Calculate broadening HWHM
... Calculating line broadening (2284 lines: expect ~ 2.28s on 1 CPU)
...... 0.00s - Precomputed DLM lineshapes (8)
...... 0.00s - Initialized vectors
...... 0.00s - Get closest matching line & fraction
...... 0.00s - Distribute lines over DLM
...... 0.02s - Convolve and sum on spectral range
...... 0.00s - Initialized vectors
...... 0.00s - Get closest matching line & fraction
...... 0.00s - Distribute lines over DLM
...... 0.02s - Convolve and sum on spectral range
... 0.04s - Calculated line broadening
... 0.00s - Calculated other spectral quantities
... 1.05s - Spectrum calculated (before object generation)
... 0.06s - Generated Spectrum object
```

File: radis/misc/profiler.py (hunk 12 above)
Comment (erwanp) [reply radis/misc/profiler.py:78]: Can you **not** store the % ? Less data on the Spectrum object eventually will be more readable / easier to serialize. 

However, the function that Prints() should show the percentages (as you've done). 
Eventually you may want a print() functino in SpectrumFactory or Spectrum ; that parses the dictionary of computed times and renders it properly. 

(Basically : storing times; and printing them; should have a different philosophy : store only the minimum ; and print which as much details informatino as the user wants )

File: radis/misc/profiler.py (hunk 11 above)
Comment (erwanp) [reply radis/misc/profiler.py:61]: @anandxkumar if you can just update the name 
```
if self.verbose >= items["verbose"]
```
to 
```
 items["level"]     or items["verbose_level"]
```

**Review by erwanp — CHANGES_REQUESTED** [review]

File: radis/misc/profiler.py (hunk 12 above)
Comment (anandxkumar) [reply radis/misc/profiler.py:78]: Actually, I'm not storing the percentage data in the Spectrum object. If you see I'm updating only the method and its time using:
```
   conditions.update(
            {
                "profiler": {
                    i: self.profiler.dict_time[i][0] for i in self.profiler.dict_time
                }
            }
        )
```
The whole calculation thing is right now stored in a different variable `sf.profiler.relative_time_percentage`, so anyone who wants the full detail can directly access it. 
I'm a little confused by the print function you have mentioned. We use the `get_conditions()` method to get the parameters, so a similar function `self.print_details` can be used to print all the details?

Comment (erwanp) [reply radis/misc/profiler.py:78]: Yes for the print, but we can always add it later depending on your needs. For now I think thr implementation is good and ready! Merging if you think it's fine

**Review by erwanp — APPROVED** [review]

File: radis/misc/profiler.py (hunk 12 above)
Comment (anandxkumar) [reply radis/misc/profiler.py:78]: Pushing one last commit!

Comment (anandxkumar) [reply radis/misc/profiler.py:78]: @erwanp I was thinking of adding a column 'name' or 'id' (eg - 'CO_9000_10000') which will be same for a single Spectrum Factory, so that it will be easier for me to distinguish between various SF in the csv generated file. The file name gets incremented when computing several spectrum from Spectrum Factory so there isn't any unique field to distinguish 1 SF to another.

Comment (erwanp) [reply radis/misc/profiler.py:78]: you can also give a name to the Spectra, even at runtime (eq_spectrum(name=...). Maybe that's as convenient ; and it's already implemented.

**Comment by minouHub** [comment]
Hi guys, 
With this new feature (that I like a lot) some stuff are still printed although I set `verbose=False`. Is it in purpose? If not, can @anandxkumar make a quick fix?

```python

T, p = 700, 0.01
wmin, wmax = 2384.05, 2384.4
s_HITEMP = calc_spectrum(
        wavenum_min=wmin,
        wavenum_max=wmax,
        mole_fraction={'CO2':0.1},
        isotope=1,
        # isotope="all",
        Tgas=T,
        pressure=p,
        path_length=10.32,
        name="HITEMP - p = {:.1f} bar, T = {:.0f} K".format(p,T),
        # databank='HITEMP-CO2',
        verbose=False,    
        wstep=0.001,
    )
```

**Output:**
```python
Scaling equilibrium linestrength
Calculating line broadening (169 lines: expect ~ 0.17s on 1 CPU)
```

PR #287, opened 2021-06-16
perf: dataframe refactors: drop id iso Qgas columns

gagan-aryan:
### Description

This pull request is to address the dataframe refactors to boost pandas' performance. The idea is to drop certain columns in the dataframe so that the overall memory that is loaded on the disc decreases. 

- [x] drop id column
- [x] single isotope - drop isotope column
- [x] drop Qgas column 

### Details

Id is dropped in `loader.py` i.e., before the dataframe is returned to other files in the `lbl` module. The case of single_isotope is handled similarly. 

For Qgas, the changes in this PR avoid loading Qgas onto the dataframe altogether and instead uses dictionary and maps. The approach is similar to the one demonstrated in this link [here.](https://stackoverflow.com/questions/20250771/remap-values-in-pandas-column-with-a-dict/41678874#41678874) [The manual benchmark on Radis benchmark](https://github.com/radis/radis-benchmark/blob/master/manual_benchmarks/test_Qgas.ipynb) shows that the new approach not only consumes lesser memory but is also faster. 

The modifications to the way we are handling Qgas and Qref reduces the memory by around 100MB (1.2GB to 1.1GB) while processing lines in the wavenumber range 2000-3000 for the case of HITEMP-CH4 databank. This also reduces the CPU time  of the calc_linestrength_method to half .


gagan-aryan:
Currently, one test of `line_survey.py` fails where [get_label_hitran](https://github.com/radis/radis/blob/4f5396024cff39a9a0d8494a206a27a000456639/radis/tools/line_survey.py#L253) tries to access the [id of a specific row](https://github.com/radis/radis/blob/4f5396024cff39a9a0d8494a206a27a000456639/radis/tools/line_survey.py#L267). I am checking on how this can be resolved.

erwanp:
Line survey will indeed be a problem  (have you tried it to see what it does?) 

Short term : can you update it to try to fetch the Metadata of the "lines" attribute, which is a copy of the SpectrumFactory df. Get lines.id  if "id" column not found. 

Long term : storing line information in the Spectrum is not really required. Instead, the Spectrum could store an index of lines and the path to the database (index of lines not even necessary if no cutoff!)  This would allow to use the line survey by fetching the database. Saves a lot of memory and disk space too. You can add this to the end of the project if you have time. Using the advanced libraries like Vaex one could stream the database during the line survey plot rather than load everything. Only con : you cannot share this spectrum with anyone that doesmt have the database, and retain the line survey method.

gagan-aryan:
Yes, I have tried out the examples of line survey. In my understanding, it plots all the lines in the range of wavenumber. 

I have implemented the short-term fix for now. Will surely add the long term at the end of the project if time permits (or even after the project period). As for the con, maybe we can make the lines of the Spectrum optional as in, a user can opt to store the line information in the spectrum or can just store the index of line and the path. 

The same tests seem to fail again in the build but all the tests pass locally. I am not sure why this is the case.

erwanp:
It may be an update of the HITRAN database. Hitran2020 is rolling out! 

I'll check

gagan-aryan:
I am not sure about the changes I have done in line_survey in the latest commit. 

Btw, this link is broken from https://radis.github.io/ is broken. 
image

erwanp:
BTW, looking at the list : abundance,  molar mass and partition function columns can be dropped in all cases, even multi isotopes. 

They should be looked up from a dictionary, but not stored in a new column 

(we can discuss it on Slack!)

gagan-aryan:
@erwanp two tests are failing (one each in broadening and calc) due to Assertionerrors in tolerance limits. I wonder if its coz of the machine error we were getting with the new approach.

radis/lbl/base.py
```diff
@@ -2035,9 +2040,21 @@ def _calc_Q(molecule, iso, state):
...
+                Qref_Qgas_ratio = {
+                    k: Qref_dict.get(k, 0) / Qgas_dict.get(k, 0)
+                    for k in set(Qref_dict) | set(Qgas_dict)
+                }
+
+                # copy iso column
+                df1["S"] = df1["iso"]
+
+                # map the ratios in the dictionary to the iso column vallues
+                df1["S"].map(Qref_Qgas_ratio)
```
erwanp, line 2057: cf Slack comment : 

you should be able to simplify a lot all of these lines. 
Eventually there could be only something like :    L2077:

```
line_strength = df1.int * df1["iso"].map(Qref_Qgas_ratio] 
```
Alternative if only one isotope : 
```
line_strength = df1.int * Qref / Qgas
```

erwanp (changes requested):

radis/lbl/base.py
gagan-aryan:
```
line_strength = df1.int * df1["iso"].map(Qref_Qgas_ratio] 
```

I don't think this is the right syntax we can use because if I do something like

```
df1["new"] = df1["iso"].map(Qref_Qgas_ratio)
```

It will give me `NaN` in all the entries of the new `new` column and hence prefilling `df["S]` with the values in `iso`. But other than that I have simplified the code by removing the redundant arrays.

radis/lbl/base.py
```diff
@@ -2124,7 +2141,9 @@ def calc_populations_eq(self, Tgas):
             molecule1 = get_molecule(id1)
             state = self.input.state
             parsum1 = self.get_partition_function_interpolator(molecule1, iso1, state)
-            df1["Qgas"] = parsum1.at(Tgas)
+            df1["nu"] = parsum1.at(
+                Tgas
+            )  # Storing Qgas in nu column to avoid loading another column
```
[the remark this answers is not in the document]
gagan-aryan: [Pre-filling would indeed be required to map.](https://github.com/radis/radis/pull/287#discussion_r657418960) Do you suggest that I just drop df1["Qgas"] at the end after it is used in the equation?

radis/tools/line_survey.py
```diff
@@ -263,24 +263,25 @@ def get_label_hitran(row, details):
...
-            label = (
-                "{molec}[iso{iso:.0f}] [{branch}{jl:.0f}]({vl:.0f})->({vu:.0f})".format(
-                    **dict(
-                        [(k, row[k]) for k in ["vu", "vl", "jl", "iso"]]
-                        + [
-                            ("molec", molecule),
-                            ("branch", _fix_branch_format[row["branch"]]),
-                        ]
-                    )
+            label = "{molec} [{branch}{jl:.0f}]({vl:.0f})->({vu:.0f})".format(
```
erwanp, line 274: Keep this for later; but it's very important to keep the isotope in the Line-Survey. It's one of the things people want to know : which isotopes belong to this line ? 

Now, I know that the isotope may have been removed from the Line Database exported in the spectrum ; but in that case it should be in .conditions() 

We could do something like (pseudo-code below:): 
```python
try:
    iso = row["iso"] in row
except KeyError:
    if is_float(spectrum.conditions["isotope"]):  # make sure only one ? 
          iso  = ... 
    else:
          iso = '?'
dict_of_other_labels.update({"iso":iso}
```

radis/lbl/base.py
```diff
@@ -665,17 +665,29 @@ def get_Evib_CDSD_pcN_1iso(df, iso):
             # the map below is crazy fast compared to above loop
             df["Evibu"] = df_pcN.index.map(Evib_dict.get).values
 
-            return df.loc[idx, ["Evibl", "Evibu"]]
+            try:
+                return df.loc[idx, ["Evibl", "Evibu"]]
+            except:
+                return df.loc[df.index.tolist(), ["Evibl", "Evibu"]]
```
erwanp, line 671: @gagan-aryan  why did you need this change ?

radis/lbl/factory.py
```diff
@@ -799,6 +804,9 @@ def eq_spectrum(
         if verbose:
             print("Spectrum calculated in {0:.2f}s".format(t))
 
+        if (id != "") and (s.lines is not None):
+            s.lines.attrs["id"] = id
+
```
erwanp, line 809: This is not needed, and pollutes the code here and above. 
I think you needed it to fix the line survey examples ?

Line survey is a secondary feature so we shouldn't sacrifice code clarity for it. But I think you can  : 
- get the "id" in the line survey by using the Spectrum conditions.
- or (better) : fill the attributes in `lines` by updating the get_lines() function  (L740)

radis/tools/line_survey.py
erwanp: You can also use the attributes the Lines; see my comment above and suggestion to modifiy the Factory "get_lines" function.

Question : will the attributes be correctly serialized, i.e., preserved if you Store a spectrum on disk and reload ? 
```
s.store("my.spec")
s = radis.load_spec("my.spec")
s.lines.attrs["id"]      # does it work ?
```

Also, what if we merge two single-isotope spectra ? If lines are appended, an isotope column should be added. 
Keep this in a log note in your Issue ("things to check later"). How Line Survey and Lines are handled will change eventually : we'll store in the Spectrum a link to the database used rather than all of the lines. Less memory use!

radis/lbl/base.py
gagan-aryan: I had to change this because I created an if-else block in L677 - L690. So idx was no longer accessible when `get_Evib_CDSD_pcN_1iso` was called from the else block.

radis/tools/line_survey.py
erwanp: (for reference, link to our Slack discussion @gagan-aryan   https://radis-radiation.slack.com/archives/D01PB4L6GV6/p1624543791001300?thread_ts=1624483169.008800&cid=D01PB4L6GV6 )

radis/tools/line_survey.py
```diff
@@ -267,61 +267,80 @@ def get_label_hitran(row, details, attrs):
...
+            molecule = get_molecule(id)
 
         # Get global labels
         if molecule in HITRAN_CLASS1:
+
+            add = ["vu", "vl", "jl", "iso"]
+
+            if "isotope" in spec.conditions:
+                iso = spec.conditions["isotope"]
+                add.append("iso")
```
erwanp, line 280: warning, "iso" is already in add here ?

erwanp (approved):

radis/tools/line_survey.py
gagan-aryan: Ya right !!

gagan-aryan: I have a few test in `test_calc.py` that seem to be passing on travis but are failing locally (after the merge from develop ). Any idea of what might be going wrong?

erwanp: Some tests are excluded on Travis ; mostly those requiring large databases. Default Travis config is defined [here](https://github.com/radis/radis/blob/develop/setup.cfg#L12).  You may get these defaults if you run pytest from the level root; but not if you're in a folder.

erwanp:
@gagan-aryan i think we're getting ready to merge ! 
Can you summarize in the PR description the main take-aways; link to the proof-of-concepts if there are, and add some performance results ? I think most of these informatinos are either on the Slack or in the comments above, so it's just about summarizing on the PR

gagan-aryan:
Done.

radis/lbl/base.py
erwanp: Following up : this `tolist()` is unusual to me; but I now understand why you had to add it. 

It should be a simpler syntax : 
```
return df.loc[:, ["Evibl", "Evibu"]]
```
on the model of `get_Evib123_RADIS_cls5_1iso` (L1149). `idx` here is undefined ; it seems like an error, which didn't have consequence because the "idx" would be picked from the parent function, and correspond to the same lines. 

The syntax above will fix everything.

radis/lbl/base.py
```diff
@@ -660,17 +660,29 @@ def get_Evib_CDSD_pcN_1iso(df, iso):
...
-            if radis.DEBUG_MODE:
-                assert (df.loc[idx, "iso"] == iso).all()
+                if radis.DEBUG_MODE:
+                    assert (df.loc[idx, "iso"] == iso).all()
+
+        else:
+            iso = df.attrs["iso"]
+            df.loc[df.index.tolist(), ["Evibl", "Evibu"]] = get_Evib_CDSD_pcN_1iso(
+                df.loc[df.index.tolist()], iso
+            )
```
erwanp, line 685: same, here you can probably try : 
```suggestion
            df.loc[:, ["Evibl", "Evibu"]] = get_Evib_CDSD_pcN_1iso(df, iso)
```

radis/lbl/base.py
```diff
@@ -781,11 +793,20 @@ def get_Evib_CDSD_pcJN_1iso(df, iso):
...
-            if radis.DEBUG_MODE:
-                assert (df.loc[idx, "iso"] == iso).all()
+                if radis.DEBUG_MODE:
+                    assert (df.loc[idx, "iso"] == iso).all()
+
+        else:
+            iso = df.attrs["iso"]
+            df.loc[df.index.tolist(), ["Evibl", "Evibu"]] = get_Evib_CDSD_pcJN_1iso(
+                df.loc[df.index.tolist()], iso
+            )
```
erwanp, line 809: same

radis/lbl/base.py
```diff
@@ -1024,8 +1045,17 @@ def get_Evib_RADIS_cls1_1iso(df, iso):
...
+            for iso, idx in df.groupby("iso").indices.items():
+                df.loc[idx, ["Evibl", "Evibu"]] = get_Evib_RADIS_cls1_1iso(
+                    df.loc[idx], iso
+                )
+
+        else:
+            iso = df.attrs["iso"]
+            df.loc[df.index.tolist(), ["Evibl", "Evibu"]] = get_Evib_RADIS_cls1_1iso(
+                df.loc[df.index.tolist()], iso
+            )
```
erwanp, line 1058: same

radis/lbl/base.py
erwanp: There is the same problem L555. You cna fix it

radis/lbl/base.py
```diff
@@ -1596,20 +1714,45 @@ def calc_weighted_trans_moment(self):
...
+                            assert (df.loc[idx, "id"] == id).all()
+                        assert (df.loc[idx, "iso"] == iso).all()
+
+            else:
                 state = self.input.state
                 parsum = self.get_partition_function_calculator(
-                    molecule, iso, state
+                    molecule, isotope_set[0], state
                 )  # partition function
                 df.at[idx, "Qref"] = parsum.at(Tref, update_populations=False)
```
erwanp, line 1747: is `idx` defined here?  (we're not in the groupby() case. Why didn't it trigger a syntax error ? )

radis/lbl/base.py
```diff
@@ -1873,7 +2014,11 @@ def _calc_Q(molecule, iso, state):
         # Einstein A coefficient and the populations (see Klarenaar 2017 Eqn. 12)
 
         # correct for Partition Function
-        line_strength = df1.int * (df1.Qref / df1.Qgas)
+        if "iso" in df1:
```
erwanp, line 2017: for all these "if 'iso' in df1" lines , i think a comment "#multiple-isotopes in database" would be nice for future develoeprs !

radis/lbl/base.py
```diff
@@ -1929,28 +2074,85 @@ def calc_populations_eq(self, Tgas):
...
-        # ... now fill the rest:
-        for (id, iso), idx in dgb.indices.items():
-            if (id, iso) == (id1, iso1):
-                continue
+            # ... optimize by filling all with first isotope first
+            id1, iso1 = list(dgb.indices.keys())[0]
+            molecule1 = get_molecule(id1)
+            state = self.input.state
+            parsum1 = self.get_partition_function_interpolator(molecule1, iso1, state)
+            Qgas[iso1] = parsum1.at(Tgas)
```
erwanp, line 2088: this "trick" was to fill the Qgas column with a first value. No, there is no more Qgas column. We can remove it

radis/lbl/base.py
```diff
@@ -1929,28 +2074,85 @@ def calc_populations_eq(self, Tgas):
...
+            # Load partition function values
+            dgb = df1.groupby(by=["id"])
+            iso = df1.attrs["iso"]
+
+            # ... optimize by filling all with first isotope first
+            id1 = list(dgb.indices.keys())[0]
+            molecule1 = get_molecule(id1)
+            state = self.input.state
+            parsum1 = self.get_partition_function_interpolator(molecule1, iso, state)
+            Qgas[iso] = parsum1.at(Tgas)
```
erwanp, line 2133: same

radis/lbl/base.py
```diff
@@ -2318,14 +2584,93 @@ def _calc_populations_noneq_multiTvib(
...
+                if radis.DEBUG_MODE:
+                    assert (df.loc[idx, "id"] == id).all()
+                    assert (df.loc[idx, "iso"] == iso).all()
 
+        elif "id" not in df and "iso" in df:
+            dgb = df.groupby(by=["iso"])
+            id = df.attrs["id"]
             molecule = get_molecule(id)
+
+            # TODO: optimize with np.take trick (see equilibrium case)
```
erwanp, line 2619: can be removed, your method is better : no more column :)

radis/lbl/base.py
```diff
@@ -2318,14 +2584,93 @@ def _calc_populations_noneq_multiTvib(
...
+                #            dg_list.append(dg)
+
+                if radis.DEBUG_MODE:
+                    assert (df.loc[idx, "iso"] == iso).all()
+
+        elif "id" in df and "iso" not in df:
+            dgb = df.groupby(by=["id"])
+            iso = df.attrs["iso"]
+
+            # TODO: optimize with np.take trick (see equilibrium case)
```
erwanp, line 2645: same

radis/lbl/base.py
```diff
@@ -2318,14 +2584,93 @@ def _calc_populations_noneq_multiTvib(
...
+                    Tvib,
+                    Trot,
+                    vib_distribution=vib_distribution,
+                    rot_distribution=rot_distribution,
+                    overpopulation=overpopulation,
+                    update_populations=self.misc.export_populations,
+                )
+
+                #            df.loc[idx, 'Qvib'] = Qvib
+                df.loc[idx, "Q"] = Q
```
erwanp, line 2662: warning; here and above (L2635) we're re-adding a "Q" column. That's not needed anymore, and not something we want !

radis/lbl/base.py
```diff
@@ -2337,13 +2682,9 @@ def _calc_populations_noneq_multiTvib(
             )
 
             #            df.loc[idx, 'Qvib'] = Qvib
-            df.loc[idx, "Q"] = Q
+            df.loc[df.index.tolist(), "Q"] = Q
```
erwanp, line 2685: same; we shouldn't had "Q" anymore. 

[and if we had to, using `df.loc[:, "Q"]` would be better

radis/lbl/loader.py
```diff
@@ -1821,6 +1831,26 @@ def load_and_concat(files):
...
+        df.drop("id", axis=1, inplace=True)
+        df_metadata.append("id")
+        df.attrs["id"] = id_set[0]
+
+        isotope_set = df.iso.unique()
+
+        if len(isotope_set) == 1:
+            df.drop("iso", axis=1, inplace=True)
+            df_metadata.append("iso")
+            df.attrs["iso"] = isotope_set[0]
```
erwanp, line 1852: Are `df_metadata` and `df.attrs` two different storage ? Let's keep one only (maybe future PR , this one is huge already!).  If you agree just open a quick Refactor issue from this comment.

erwanp (changes requested):

radis/lbl/base.py
gagan-aryan: The entire else block which contained this line was actually redundant. All the cases already seemed to be handled prior to this. So the lines that followed were a miss. So, it was not giving any `Variable used before declaration error`. 

As for the fix in L555. `get_Evib_CDSD_pc_1iso` is called in L562 and we are already passing idx in there. So we do not require this a `try-except` here.

radis/lbl/loader.py
gagan-aryan: Isn't df_metadata just a manually created dictionary in `loader.py`? I appended `iso` to this dictionary as well to make it consistent with the few other parts of the code where df_metadata is used (it is only used at one place while transferring metadata). I think we should drop this dictionary and just stick with `df.attrs` instead. 

I am creating a refactor issue.

radis/lbl/base.py
```diff
@@ -660,17 +660,29 @@ def get_Evib_CDSD_pcN_1iso(df, iso):
             # the map below is crazy fast compared to above loop
             df["Evibu"] = df_pcN.index.map(Evib_dict.get).values
 
-            return df.loc[idx, ["Evibl", "Evibu"]]
+            try:
+                return df.loc[idx, ["Evibl", "Evibu"]]
```
erwanp, line 664: Please have a look and carefully confirm, but I think idx is not defined in this function. It should be ":"  always, no need for try/except . 

Warning with try/except in general. It's OK at the user interface level but it's dangerous when changing  the internal physics part of the code like here

radis/lbl/base.py
```diff
@@ -1929,28 +2055,67 @@ def calc_populations_eq(self, Tgas):
...
 
-        # ... now fill the rest:
-        for (id, iso), idx in dgb.indices.items():
-            if (id, iso) == (id1, iso1):
-                continue
+        elif "id" not in df1 and "iso" in df1:
+            # Load partition function values
+            dgb = df1.groupby(by=["iso"])
+
+            # ... optimize by filling all with first isotope first
```
erwanp, line 2079: This optimization doesn't seem to work anymore with the new implementation where no Qgas is created. Use your proof of concept code to test it, but I think it should be fully removed

radis/lbl/base.py
```diff
@@ -2142,50 +2322,101 @@ def calc_populations_noneq(
...
+                        rot_distribution=rot_distribution,
+                        overpopulation=overpopulation,
+                        returnQvibQrot=True,
+                        update_populations=self.misc.export_populations,
+                    )
+
+                    # ... make sure PartitionFunction above is calculated with the same
+                    # ... temperatures, rovibrational distributions and overpopulations
+                    # ... as the populations of active levels (somewhere below)
+                    df.at[idx, "Qvib"] = Qvib
```
erwanp, line 2398: From here it starts to be a bit complicated, as you'll have to handle multiple partition functions.

You may need it to write another proof of concept to fully understand this case. 
But we may be able to merge the improvements for the equilibrium case already.

radis/lbl/base.py
```diff
@@ -2617,16 +2951,30 @@ def _calc_linestrength_noneq(self):
...
             # partition function
-            dgb = df.groupby(by=["id", "iso"])
-            for (id, iso), idx in dgb.indices.items():
+            if "id" in df:
+                dgb = df.groupby(by=["id", "iso"])
+                for (id, iso), idx in dgb.indices.items():
+                    molecule = get_molecule(id)
+                    state = self.input.state
+                    parsum = self._get_parsum(molecule, iso, state)
+                    df.at[idx, "Qref"] = parsum.at(Tref, update_populations=False)
```
erwanp, line 2960: The Qref column should definitly be replaced by a map()

radis/lbl/base.py
gagan-aryan: I Will be setting up a proof of concept as discussed [here.](https://radis-radiation.slack.com/archives/D01PB4L6GV6/p1624953400003900) Marking this comment as resolved here for ease of work on the PR.

radis/lbl/base.py
gagan-aryan: Yes, you are right. I remember adding this because some tests fail; maybe I added it hereafter, adding it at some other place. Not sure. Will be careful with try/except while handling internal physics code henceforth.

radis/lbl/base.py
gagan-aryan: Tested. Removed.

radis/lbl/base.py
gagan-aryan: Done.

erwanp: Super efficient, congrats ! 👏

radis/lbl/base.py
```diff
@@ -2640,7 +2973,14 @@ def _calc_linestrength_noneq(self):
         line_strength = df.int.copy()
 
         # ... correct for lower state population
-        line_strength /= df.gl * exp(-hc_k * df.El / Tref) / df.Qref
+        if "iso" in df:
+            print(Qref_dict)
```
erwanp, line 2977:
```suggestion
```

radis/lbl/loader.py
erwanp: (for reference : https://github.com/radis/radis/issues/300 )

radis/lbl/base.py
```diff
@@ -1816,9 +1935,16 @@ def _calc_Q(molecule, iso, state):
                     )
             return Qref, Qgas
 
-        id_set = df1.id.unique()
-        if len(id_set) == 1:
-            id = list(id_set)[0]
+        if "id" in df1:
```
erwanp, line 1938:
```suggestion
        if "id" in df1.columns:
```

erwanp (changes requested):

radis/lbl/base.py
```diff
@@ -2400,22 +2686,46 @@ def _calc_populations_noneq_multiTvib(
...
+                    / Q_dict[0]
+                )
+                df["nl"] = (
+                    nl_vib1Qvib1
+                    * nl_vib2Qvib2
+                    * nl_vib3Qvib3
+                    * df.grotl
+                    * exp(-df.Erotl * hc_k / Trot)
+                    / Q_dict[0]
+                )
```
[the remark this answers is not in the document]
gagan-aryan: So, I though it would be difficult to generalise the if/else blocks for both eq and non_eq so I have split the handling of Q to two functions. One of those functions is identical to the psuedo code as you suggested. I have used this function in the non_eq handling as well. If you approve of these changes, I will add docstrings and comments.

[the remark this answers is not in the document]
erwanp: It looks very good, and having one for eq and one for noneq makes a lot of sense.

erwanp (approved):

radis/lbl/base.py
```diff
@@ -2073,6 +2074,7 @@ def calc_populations_eq(self, Tgas):
 
         # multiple-isotopes in database
 
+        # extract Qgas in the required form from the dictionary
         Q_req = self.getQ_from_dict(df1, Qgas)
```
erwanp, line 2078: for clarity, maybe rename Q_req as Qgas  and Qgas as Qgas_dict ?

erwanp (approved):
Looks good ! I think we can merge tomorrow !!

dcmvdbekerom:
Looks very promising, I'll try to have a look this evening!

radis/lbl/factory.py
```diff
@@ -911,10 +911,14 @@ def eq_spectrum_gpu(
...
 
         try:
-            id_set = self.df0[
-                "id"
-            ].unique()  # get all the molecules in the dataframe, should ideally be 1 element for GPU
-            mol_id = id_set[0]
+            if "id" in self.df0:
+                id_set = self.df0[
+                    "id"
+                ].unique()  # get all the molecules in the dataframe, should ideally be 1 element for GPU
```
dcmvdbekerom, line 917: >should ideally be 1 element for GPU
I think GPU can handle multiple species without too much problems, as long as we combine the mol_id and iso in a single identification array. For all intents and purposes, different isotopes might as well be diffent molecules.

radis/lbl/factory.py
```diff
@@ -911,10 +911,14 @@ def eq_spectrum_gpu(
...
+                mol_id = id_set[0]
```
dcmvdbekerom, line 918: What if there's more than one element in id_set?

gagan-aryan: The code before already seemed to assume one element. Just extended the same here.

dcmvdbekerom: I would put an assertion here to be safe, otherwise if someone unknowingly extends to multiple id's in the future there will be no exception thrown here.

dcmvdbekerom (approved):
Looking good, nice additions!!

radis/lbl/factory.py
erwanp: Mole fractions may be different : you still need a specific mol_id

erwanp (approved):

erwanp:
Hello @gagan-aryan . Realized there was a [Qref column created](https://github.com/radis/radis/blob/develop/radis/lbl/base.py#L1727) during the calculation of Weighted Transiitno Square Moments (used for nonequilibrium calculations). 

This would be an easy improvement using the usual map(dict) trick ! 

You can open a small issue from this comment an assign it to your Project. 

---

Side note ; I also checked that it's still worth using Qref as an attribute if single isotope : 

```
from radis.io.hitemp import fetch_hitemp
df = fetch_hitemp("CO",  isotope="1")
Qdict={1:300}
%timeit df.int/Qdict[1]
# 1.32 ms ± 286 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
%timeit df.int/df.iso.map(Qdict)
# 5.71 ms ± 1.82 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)
# so for one entry only; .map() is 4 times slower. 
# even accounting the time to "check" is there is only one isotope : 
%timeit df.iso.unique()
1.8 ms ± 512 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

radis/lbl/base.py
```diff
@@ -1968,14 +2071,116 @@ def calc_populations_eq(self, Tgas):
...
+                if radis.DEBUG_MODE:
+                    assert (df.loc[idx, "id"] == id).all()
+                    assert (df.loc[idx, "iso"] == iso).all()
+
+        elif "id" not in df and "iso" in df:
+
+            dgb = df.groupby(by=["id"])
+
+            id = df.attrs["id"]
+            iso1 = list(dgb.indices.keys())[0]
```
erwanp, line 2152: @gagan-aryan this looks like an error : you're using the same isotope value (iso1) for all partition functions ?

PR #291, opened 2021-06-18
Add ruler tool in plots

erwanp:
### Description

- Add a Cursor in Spectrum.plot()   (as was in plot_diff already)
- (experimental) Add a Ruler tool in Spectrum.plot() and plot_diff

image

Still some tiny problems with Ruler tool : 
- [ ] once initialized, it resets the zoom the first time. I don't know why.


erwanp:
@CorentinGrimaldi could easily be added to [Fitroom](https://github.com/radis/fitroom) too  (it's just an `add_ruler()` function) !

minouHub:
Hi Erwan,
Pulled your branch. It works. However, I cannot zoom and **then** use the ruler.

Also got a lot of (intended?) warnings:
```python
<matplotlib.lines.Line2D at 0x21bd051ccd0>e:\python\radis\radis\tools\plot_tools.py:199: MatplotlibDeprecationWarning: Setting the line's pick radius via set_picker is deprecated since 3.3 and will be removed two minor releases later; use set_pickradius instead.
  (self._marker_a,) = self.ax.plot((x0, y0), **used_markerprops)
e:\python\radis\radis\tools\plot_tools.py:200: MatplotlibDeprecationWarning: Setting the line's pick radius via set_picker is deprecated since 3.3 and will be removed two minor releases later; use set_pickradius instead.
  (self._marker_b,) = self.ax.plot((x0, y0), **used_markerprops)
e:\python\radis\radis\tools\plot_tools.py:201: MatplotlibDeprecationWarning: Setting the line's pick radius via set_picker is deprecated since 3.3 and will be removed two minor releases later; use set_pickradius instead.
  (self._marker_c,) = self.ax.plot((x0, y0), **used_markerprops)
```

erwanp:
- The first warning is unfortunate. It is already suppressed in RADIS, but it only works with matplotlib>=3.4 which itself requires Python 3.7+.  (https://github.com/matplotlib/matplotlib/issues/15284)
I initially enforced matplotlib>=3.4 but many Radis users probably still have Python 3.5 or 3.6 and may not want to upgrade yet ; they better receive the warnings. 

- I think I fixed the other 2 warnings.

- When you say you cannot Zoom & user the ruler  : can't you unselect the Zoom and then use the ruler ? 

- I also noticed that the ruler is not enabled on the first plot ; can you confirm this ?

minouHub:
> I also noticed that the ruler is not enabled on the first plot ; can you confirm this 
Yes, that's quick funny. It works the second time.

**First run:**
```python
e:\python\radis\radis\tools\plot_tools.py:561: UserWarning: Couldn't add Ruler tool (still an experimental feature in RADIS : please report the error !)
  warn(
```
- [ ] no ruler

**Second run:**
```python
e:\python\radis\radis\tools\plot_tools.py:561: UserWarning: Couldn't add Ruler tool (still an experimental feature in RADIS : please report the error !)
  warn(
```

- [ ] many warnings ruler
- [ ] (minor) If I zoom or move the plot first, clicking on the ruler does not unselect the drag/zoom button. It took me 5 minutes to realize this is why I could not use the ruler after a zoom ^^

erwanp:
Ok so none of this looks production-ready. 
What about adding it, but as an experimetanl option in the Radis.json, False by default ?.

minouHub:
I think that's a very good idea to add a `False/True` somewhere. This way, I can use the ruler and it leaves the mainstream radis clean for other users. However, I expected this solution to be too complicated as it might mess up with the matplotlib version?

minouHub (approved):
Pulled!
Alright for me :)

PR #293, opened 2021-06-22
add: plot_linestrength example

gagan-aryan:
### Description

This pull request iadds the example to plot linestrengths as a bar plot in the waverange.

### DIFF
--- a/examples/plot_linestrengths.py
+++ b/examples/plot_linestrengths.py
@@ -1,0 +1,79 @@
+import numpy as np
+from matplotlib import pyplot as plt
+from numpy import exp
+
+from radis.db.classes import get_molecule, get_molecule_identifier
+from radis.io.hitemp import fetch_hitemp
+from radis.levels.partfunc import PartFuncHAPI
+from radis.phys.constants import hc_k
+
+
+def get_Qgas(molecule, iso, T):
+
+    M = get_molecule_identifier(molecule)
+
+    Q = PartFuncHAPI(M, iso)
+    return Q.at(T=T)
+
+
+def calc_linestrength_eq(df, Tref, Tgas):
+
+    print("Scaling equilibrium linestrength")
+
+    # %% Load partition function values
+
+    def _calc_Q(molecule, iso, T_ref, T_gas):
+
+        Qref = get_Qgas(molecule, iso, T_ref)
+        Qgas = get_Qgas(molecule, iso, T_gas)
+
+        return Qref, Qgas
+
+    id_set = df.id.unique()
+    id = list(id_set)[0]
+    molecule = get_molecule(id)  # retrieve the molecule
+    iso_set = set(df.iso)  # df1.iso.unique()
+
+    iso_arr = list(range(max(iso_set) + 1))
+
+    Qref_arr = np.empty_like(iso_arr, dtype=np.float64)
+    Qgas_arr = np.empty_like(iso_arr, dtype=np.float64)
+    for iso in iso_arr:
+        if iso in iso_set:
+            Qref, Qgas = _calc_Q(molecule, iso, Tref, Tgas)
+            Qref_arr[iso] = Qref
+            Qgas_arr[iso] = Qgas
+
+    df["Qref"] = Qref_arr.take(df.iso)
+    df["Qgas"] = Qgas_arr.take(df.iso)
+
+    # Scaling linestrength with the equations from Rotham's paper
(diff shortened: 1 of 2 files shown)


erwanp:
Super nice ! Does it correctly build the docs locally ?

One issue I can see about adding this in the main Examples : it uses CH4 ; which is very long to download and process. The examples will be re-computed for every change to the `develop` branch, so we want to limit data traffic. What about using CO for the online version ?

gagan-aryan:
@erwanp I have replaced it with CO and yes the docs pass locally.

examples/plot_linestrengths.py
```diff
@@ -0,0 +1,92 @@
...
+================================================
+Scale Linestrengths of carbon-monoxide
+================================================
+
+This example scales the linestrengths of CO to Tgas=300 from Tref = 296 and then plots
+the linestrengths against the wavenumbers. We are using :py:func:`~radis.io.hitemp.fetch_hitemp`
+function to retrieve the dataframe from the HITEMP-CO databank. The calculations done
+for scaling use the  equation that can be found in Rotham's paper (equation A11).
+
+"""
```
erwanp, line 12: We can use these examples to give information about what's done in the code (good for future spectroscopy tutorials !)

```suggestion
Similar functions are used directly at the hearth of RADIS's SpectrumFactory in the :py:meth:`~radis.lbl.BaseFactory.calc_linestrength_eq` and :py:meth:`~radis.lbl.BaseFactory.calc_linestrength_noneq` methods
"""
```

I also suggest that you : 
- borrow the equation and link to the references of BaseFactory.calc_linestrength_eq docstrings, and add it here. Will look very nice !
- mention this example in BaseFactory.calc_linestrength_eq docstring, by adding 
```
.. minigallery:: radis.lbl.base.BaseFactory.calc_linestrength_eq
```
- make BaseFactory._calc_linestrength_noneq a public method by renaming it BaseFactory.calc_linestrength_noneq

examples/plot_linestrengths.py
```diff
@@ -0,0 +1,92 @@
...
+    Tref = 296
+    df = fetch_hitemp(
+        molecule="CO",
+        databank_name="HITEMP",
+        isotope="1, 2, 3",
+        load_wavenum_min=2000,
+        load_wavenum_max=2250,
+    )
+
+    Tgas = 300
```
erwanp, line 86: Scaling from 296 to 300 K doesn't sound like a lot of changes. What about scaling up to, let's say, 450 K ?

erwanp (changes requested):

examples/plot_linestrengths.py
```diff
@@ -0,0 +1,92 @@
...
+    Qref_arr = np.empty_like(iso_arr, dtype=np.float64)
+    Qgas_arr = np.empty_like(iso_arr, dtype=np.float64)
+    for iso in iso_arr:
+        if iso in iso_set:
+            Qref, Qgas = _calc_Q(molecule, iso, Tref, Tgas)
+            Qref_arr[iso] = Qref
+            Qgas_arr[iso] = Qgas
+
+    df["Qref"] = Qref_arr.take(df.iso)
+    df["Qgas"] = Qgas_arr.take(df.iso)
```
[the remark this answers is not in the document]
gagan-aryan: So we will be able to merge this once we the new approach in merged to the main code ?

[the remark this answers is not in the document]
erwanp: Yes. We can even add your example before the main code is upgraded;  but the example should use the map(dict) approach not to have to modify it afterwards (also, it's faster as you've shown !)

examples/plot_linestrengths.py
```diff
@@ -49,19 +57,15 @@ def _calc_Q(molecule, iso, T_ref, T_gas):
 
     iso_arr = list(range(max(iso_set) + 1))
 
-    Qref_arr = np.empty_like(iso_arr, dtype=np.float64)
-    Qgas_arr = np.empty_like(iso_arr, dtype=np.float64)
+    Qref_Qgas_ratio = {}
+
     for iso in iso_arr:
         if iso in iso_set:
```
erwanp, line 63: @gagan-aryan one more comment : why do you need to create iso_arr ? Can't we just do `for iso in iso_set` ?

erwanp (approved):

erwanp (approved):

PR #294, opened 2021-06-24
Quick Fix for zero_padding problems + minor changes

CorentinGrimaldi:
### Description

Quick Fix for  zero_padding problems while trying to generate a database   _(Edit @erwanp  : pre-computed spectra are systematically re-computed instead of being loaded)_, becausee zero_padding _(input : -1 )_ changes after evaluation _(and the evaluated value is stored in the Spectrum conditions)_ . Make it a misc parameter.

Minor changes in docstring and error messages

### DIFF
--- a/radis/lbl/broadening.py
+++ b/radis/lbl/broadening.py
@@ -1449,7 +1449,7 @@ class BroadenFactory(BaseFactory):
             wstep = self.params.wstep
             w_lineshape_ft = np.fft.rfftfreq(
                 2 * len(w), wstep
-            )  # TO-DO: add  + self.params.zero_padding
+            )  # TO-DO: add  + self.misc.zero_padding
 
             w_fold = (w_lineshape_ft, w_lineshape_ft[::-1])
 
@@ -1462,7 +1462,7 @@ class BroadenFactory(BaseFactory):
                         w_lineshape_ft, wG[l] / 2, wL[m] / 2
                     )
 
-                    ## Add folding until threshold is reached:
+                    # Add folding until threshold is reached:
                     n = 1
                     while (
                         voigt_FT(n / (2 * wstep), wG[l] / 2, wL[m] / 2)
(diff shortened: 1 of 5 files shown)


radis/lbl/broadening.py
```diff
@@ -2003,10 +2003,10 @@ def _broaden_lines(self, df):
         # Get which optimization method to use:
         optimization = self.params.optimization
 
-        if self.params.zero_padding < 0 or self.params.zero_padding > len(
+        if self.misc.zero_padding < 0 or self.misc.zero_padding > len(
```
anandxkumar, line 2006: @erwanp, @CorentinGrimaldi, how exactly does changing `self.params.zero_padding` to  `self.misc.zero_padding` solves the issue? Won't it have the same meaning? This PR is based on radis before implementation of `wstep = auto`. So it will be initialized during Spectrum Factory only and won't change since `_generate_wavenumber_range` will be executed once only. But considering the addition of `wstep = auto` now whenever we compute a Spectrum using `SF.eq_spectrum()`, the `len(wavenumber_calc)` will change based on changing the parameters like wstep. So it will change if `self.misc.zero_padding > new len(wavenumber_calc)`. If we want to keep `zero_padding = -1` for every new evaluation we can introduce a new parameter `self._zero_padding = -1` in `Spectrum Factory.__init__` and update the condition like below:
```
if self.misc._zero_padding < 0 or self.misc._zero_padding > len(
            self.wavenumber_calc
        ):
            self.misc.zero_padding = len(self.wavenumber_calc)
```
Will it solve the problem? Correct me if I said anything wrong here!

erwanp: The code that evaluates if spectra have already been computed or not in the SpecDatabase  just looks at the parameters in "self.params" or "self.input",  and discards the parameters in "self.misc"

erwanp (approved):

anandxkumar (approved):

PR #298, opened 2021-06-26
do not load energies if equilibrium calculation in calc_spectrum

### LINKED ISSUE #297: H2O equilibrium calc with direct HITEMP file  raises spectroscopic constants not found
### 🐛 Describe the bug

Reported by [周周zz](https://twitter.com/zz99583643/status/1408739988439666689)  on Twitter

### 🖥 Steps To Reproduce

```python

s = calc_spectrum(wavenum_min = 7500 / u.cm,
                  wavenum_max = 7600 / u.cm,
                  molecule='H2O',
                  isotope='1,2,3',
                  pressure=1.01325 * u.bar,
                  mole_fraction=0.58,
                  path_length=6.8 * u.cm,
                  verbose=3,
                  Tgas=3000,   
                  databank=r'\PATH_TO_\01_7500-8000_HITEMP2010.par',
                  )
s.plot('radiance_noslit', nfig='same')
```

Raises : 

```
KeyError: "H2O is not defined in molecules with built-in spectroscopic constants. Choose one of: ['CO', 'CO2']"
```

Which is needed only for nonequilibrum calculation (which is not required here)

erwanp:
### Description

Fixes #297  ; directly from calc_spectrum 
(was already fixed by https://github.com/radis/radis/pull/252 at the SpectrumFactory level by delaying the loading of rovibrational energies until a noneq calculation was triggered ; here we use the extra information we have with calc_spectrum (is it an eq or non eq calculation) to directly load rovibrational energies from the start.

### DIFF
--- a/radis/lbl/calc.py
+++ b/radis/lbl/calc.py
@@ -588,7 +588,7 @@ def _calc_spectrum(
             # constants (not all molecules are supported!)
             conditions["levelsfmt"] = "radis"
             conditions["lvl_use_cached"] = use_cached
-            conditions["load_energies"] = True
+        conditions["load_energies"] = not _equilibrium
         # Details to identify lines
         conditions["parse_local_global_quanta"] = (not _equilibrium) or export_lines
         sf.fetch_databank(**conditions)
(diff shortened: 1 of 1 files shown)


anandxkumar (approved):

anandxkumar:
Small but nice change!

erwanp:
Thanks for the careful review! 

Noneq CO2: we cannot do more than what's written. The database is missing some lines, and Radis shouldn't try to fix it alone without the user knowing. So the suggestion in the error message is the best we can do

PR #299, opened 2021-06-27
Quick Fix: Statements printing on verbose = 0

anandxkumar:

Fixes statements printing on verbose = False [link](https://github.com/radis/radis/pull/286#issuecomment-869208816)

### DIFF
--- a/radis/lbl/broadening.py
+++ b/radis/lbl/broadening.py
@@ -60,7 +60,6 @@ Formula in docstrings generated with :py:func:`~pytexit.pytexit.py2tex` ::
 ----------
 
 """
-from time import time
 from warnings import warn
 
 import matplotlib.pyplot as plt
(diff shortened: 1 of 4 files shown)


radis/lbl/broadening.py
```diff
@@ -2390,7 +2396,9 @@ def calculate_pseudo_continuum(self, noneq=False):
 
         if self.params.pseudo_continuum_threshold > 0:
 
-            self.profiler._print(2, "Calculating pseudo continuum")
+            if self.verbose >= 2:
+                print("... Calculating pseudo continuum")
+
```
erwanp, line 2401: Hello. Can you add an optional "start" message to `self.profiler.start` instead ?  It would remove all these lines and make the source code cleaner. Also, can you remove t0 = time() as it should be embedded in the Profiler too ?

anandxkumar: Did you mean `self.profiler.start("key",vebose_level, optional="")`. We can use it to print the statements prior to `start()` but then we need to make sure the verbose level matches of start and the prior statement and if it matches then we can simply print it using `("... Calculating pseudo continuum")` or `"..." * (verbose_level -1) + optional` which will be similar to `_print` method. So should not use the previous proposed [implementation](https://github.com/radis/radis/pull/286#issuecomment-868400588), will cover a larger area? Or the above discussed `optional` method?

minouHub (approved):
The verbose = 0 acts as expected. Don't forget Erwan's comments

radis/tools/database.py
```diff
@@ -2372,7 +2372,11 @@ def _load_new_file(self, file, binary=False):
 
         out = s.get_conditions().copy()
 
-        # Add filename, and a link to the Spectrum object itself
+        # Add filename, name and a link to the Spectrum object itself
+        if s.name == None:
+            out["name"] = s.get_name()
```
anandxkumar, line 2377: @erwanp when `s.name == None`, the csv actually stores it as the Filename (eg -`20210629_Tvib3000K_Trot3000K.spec`
, I wonder why?

anandxkumar, line 2377: @erwanp, I'm actually using this to store the name column in the csv file instead of updating the `conditions`. If `s.name==None` it stores the filename else the name given to it. It gets the job done. Any better location for doing this?

erwanp: Good idea! Let's keep it here!

erwanp (approved):

PR #302, opened 2021-07-02
perf: dataframe refactors : drop molar mass and abundance columns

gagan-aryan:
### Description

This pull request is to implement the following df-refactors to reduce memory and boost pandas' performance. 

- [x] drop molar_mass column
- [x] drop abundance column

### DIFF
--- a/radis/lbl/broadening.py
+++ b/radis/lbl/broadening.py
@@ -69,6 +69,8 @@ from numpy import arange, exp
 from numpy import log as ln
 from numpy import pi, sin, sqrt, trapz, zeros_like
 
+from radis.db.classes import get_molecule_identifier
+from radis.db.molparam import MolParams
 from radis.lbl.base import BaseFactory
 from radis.misc.arrays import add_at, numpy_add_at
 from radis.misc.basics import is_float
@@ -800,6 +802,21 @@ class BroadenFactory(BaseFactory):
         # Init variables
         df = self.df1
 
+        try:
+            arr = list(self.df0.attrs.keys())
+
+            for k in arr:
+                df.attrs[k] = self.df0.attrs[k]
+
+        except:
+            isotope_set = list(self.input.isotope)
+            if len(isotope_set) == 1:
+                df.attrs["iso"] = int(isotope_set[0])
+            else:
+                df.attrs["iso"] = isotope_set
+            M = get_molecule_identifier(self.input.molecule)
+            df.attrs["id"] = M
+
         if len(df) == 0:
             return  # no lines
 
@@ -945,6 +962,8 @@ class BroadenFactory(BaseFactory):
         else:
             Tdpsel = df.Tdpsel
 
+        molar_mass = get_molar_mass(df)
+
         # Calculate broadening FWHM
         wv, wl, wg = voigt_broadening_HWHM(
             df.airbrd,
@@ -952,7 +971,7 @@ class BroadenFactory(BaseFactory):
             df.Tdpair,
             Tdpsel,
             df.wav,
-            df.molar_mass,
+            molar_mass,
             pressure_atm,
             mole_fraction,
             Tgas,
@@ -1030,8 +1049,13 @@ class BroadenFactory(BaseFactory):
             - ``hwhm_gauss``
         """
 
+        print(df.head())
+        print(df.attrs)
+        molar_mass = get_molar_mass(df)
+
         # Calculate broadening HWHM
-        wg = doppler_broadening_HWHM(df.wav, df.molar_mass, Tgas)
+        print("it is here")
+        wg = doppler_broadening_HWHM(df.wav, molar_mass, Tgas)
         # Note @EP: should we use the pressure-shifted wavenumber instead of df.wav?
 
         # Update dataframe
@@ -2819,6 +2843,33 @@ def project_lines_on_grid_noneq(df, wavenumber, wstep):
         Ei_density_on_grid,
         line2grid_projection_left,
     )
+
+
+def get_molar_mass(df):
+
+    molpar = MolParams()
+
+    if "id" in df.columns:
+        id = df.id.unique()
+        id = id[0]
+
+    else:
+        id = df.attrs["id"]
+
+    if "iso" in df.columns:
+        iso_set = df.iso.unique()
+        molar_mass = {}
+        for iso in iso_set:
+            molar_mass[iso] = molpar.get(id, iso, "mol_mass")
+
+        req = df["iso"].map(molar_mass)
+    else:
+        iso = df.attrs["iso"]
+        molar_mass = molpar.get(id, iso, "mol_mass")
+
+        req = [molar_mass] * len(df.index)
+
+    return req
 
     #
 
--- a/radis/lbl/broadening.py
+++ b/radis/lbl/broadening.py
@@ -2837,6 +2837,19 @@ def project_lines_on_grid_noneq(df, wavenumber, wstep):
 
 def get_molar_mass(df):
 
+    """Returns molar mass
+
+    Parameters
+    ----------
+
+    df: dataframe
+
+    Returns
+    -------
+
+    The molar mass of all the isotopes in the dataframe
+    """
+
     molpar = MolParams()
 
     if "id" in df.columns:
--- a/radis/lbl/broadening.py
+++ b/radis/lbl/broadening.py
@@ -802,20 +802,13 @@ class BroadenFactory(BaseFactory):
         # Init variables
         df = self.df1
 
-        try:
-            arr = list(self.df0.attrs.keys())
-
-            for k in arr:
-                df.attrs[k] = self.df0.attrs[k]
-
-        except:
-            isotope_set = list(self.input.isotope)
-            if len(isotope_set) == 1:
-                df.attrs["iso"] = int(isotope_set[0])
-            else:
-                df.attrs["iso"] = isotope_set
-            M = get_molecule_identifier(self.input.molecule)
-            df.attrs["id"] = M
+        isotope_set = list(self.input.isotope)
+        if len(isotope_set) == 1:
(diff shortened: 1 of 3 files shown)


gagan-aryan:
@erwanp one test from the line_survey module currently fails. If you feel it is valuable information to be displayed in the survey, I will write the code to calculate einstein's coefficient in the line survey module. Otherwise, I will drop it from the details.

radis/lbl/broadening.py
```diff
@@ -800,6 +802,21 @@ def _calc_broadening_HWHM(self):
         # Init variables
         df = self.df1
 
+        try:
```
erwanp, line 805: This try /except is hard to understand. 

What's the error you're expecting? You should write it in except, like except AttributeError or KeyError. 
Add some comments too?

radis/lbl/broadening.py
```diff
@@ -1030,8 +1049,13 @@ def _add_doppler_broadening_HWHM(self, df, Tgas):
             - ``hwhm_gauss``
         """
 
+        print(df.head())
+        print(df.attrs)
+        molar_mass = get_molar_mass(df)
+
         # Calculate broadening HWHM
-        wg = doppler_broadening_HWHM(df.wav, df.molar_mass, Tgas)
+        print("it is here")
```
erwanp, line 1057: BTW, I insert a lot of `raise` or stop points in my own debugging, then use `%debug` from an IPython console to get access to local variables, and go up/down the parent and child functions with `u` and `d`

radis/lbl/broadening.py
```diff
@@ -2820,6 +2844,33 @@ def rough_sum_on_grid():
...
+        molar_mass = {}
+        for iso in iso_set:
+            molar_mass[iso] = molpar.get(id, iso, "mol_mass")
+
+        req = df["iso"].map(molar_mass)
+    else:
+        iso = df.attrs["iso"]
+        molar_mass = molpar.get(id, iso, "mol_mass")
+
+        req = [molar_mass] * len(df.index)
```
erwanp, line 2870: Are you sure you need to build a full list here? That's wasting memory. One float should be enough

radis/lbl/factory.py
```diff
@@ -1709,8 +1711,24 @@ def optically_thin_power(
...
+            # ... Make sure degeneracies are calculated
+            if not all_in(["gju", "gjl", "gvibu", "gvibl", "gu", "gl"], self.df1):
+                self._calc_degeneracies(self.df1)
+
+            self.calc_weighted_trans_moment(self.df1)
+            self.calc_einstein_coefficients(self.df1)
+
             self.calc_populations_eq(Tgas)
-            self.df1["Aul"] = self.df1.A  # update einstein coefficients
+            self.df1["Aul"] = self.df1.Aul  # update einstein coefficients
```
erwanp, line 1731: Not required, is it?

radis/lbl/loader.py
```diff
@@ -704,7 +704,7 @@ def fetch_databank(
         parfuncfmt="hapi",
         levels=None,
         levelsfmt="radis",
-        load_energies=False,
+        load_energies=True,
```
erwanp, line 707: This was @anandxkumar  change. Energy not loaded by default unldss we actually need a nonequilibrium calculation. Let's try to keep it like this

radis/lbl/factory.py
```diff
@@ -1554,10 +1554,12 @@ def _get_S0(self, Ia_arr):
         if "S0" in df.columns:
             return df["S0"]
 
+        self.calc_einstein_coefficients()
```
erwanp, line 1557: This may need some discussion on Slack. I'm worried this function is quite expensive to compute. (deal with other comments before this one :)  )

erwanp (changes requested):

radis/lbl/broadening.py
gagan-aryan: SpectrumFactory does not have a df0. So I have put a try/except for that case.

radis/lbl/factory.py
gagan-aryan: Yes, it isn't. 😅

radis/lbl/broadening.py
gagan-aryan: Thank you. Will try using this henceforth.

radis/lbl/loader.py
gagan-aryan: Since I am now calculating the trans_moment in factory.py I had to set the load_energies parameter. Any other way energies could be loaded at runtime?

radis/lbl/broadening.py
erwanp: It always has df0, unless database wasn't loaded (but then it should fail elsewhere). Can you provide an example?

gagan-aryan: Examples - [1](https://github.com/radis/radis/blob/7c1be93c94ee3f50f03ef362e36a46cb6068f57d/radis/test/io/test_hitemp.py#L80) [2](https://github.com/radis/radis/blob/7c1be93c94ee3f50f03ef362e36a46cb6068f57d/radis/test/io/test_hitemp.py#L97)

radis/lbl/loader.py
erwanp: If that's required it would fail for all species that are not ready for nonequilibrium (I.e, all but CO and CO2). Can you run some tests with OH for instance? 

As I say in another comment, calculation of the transition moment looks expensive, maybe it isnt the right strategy. 

Before pushing forward, maybe just focus on molar mass and abundance, which are easy to implement. Then let's discuss about Einstein coefficients / linestrengths. It may also depend on what's usually available in data search others than HITRAN / HITEMP. If Exomol or GEISA provide Einstein A rather than Linestengths it may be more interesting to keep A and drop the linestength. Should be discussed on Slack!

radis/lbl/factory.py
gagan-aryan: https://radis-radiation.slack.com/archives/C0224856BUM/p1625263911008400 Link to the discussion

radis/lbl/loader.py
gagan-aryan: Yes, sure !!

gagan-aryan: I will revert to my molar_mass commit, start working from there, and once abundance work is done we can get back to this.

erwanp: Don't revert everything ; just create a new branch (e.g: `Einstein` ) from here to keep the preliminary work on Einstein coefficients; and reset your `df-refactors` to the molar_mass commit

gagan-aryan:
Still facing issues with the calculations of Intensity. Pushed for either @dcmvdbekerom @erwanp to look into if time permits.

radis/lbl/broadening.py
```diff
@@ -565,7 +565,7 @@ def voigt_lineshape(w_centered, hwhm_lorentz, hwhm_voigt, jit=True):
     wv = 2 * hwhm_voigt  # HWHM > FWHM
 
     if jit:
-        lineshape = _whiting_jit(w_centered, wl, wv)
+        lineshape = _whiting_jit(w_centered, wl.astype("float64"), wv.astype("float64"))
```
erwanp, line 568: If you're doing this then you're loosing the memory gained earlier. Is this because @jit expecteds float64? Then we'd need to update the definition in @jit.

radis/io/hitran.py
```diff
@@ -951,9 +956,62 @@ def parse_global_quanta(df, mol):
             "Unknown class for molecule {0}. Cant parse global quanta".format(mol)
         )
 
+    df = reduce_mem_usage(df)
+
     return df
 
 
+def reduce_mem_usage(props):
```
erwanp, line 964: do you have any idea of how long it takes to run this function ?

Also ; even if the general-purpose function is great for testing, I think it's very dangerous to auto-convert quantities which may differ from one database to another.

I'd rather hardcode the column names.

dcmvdbekerom:
I'll have a look tonight!

radis/io/hitran.py
```diff
@@ -284,6 +285,8 @@ def hit2df(
     # but files that have partly relevant lines are fully loaded.
     # Note : cache file is generated with the full line list.
 
+    df = reduce_mem_usage(df)
```
erwanp, line 288: note : this doesn't change the format in the cache HDF5 stored on disk (already written above).

radis/lbl/broadening.py
gagan-aryan: Yes, there was a conflict in the datatypes of `w_centered(float64)`, `wl(float32)` and `wv(float32)`. I did this because of that.

radis/io/hitran.py
gagan-aryan: I will dig into this and create a fresh PR.

dcmvdbekerom:
Hi @gagan-aryan! What exactly was the issue with the the calculations of Intensity?

gagan-aryan:
Hey @dcmvdbekerom ! So after talking to Erwan about this [on slack](https://radis-radiation.slack.com/archives/C0224856BUM/p1625669411012900), I removed the commits that added the datatypes crunching from this PR.

radis/lbl/broadening.py
erwanp: Back on this one : we shouldn't have a try/except here. Maybe the attributes are not transfered when df1 is created? Can you look it up?

erwanp (changes requested):

radis/lbl/broadening.py
gagan-aryan: I believe the problem is arising because even we are no longer transferring the attributes that are stored as `attrs` from `df0` to `df1` anymore. I can alternatively, remove the try/except and instead retain the code present in the except block alone.

erwanp (approved):
Nice! Do you think it's ready for merging?

gagan-aryan:
@erwanp We should be able to merge it now 😄

Issue #179, opened 2021-01-10
Add plotly/bokeh backend for plots


erwanp:
### 🎯 Describe what you'd like

@suzil in [radis-app](https://github.com/suzil/radis-ui) uses Plotly to plot the spectra (see output here : https://radis.app/ ) 

The interactive experience is better on a webpage than the Matplotlib plots that we have, and could be built directly in the RADIS library. 

It won't replace the current Maptlotlib inline/QT backends for most research applications, but could be used as the default in the Notebooks of [RADIS-lab](https://radis.github.io/radis-lab).  

### 💡 Possible implementations

- to start with, build a backend Spectrum._plot_plotly() version 

- move the current Spectrum.plot in Spectrum._plot_matplotlib() ; add a parameter in [Spectrum.plot()](https://radis.readthedocs.io/en/latest/source/radis.spectrum.spectrum.html#radis.spectrum.spectrum.Spectrum.plot) or (better) in the user config file #167  ?  

- Ideally, we should detect what environnement we're in; and use Plotly in Jupyter notebooks / Jupyter labs  See for instance https://stackoverflow.com/questions/34091701/determine-if-were-in-an-ipython-notebook-session 

Note : we already use Plotly for the Line survey plots. #173  so no extra dependency is needed

Edit : Alternatively, Bokeh and Holoviews are nice librairies with similar features as Plotly ; Bokeh is being considered for the RADIS-Fitroom sub-project . 
Edit 2 : Plotly has more traction, let's use Plotly.

erwanp:
@gh4ag if you know !

erwanp:
Minimal code : 
(Plotly works much better when giving a DataFrame as input)

```
import pandas as pd
import plotly.express as px
px.line(pd.DataFrame({'Wavenumber':s.get('radiance_noslit')[0], 'Radiance_noslit':s.get('radiance_noslit')[1]}), x="Wavenumber", y="Radiance_noslit", template='plotly_dark')
```

Renders nice on [RADIS-Lab](https://radis.github.io/radis-lab/) with Black theme :

![image](https://user-images.githubusercontent.com/16088743/108196448-6a82a500-7119-11eb-80d5-24807860633e.png)


Of course we should add all units, etc.. 

Note : adding a Spectrum.to_pandas() function could help here, as well as exporting all quantities to .csv as suggested by @ulyssed

erwanp:
[...]

PR #309, opened 2021-07-07
Improve radis import speed

erwanp:
### Description

Improved RADIS import time by delaying the import of packages not necessarily used immediatly (or ever). 
(also added a test for Cantera calculations on the default Travis test environment)

Tested with [tuna](https://github.com/nschloe/tuna)

```
python -X importtime -c "import radis" 2> radimport.log
tuna radimport.log
```

Before : 

image
image

After:

image

So import is 75% faster (7s -> 4s) ! 

Delayed : 
- [x] Cantera      (rarely used)
- [x] Matplolib    (maybe not necessarily used in the future ; with a Bokeh/Plotly backend #179 )
- [x] Astroquery   (used only if fetching from HITRAN) 

What's left   (not for this PR)
- [ ] Astropy units . ; it's hard to delay Astropy because we need Astropy.units to know if user give Astropy.units. A way would be to "duck-test" it; i.e. evaluate a property. 
- [ ] Numba & underlying llvm  : is used here and there; in particular for Whiting's JIT. Would be a lot of work to delay this ; and may leave some bugs undected.
- [x] Pandas / Scipy / Numpy are required anyway. 

Notice the time it takes Numba to compute Whiting's JIT.

### DIFF
--- a/radis/io/query.py
+++ b/radis/io/query.py
@@ -23,8 +23,6 @@ from os.path import exists, isfile, join
 
 import numpy as np
 import pandas as pd
-from astropy import units as u
-from astroquery.hitran import Hitran
 
 import radis
 from radis.db.classes import get_molecule, get_molecule_identifier
@@ -79,6 +77,9 @@ def fetch_astroquery(
     :py:attr:`astroquery.query.BaseQuery.cache_location`
 
     """
+    from astropy import units as u
+    from astroquery.hitran import Hitran
+
     # Check input
     if not is_float(molecule):
         mol_id = get_molecule_identifier(molecule)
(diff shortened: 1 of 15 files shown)


gagan-aryan:
@erwanp Does the tool render different maps on different machines? I fetched this branch and ran tuna and I seemed to receive a different map. But there was a significant difference with this branch though.

radis/io/query.py
```diff
@@ -79,6 +77,9 @@ def fetch_astroquery(
     :py:attr:`astroquery.query.BaseQuery.cache_location`
 
     """
+    from astropy import units as u
```
gagan-aryan, line 80: Basically import packages when necessary (?)

erwanp: Yes ! Imports within functions is harder to maintain so you don't want to do that for things you'll have to load anyway. 

For astropy.units it didn't really work; as it's still imported at one point to check the inputs.

Issue #136, opened 2020-12-04
Add (more) mini galleries of examples in Documentation


erwanp:
### 🎯 Describe what you'd like

Add examples and Tutorials that shows up in the Example Gallery . Inspiration : https://psyplot.readthedocs.io/en/latest/index.html

Status : 
- ✔️ Gallery examples working, see https://radis.readthedocs.io/en/latest/auto_examples/index.html 
- 💪 Help needed to build more examples !

# Ideas of examples

- [x] Import HITEMP files (turn [radis/hitemp](https://github.com/radis/hitemp) into a Notebook ?)
- [ ] Post-processing methods, in particular inspired from [The Spectrum object page](https://radis.readthedocs.io/en/latest/spectrum/spectrum.html#) 
- [x] compute a spectrum on different continuous spectral ranges. See [example in MergeSlabs](https://radis.readthedocs.io/en/latest/source/radis.los.slabs.html#radis.los.slabs.MergeSlabs)  
- [x] Initialize a [SpecDatabase](https://radis.readthedocs.io/en/latest/source/radis.tools.database.html#radis.tools.database.SpecDatabase) and show the output of the generated summary .csv file, to print & compare your experiment conditions.  
- [ ] Example printing the default spectroscopic constants used 
- [ ] Notebook : how to compute a high-temperature HITEMP H2O spectrum ? (asked by a user on [Twitter](https://twitter.com/zz99583643/status/1408105514764963841))
- [ ] Cantera example with [get_eq_mole_fractions](https://radis.readthedocs.io/en/latest/gen_modules/radis.tools.html?highlight=eq_mole#radis.tools.get_eq_mole_fraction)

See also : 
- [Ideas of Dashboards](https://github.com/radis/radis-lab/issues/23)

# How to write an example ? 

To write an example, all you have to do really is add a .py file in https://github.com/radis/radis/tree/develop/examples 
- The best examples include a plot : for the plot to be visible on the Gallery, the file name should starts with `plot_*`  and at least one of the commands should return a figure. 
- To include the Galleries in a function documentation, or in another .rst file from the documentation, you should add the following line : 
```
.. minigallery:: radis.calc_spectrum
```
which will add a nice gallery of all examples using this function (present case : `calc_spectrum`). 
To know which path to use (is it `radis.calc_spectrum`, or `radis.lbl.calc.calc_spectrum` ?), have a look at the files generated in `radis/docs/source/backreferences` after building the docs locally.

[...]

PR #315, opened 2021-07-13
Add: SpecDatabase Example in Gallery

anandxkumar:
### Description

This pull request adds example of SpecDatabase in Gallery.
Screenshot from 2021-07-14 00-41-14

### DIFF
--- a/radis/tools/database.py
+++ b/radis/tools/database.py
@@ -1668,6 +1668,8 @@ class SpecList(object):
 
             >>> db.plot(Tvib, Trot, residual)     # where residual is calculated by a fitting
                                                   # procedure...
+        -------
+        .. minigallery:: radis.tools.database.SpecList.plot_cond
         """
         # %%
 
@@ -1686,7 +1688,7 @@ class SpecList(object):
 
         # Overlay color
         if z_value is not None:
-            assert (len(z_value)) == len(self.df)
+            assert (len(self.df[z_value])) == len(self.df)
 
             z = np.array(z_value) ** 0.5  # because the lower the better
 
@@ -1805,7 +1807,6 @@ class SpecDatabase(SpecList):
         requires all conditions to be either float, string, or boolean. List
         won't work!
 
-
     See Also
     --------
     :func:`~radis.tools.database.load_spec`,
@@ -1827,7 +1828,10 @@ class SpecDatabase(SpecList):
 
     Compare another Spectrum to all spectra in the database:
 
-    :meth:`~radis.tools.database.SpecDatabase.fit_spectrum`,
+    :meth:`~radis.tools.database.SpecDatabase.fit_spectrum`
+
+
+    .. minigallery:: radis.SpecDatabase
     """
 
     def __init__(
--- a/radis/tools/database.py
+++ b/radis/tools/database.py
@@ -1688,7 +1688,6 @@ class SpecList(object):
 
         # Overlay color
         if z_value is not None:
-            print(type(z_value))
             if type(z_value) is str:
                 z = self.df[z_value]
             else:
--- a/radis/tools/database.py
+++ b/radis/tools/database.py
@@ -1688,17 +1688,22 @@ class SpecList(object):
 
         # Overlay color
         if z_value is not None:
-            assert (len(self.df[z_value])) == len(self.df)
+            print(type(z_value))
+            if type(z_value) is str:
+                z = self.df[z_value]
+            else:
+                z = z_value
 
-            z = np.array(z_value) ** 0.5  # because the lower the better
+            assert len(z) == len(self.df)
 
+            z = np.array(z) ** 0.5  # because the lower the better
             #            norm = cm.colors.Normalize(vmax=z.max(), vmin=z.min())
             #            cmap = cm.PRGn
 
             xarr = np.linspace(min(x), max(x))
             yarr = np.linspace(min(y), max(y))
             mx, my = np.meshgrid(xarr, yarr)
-            zgrid = griddata((x, y), z, (mx, my), method="linear")
+            zgrid = griddata((x, y), z, (mx, my), method="nearest")
             levels = np.linspace(min(z), max(z), 20)
             ax.contourf(
                 mx,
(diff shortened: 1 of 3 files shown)


erwanp:
Hello !! Nice to see SpecDatabase being advertised. 

A few things to make it great : 

- for the example to work on ReadTheDocs, they'll need to be generated online. For this you cannot use a local database. fetch_hitemp("CO") may work; but it's a slightly big database (no good for 🌳 , remember it's run on every merge to `develop`!); what about using "OH" for your example ? 
- for the image, can you try to use [plot_cond](https://radis.readthedocs.io/en/master/source/radis.tools.database.html#radis.tools.database.SpecList.plot_cond) which plots a 2D map of the database conditions ? 
- could you add a 4th method where you make use of the init_database class of SpectrumFactory ? 
- can you add some clickable links to the main classes and methods in the description, i.e. `:py:class:`~radis.tools.database.SpecDatabase`  ; same with plot_cond and init_database ? 
- last but not least, the example gallery can appear in the docstrings of the functions and methods used. For that, you'll need to add  ``` .. minigallery:: radis.PATH/TO/FUNCTION``` in their docstrings. See https://github.com/radis/radis/issues/136 to know which path to use

erwanp:
@gugzy and @CorentinGrimaldi , adding you for Review as you use the SpecDatabases ; if you think of features that should appear on the example gallery !

radis/tools/database.py
```diff
@@ -1686,7 +1688,7 @@ def plot_cond(self, cond_x, cond_y, z_value=None, nfig=None):
 
         # Overlay color
         if z_value is not None:
-            assert (len(z_value)) == len(self.df)
+            assert (len(self.df[z_value])) == len(self.df)
 
             z = np.array(z_value) ** 0.5  # because the lower the better
```
anandxkumar, line 1693: @erwanp I believe the above line `assert (len(z_value)) == len(self.df)` needs to be changed into `assert (len(self.df[z_value])) == len(self.df)` because z_value is just the column name.

Also the above below gives error:
 ```
   z = np.array(z_value) ** 0.5  # because the lower the better

TypeError: ufunc 'power' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe
```

**Comment by anandxkumar** [comment]
Screenshot from 2021-07-14 19-18-15-1

Plotting 2 graphs, one simple plot of `radiance_noslit` of different spectrums in SpecDatabase and other Tgas vs wstep using plot_cond(). Any change in the example?

Also added the mini gallery to SpecDatabase, plot_cond, and init_database. All getting properly rendered.

File: radis/tools/database.py (hunk 1 above)
Comment (erwanp) [reply radis/tools/database.py:1693]: I looks like that z_value could be used as an external array. What about : 
- if z_value is a `str` ; and a column name, then use `z_value = self.df[z_value]`
- else let it as it is (so the user can still use an external array of any kind) 

```np.array(z_value) ** 0.5```
looks like something added for specific (old) purposes, it could be removed.

File: radis/tools/database.py [hunk 2]
```diff
@@ -1827,7 +1828,10 @@ class SpecDatabase(SpecList):
 
     Compare another Spectrum to all spectra in the database:
 
-    :meth:`~radis.tools.database.SpecDatabase.fit_spectrum`,
+    :meth:`~radis.tools.database.SpecDatabase.fit_spectrum`
+
+
+    .. minigallery:: radis.SpecDatabase
```
erwanp, line 1834: Great !

erwanp (changes requested):
Tiny changes, looks great!!

radis/tools/database.py
anandxkumar: Plotted: `db_new.plot_cond('Tgas', 'wstep', 'wstep')` where wstep value was decreasing, got the below plot-

Figure_1

I believe the color of the point should be different instead of the whole area?

erwanp: I think I use it to plot the residuals in Fitroom, if you had a look at it. So the area is what we want. Let's use it : maybe compute  small matrix of Tgas and wstep (15-20 spectra max, but larger ranges, like up to 3000 K to show some differences) , and plot the calculation time as the Z_label? 

It's just to inspire people, no need to spend much time here :)

anandxkumar: Added a better example where I'm halving the wstep at every iteration. Now the difference is clearer and we can infer that Violent -> Minimum calc time, yellow -> Maximum Calc Time. Also, the max time taken is `0.17s` and overall takes around `0. 40` seconds to fully execute it so it's not expensive.

Screenshot from 2021-07-15 03-49-33

erwanp:
Merging to see if it works on the online docs!

erwanp (approved):

erwanp:
Hello @anandxkumar ; the [docs build failed](https://readthedocs.org/projects/radis/builds/14245271/) : 

```
generating gallery for auto_examples... [ 90%] plot_SpecDatabase.py
WARNING: /home/docs/checkouts/readthedocs.org/user_builds/radis/checkouts/develop/examples/plot_SpecDatabase.py failed to execute correctly: Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/radis/checkouts/develop/examples/plot_SpecDatabase.py", line 37, in <module>
    db = SpecDatabase(my_folder, lazy_loading=False)
  File "/home/docs/checkouts/readthedocs.org/user_builds/radis/envs/develop/lib/python3.6/site-packages/radis/tools/database.py", line 1867, in __init__
    os.mkdir(path)
FileNotFoundError: [Errno 2] No such file or directory: '/home/pipebomb/Desktop/SpecDatabase_Test/'

generating gallery for auto_examples... [100%] plot_linestrengths.py
```

Can you use relative links to build the database ?

---

Also; can you use the new PR to build a (wstep, Tgas) matrix of spectra rather than 25 spectra with different Tgas. Will be good-looking with plot_cond. 
Like `wstep = [0.001 :: 0.1, 4]    ;  Tgas = [300 :: 3000, 5 ]`   (20 spectra total)
(```warnings={"AccuracyError":"ignore"}```  may be needed

PR #318, opened 2021-07-16
Switch to memory_usage(deep=false)

anandxkumar:
### Description

This pull request is to change `sys.getsizeof()` with `memory_usage(deep=false)` and checks if `object` type column is present in df, if yes we will have a different threshold to compare with.

On trying to get the relation of the number of lines, calculation time and memory_usage(deep=false) no relation could be derived between time taken to load a line or column. Below is the plot between `N_lines`, ` Time_taken` by memory_usage(deep=false) and size of data frame as heat map:

Time Taken
Clearly the computation is so fast that its dependence is negligible on the number of lines and we can expect a constant value for time taken irrespective size of dataframe.

Replacing `sys.getsizeof` with `memory_usage(deep=false)`  saved us around `20-50%` time: A simple comparison below for CH4 for different number of lines
```# Computation parameters
wmin = 2000
wmax = 4200
wstep = 0.01
T = 300.0 #K
p = 1 #bar
broadening_max_width=5
molecule = 'CH4'
```
Screenshot from 2021-07-17 04-31-11

Also reinitialize is dominated by `self.df1 = self.df0.copy()` (here reinitialize_1):
Reinitialize_new

### DIFF
--- a/radis/lbl/base.py
+++ b/radis/lbl/base.py
@@ -68,6 +68,7 @@ import numpy as np
 import pandas as pd
 from astropy import units as u
 from numpy import exp, pi
+from psutil import virtual_memory
 
 import radis
 
@@ -3253,11 +3254,22 @@ class BaseFactory(DatabankLoader):
 
         # Check memory size
         try:
+            # Retrieving total user RAM
+            mem = virtual_memory()
+            mem = mem.total  # total physical memory available
             # Checking if object type column exists
             if "O" in self.df1.dtypes.unique():
-                limit = 80e6
+                limit = (
+                    mem / 25  # 4% of user RAM
+                )  # Since the memory_usage(deep=False) will be alot less than actual
+                self.warn(
+                    "'object' type column found in database, calculations and "
+                    + "memory usage would be faster with a numeric type. Possible "
+                    + "solution is to not use 'save_memory' and convert the columns to dtype."
+                )
             else:
-                limit = 500e6
+                limit = mem * 2 / 5  # 40% of User available RAM
+
             df_size = self.df1.memory_usage(deep=False).sum()
 
             if df_size > limit:
--- a/radis/lbl/base.py
+++ b/radis/lbl/base.py
@@ -3259,16 +3259,17 @@ class BaseFactory(DatabankLoader):
             mem = mem.total  # total physical memory available
             # Checking if object type column exists
             if "O" in self.df1.dtypes.unique():
-                limit = (
-                    mem / 25  # 4% of user RAM
-                )  # Since the memory_usage(deep=False) will be alot less than actual
+                limit = mem / 25  # 4% of user RAM
                 self.warn(
                     "'object' type column found in database, calculations and "
                     + "memory usage would be faster with a numeric type. Possible "
-                    + "solution is to not use 'save_memory' and convert the columns to dtype."
+                    + "solution is to not use 'save_memory' and convert the columns to dtype.",
+                    "PerformanceWarning",
                 )
             else:
-                limit = mem * 2 / 5  # 40% of User available RAM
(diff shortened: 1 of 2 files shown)


erwanp:
Hello! So almost 2x faster spectra for millions of lines, that's very good! 

Can you : 
- compare memory limit to the user available RAM (see psutils)  rather than a hard-coded value? For instance add a warning if 30-50% of ram? (remember the rule of thumb from your first blog) 
- use the same limit whether there is object type or not. It does not matter, what matters is user available RAM. 

Checking for "object"  type is good,  because eventually we should make sure no object type remains the database. You could add a new warning , if "object" type is found in database, saying that calculations and memory usage would be faster with a numeric type.

anandxkumar:
@erwanp, when we are using `self.df1.memory_usage(deep=False).sum()` and the dataframe has object type column, the difference between deep=True and deep=False is around `4 times` so we need to set different percentage for both with and without object type column. I ran a simple example of CH4 for number of line=`4366936`:

sf.df1.memory_usage(deep=False).sum() = 628838912
sf.df1.memory_usage(deep=True).sum() = 2646363344 (4.2 times larger)
mem = virtual_memory().total = 16638660608 (16GB)

upon running the Spectrum my memory was filled upto 75%, so I guess it will be a safe bet to give warning  at this point.

`mem/25 = 665546424 ~ sf.df1.memory_usage(deep=False).sum()` so a 4 % threshold came out! Tried it with other cases like CO and NO and for even large spectrum it was very less (for 711929 lines of NO it was around 1/6th of threshold). 

*NOTE* - Also note seems like every Spectrum except CO2 (which is loaded manually) uses object type column so we will get warning for almost every Spectrum for object column.

Also travis failed due to server issue I guess because it passed locally?

erwanp:
Ok,  good explanation. Can you add this comment "the difference between deep=True and deep=False is around 4 times"  directly in the code? 

You can also give a type to the warning with the second key of self.warn (allows user to deactivate them if required), I think there is something as PerformanceWarning, would be appropriate

erwanp (approved):

anandxkumar:
Something seems wrong with the hitran server, I deleted `.radisdb` and my pytest is stuck on `test_hitemp.py` even on develop branch, can you confirm it?

erwanp:
Yes; same here; will retry tests tomorrow. 

Your treshold seems very small, it triggers a warning at 1/25 * 4 ~ 16% of total RAM; set this up to 50% of RAM ? Sounds more reasonable.

erwanp:
Tests pass, just update the limit to about 50% and we can merge

anandxkumar:
The same issue as before passed locally but the following error in `fetch_astroquery`:
`ValueError: invalid literal for int() with base 10: '<!'`

radis/lbl/base.py
```diff
@@ -3267,9 +3267,9 @@ def _reinitialize(self):
                     "PerformanceWarning",
                 )
             else:
-                limit = (
-                    mem / 25 * 4
-                )  # the difference between deep=True and deep=False is around 4 times
+                limit = mem / 2  # 50 % of user RAM
+
+            # Note: the difference between deep=True and deep=False is around 4 times
```
erwanp, line 3272: you didn't fix in the "object" case neither?

anandxkumar: Actually, this case might be deceiving a little. When using CH4 for 4 million lines, it was occupying 75% of RAM already and value of sf.df.memory_usage(deep=False) was only 4% of the total memory, that's why I doubted a larger threshold might lead to full usage of memory already.

erwanp: Ok. Let's try like this ; we'll see after usage and adjust if needed.

PR #325, opened 2021-07-27
use tuna 🐟  to generate visual/interactive performance profiles

erwanp:
### Description

use [tuna](https://github.com/nschloe/tuna) to generate visual/interactive performance profiles; and the update the console profiler:

Syntax : 

```
s = calc_spectrum(...)
s.generate_perf_profile()
```

Typical output (opens a local web page) : 

image

Console profiler:

```
s = calc_spectrum(...)
s.print_perf_profile()
```

Typical output : 

```
        spectrum_calculation      0.844s ████████████████
            check_line_databank              0.001s
            check_non_eq_param               0.070s █
            fetch_energy_5                   0.027s
            calc_weight_trans                0.016s
            reinitialize                     0.003s
                copy_database                    0.000s
                memory_usage_warning             0.003s
                reset_population                 0.000s
            calc_noneq_population            0.066s █
                part_function                    0.056s █
                population                       0.010s
            scaled_non_eq_linestrength       0.004s
                map_part_func                    0.001s
                corrected_population_se          0.003s
            calc_emission_integral           0.008s
            applied_linestrength_cutoff      0.003s
            calc_lineshift                   0.001s
            calc_hwhm                        0.008s
            generate_wavenumber_arrays       0.002s
            calc_line_broadening             0.668s ████████████
                precompute_DLM_lineshapes        0.020s
                DLM_Initialized_vectors          0.000s
                DLM_closest_matching_line        0.001s
                DLM_Distribute_lines             0.002s
                DLM_convolve                     0.304s █████
            calc_other_spectral_quan         0.005s
            generate_spectrum_obj            0.000s
```

### Implementation

- [x] Proof of concept
- [x] Test
- [x] update "profiler" format consistently : @anandxkumar see below 
- [x] mention in docs/performances

@anandxkumar this expects the `conditions["profiler"]` dict to be of the format:  
```python
[...]

PR #329, opened 2021-07-27
Fixes: Issue322 - update Profiler

### LINKED ISSUE #322: Profiler doesnt sum to total calculation time
### 🐛 Describe the bug

Total time registered in the profiler doesn't sum to the total calculation time. Actually, it's about 2x the calculation time, because some parameters are counted twice, for instance `part_function` and `calc_noneq_population_multiTvib` 

### 🖥 Steps To Reproduce

```python
s  = calc_spectrum(2284.2, 2284.6,
                     Trot=491,
                     Tvib=(517, 517, 2641),
                     wstep='auto', 
                     pressure=20*1e-3,           # bar
                     molecule='CO2',
                     isotope='1,2',
                     path_length=10,             # cm-1
                     mole_fraction=0.1*28.97/44.07,
                     broadening_max_width=1,     # cm-1
                     medium='vacuum',
                     databank='HITEMP-CO2-TEST',
                     )

print(s.conditions['profiler'])
>>> {'fetch_energy_6': 0.9933407306671143, 'calc_weight_trans': 0.08555841445922852, 'check_non_eq_param': 1.1167986392974854, 'part_function': 0.13460326194763184, 'populations': 0.00997471809387207, 'calc_noneq_population_multiTvib': 0.1445779800415039, 'map_part_func': 0.0, 'corrected_population_se': 0.003988027572631836, 'scaled_non_eq_linestrength': 0.003988027572631836, 'calc_emission_integral': 0.01595783233642578, 'applied_linestrength_cutoff': 0.003989458084106445, 'calc_lineshift': 0.0009968280792236328, 'calc_hwhm': 0.014960289001464844, 'precompute_DLM_lineshapes': 0.0009970664978027344, 'DLM_Initialized_vectors': 0.0, 'DLM_closest_matching_line': 0.0, 'DLM_Distribute_lines': 0.0, 'DLM_convolve': 0.0, 'calc_line_broadening': 0.0019965171813964844, 'calc_other_spectral_quan': 0.0049893856048583984, 'spectrum_calc_before_obj': 1.3212203979492188}

print(sum(list(s.conditions["profiler"].values())))
>>> 3.8579375743865967

print(s.conditions['calculation_time'])
>>> 1.3212203979492188
```

### 🎯 Expected behavior

Store calculation times should be a dict with nested levels. 
Total of each level should be (approximately) correct
We should add a test (with some tolerance)

anandxkumar:
### Description

This pull request modifies the profiler to store time in a tree structure form. 

```
sf = SpectrumFactory(wavenum_min=2900,
                     wavenum_max=4200,
                     molecule='CO',
                     broadening_max_width=10,  # cm-1
                     medium='vacuum',
                     verbose=0,    # more for more details
                     pressure = 10,
                     wstep = 0.01,
                     )

sf.fetch_databank('hitemp')
s1 = sf.eq_spectrum(Tgas=300, path_length=1)

c = sf.profiler.final
print(yaml.dump(c,allow_unicode=True, default_flow_style=False))
```

Output:
```
spectrum_calculation:
  calc_hwhm: 0.011423110961914062
  calc_line_broadening:
    DLM_Distribute_lines: 0.018210649490356445
    DLM_Initialized_vectors: 1.2159347534179688e-05
    DLM_closest_matching_line: 0.0053594112396240234
    DLM_convolve: 1.6507983207702637
    precompute_DLM_lineshapes: 0.05847477912902832
    value: 1.7343709468841553
  calc_lineshift: 0.0011224746704101562
  calc_other_spectral_quan: 0.0058553218841552734
  check_line_databank: 0.0006852149963378906
  generate_spectrum_obj: 0.0006301403045654297
  generate_wavenumber_arrays: 0.004528522491455078
  reinitialize:
    copy_database: 0.0014543533325195312
    memory_usage_warning: 0.0031118392944335938
    reset_population: 8.177757263183594e-05
    value: 0.0046651363372802734
  scaled_eq_linestrength: 0.013865232467651367
  spectrum_calc_before_obj: 1.7766413688659668
  value: 1.777277946472168
  ```

Fixes #322 <Issue Number>

### DIFF
--- a/radis/lbl/factory.py
+++ b/radis/lbl/factory.py
@@ -1674,6 +1674,8 @@ class SpectrumFactory(BandFactory):
         # New Profiler object
         self.profiler = Profiler(verbose)
 
+        self.profiler.start("optically_thin_power_calculation", 1)
+
         # Make sure database is loaded
         if self.df0 is None:
             if not self.save_memory:
@@ -1737,6 +1739,10 @@ class SpectrumFactory(BandFactory):
 
         # Optically thin case (no self absorption):
         Ptot = Pv * path_length  # (mW/sr/cm2)
+
+        self.profiler.stop(
+            "optically_thin_power_calculation", "Optically thin power calculation"
+        )
 
         return conv2(Ptot, "mW/cm2/sr", unit)
 
--- a/radis/lbl/factory.py
+++ b/radis/lbl/factory.py
@@ -659,8 +659,9 @@ class SpectrumFactory(BandFactory):
         # %% Start
         # --------------------------------------------------------------------
 
-        self.profiler.start("spectrum_calc_before_obj", 2)
         self.profiler.start("spectrum_calculation", 1)
+        self.profiler.start("spectrum_calc_before_obj", 2)
+
         if verbose:
             self.print_conditions("Calculating Equilibrium Spectrum")
 
@@ -947,8 +948,8 @@ class SpectrumFactory(BandFactory):
         Ia_arr[np.isnan(Ia_arr)] = 0
         molarmass_arr[np.isnan(molarmass_arr)] = 0
 
-        self.profiler.start("spectrum_calc_before_obj", 2)
         self.profiler.start("spectrum_calculation", 1)
+        self.profiler.start("spectrum_calc_before_obj", 2)
 
         # generate the v_arr
         v_arr = np.arange(
--- a/radis/lbl/factory.py
+++ b/radis/lbl/factory.py
@@ -80,6 +80,7 @@ from warnings import warn
 
 import astropy.units as u
 import numpy as np
+import yaml
 from numpy import arange, exp
 from scipy.constants import N_A, c, k, pi
 
@@ -1736,6 +1737,27 @@ class SpectrumFactory(BandFactory):
         )
 
         return conv2(Ptot, "mW/cm2/sr", unit)
+
+    def print_perf_profile(self):
+        """Prints Profiler output dictionary in a structured manner.
+        example:
+        SpectrumFactory.print_perf_profile()
+        ----------------------------
+        spectrum_calculation:
+            calc_hwhm: 0.007350444793701172
+            calc_line_broadening:
+                DLM_Distribute_lines: 0.005137443542480469
+                DLM_Initialized_vectors: 8.106231689453125e-06
+                DLM_closest_matching_line: 0.0010995864868164062
+                DLM_convolve: 0.6121664047241211
+                precompute_DLM_lineshapes: 0.027348995208740234
+                value: 0.6468849182128906
+            calc_lineshift: 0.0005822181701660156
+            calc_other_spectral_quan: 0.005536317825317383
+            value: 0.6814641952514648
+        """
+        profiler = self.profiler.final
+        print(yaml.dump(profiler, allow_unicode=True, default_flow_style=False))
 
 
 # %% ======================================================================
--- a/radis/lbl/factory.py
+++ b/radis/lbl/factory.py
@@ -1254,8 +1254,8 @@ class SpectrumFactory(BandFactory):
         # %% Start
         # --------------------------------------------------------------------
 
-        self.profiler.start("spectrum_calc_before_obj", 2)
         self.profiler.start("spectrum_calculation", 1)
+        self.profiler.start("spectrum_calc_before_obj", 2)
         if verbose:
             self.print_conditions("Calculating Non-Equilibrium Spectrum")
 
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: radis/misc/profiler.py [hunk 1]
```diff
@@ -50,14 +52,44 @@ def start(self, key, verbose_level, optional=""):
...
+                self.final[self.stack[verbose_level - 1][-1]].update({key: {}})
+            else:
+                self.final[self.stack[verbose_level - 2][-1]][self.stack[2][-1]].update(
+                    {key: {}}
+                )
         if len(optional) != 0 and self.verbose >= verbose_level:
             print(optional)
 
-    def stop(self, key, details=""):
+    def stop(self, key, details, end=True):
```
erwanp, line 69: do you ever need `end` ?

erwanp (approved):
Well done ! Now let's see if it's compatible with #325

erwanp (changes requested):
Actually, I do not understand why you need two extra attributes `stack` and `final` here.

Let's say you've open a counter at level v=2. 
Now : 
- if you open a new counter at v>2; then you'll add it as a child.
- if you close the same counter at v==2, fine, append to `value`. 
- if you close another counter at v<=2 there's a problem and you should raise an error .

erwanp (approved):

radis/lbl/factory.py
```diff
@@ -739,21 +740,15 @@ def eq_spectrum(
         conditions = self.get_conditions()
         conditions.update(
             {
-                "calculation_time": self.profiler.dict_time["spectrum_calc_before_obj"][
-                    0
+                "calculation_time": self.profiler.final[list(self.profiler.final)[-1]][
+                    "spectrum_calc_before_obj"
```
anandxkumar, line 744: @erwanp  since `spectrum_calc_before_obj` sums up all steps, we should remove this and maybe add a sum function which will add time taken per verbose level; Calculation time should be equal to the whole time taken by spectrum till generating and saving the spectrum?

radis/spectrum/spectrum.py
```diff
@@ -54,6 +54,7 @@
 
 import astropy.units as u
 import numpy as np
+import yaml
```
erwanp, line 57: @anandxkumar is it in the environment.yml and setup.py ?

radis/spectrum/spectrum.py
```diff
@@ -3035,6 +3036,25 @@ def is_optically_thin(self):
                 + "value manually with s.conditions['self_absorption']=..."
             )
 
+    def print_struct(self):
```
erwanp, line 3039: - You could create a method at the SpectrumFactory level that calls the latest profiler with `print_struct` ? Like `sf.print_perf_profile` . 

- maybe also in the Spectrum object itself. `Soectrum.print_perf_profile`

- Also add a small mention under ["Performance" section in the docs](https://radis.readthedocs.io/en/latest/lbl/lbl.html#performance), for instance : 

> Monitor performances
> -------------------------
> 
> Radis allows you to profile the spectrum calculation so you can further optimize it. Use :py:meth:`~radis.lbl.factory.print_perf_profile` to print the time spent on each step of the calculation. Typical output :: 
>  XXX  
> Results are also stored in a Spectrum object so you can compare them a posteriori. Use :py:meth:`~radis.spectrum.spectrum.Spectrum.print_perf_profile`. Calculation details are also shown during the calculation itself if using a sufficiently high verbose level (e.g. : `verbose=3`)

erwanp (approved):

anandxkumar:
@erwanp seems like the code and quality checks have disappeared?

erwanp (approved):

radis/misc/profiler.py
```diff
@@ -40,8 +41,10 @@ def __init__(self, verbose):
 
         # Dev: Init here to be found by autocomplete
         self.initial = {}
-        self.dict_time = {}
         self.verbose = verbose
+        self.current_verbose2_counter = ""  # Holds current Verbose level 2 key
+        self.final = OrderedDict()
+        self.flag = False  # For checking transit between verbose level 3 to 2
```
erwanp, line 47: @anandxkumar I merged to test on my own PR #325 , and I got it working. 

However, this part of the code here looks a bit complicated : why do you need a special "flag" for difference level 2 and 3 ? What if there's level 4, or 5 ?  Would it still work ?

anandxkumar: No in that case it shouldn't work, because I wrote it assuming we are dealing with 3 levels. The reason I put a flag here is to see if we are transit from level 3 to level 2 when using `profiler.stop`. If yes then we append "value". But a better approach will be to check the len of that key in the dictionary. if `length` != 0 then we add it in the form 'value': val. This will remove the need of `self.flag`. In order to generalize it I will need to do some modifications adding the key in the dictionary so it will require keeping track of the all the previous levels latest key to add the current key in it (self.final['key1']['key2']['key3'].update({'currentkey':{}}) and yet this approach will rely on the number of levels we have overall.

PR #330, opened 2021-07-31
Updated Algorithm for storing time in profiler dictionary

anandxkumar:
Using a recursive approach to add entity in self.profiler.final dictionary removing the need of flag variable and making it compatible with n verbose levels (n>0).


erwanp:
Good improvement!

PR #332, opened 2021-08-03
Appends calculation time if same key/step is executed

anandxkumar:
Calculation time is now added into existing key instead of overwritting it. Removed functions that aren't used anymore.


anandxkumar:
@erwanp, upon running pytest, `spectrum.prof` file is created in the base directory when running `test_util`. I believe that shouldn't happen, and if it does then it should be deleted after test is successfully executed?

radis/misc/profiler.py
```diff
@@ -48,7 +48,12 @@ def __init__(self, verbose):
     # Creates profiler dictionary structure
     def add_entry(self, dictionary, key, verbose, count):
         if count == verbose:
-            dictionary[key] = {}
+            try:
+                if dictionary[key]:
+                    return
+            except:
+                dictionary[key] = {}
```
erwanp, line 55: can you have an "if" only verison ; or define the error you expect ? Also, there is no "else" here

radis/misc/profiler.py
```diff
@@ -71,10 +73,30 @@ def start(self, key, verbose_level, optional=""):
...
+            if isinstance(dictionary[key], float):
+                time_calculated = dictionary[key] + time_calculated
+                executed_more_than_once = True
+            try:
+                if "value" in dictionary[key] and isinstance(
+                    dictionary[key]["value"], float
+                ):
+                    time_calculated = dictionary[key]["value"] + time_calculated
+                    executed_more_than_once = True
+            except:
```
erwanp, line 87: define the expected error

radis/misc/profiler.py
```diff
@@ -71,10 +73,30 @@ def start(self, key, verbose_level, optional=""):
...
+                pass
+
+            if executed_more_than_once:
+                try:
+                    dictionary[key]["value"] = time_calculated
+                except:
```
erwanp, line 93: same

erwanp (changes requested):

radis/test/spectrum/test_utils.py
```diff
@@ -118,7 +122,9 @@ def test_perf_profile_from_factory(*args, **kwargs):
     with open("spectrum.prof", "rb") as f:
         assert "calculation_time" in str(f.read())
 
+    os.remove("spectrum.prof")
+
 
 if __name__ == "__main__":
     test_perf_profile()
-    # test_perf_profile_from_factory()
```
anandxkumar, line 124: @erwanp why was it commented out?

erwanp: Just because I was focusing on the other, and forgot to uncommemt. It does not really matter : Pytest picks up all functions starting with test_* . They'll be run whether they're called in the main or not

erwanp (approved):

erwanp (approved):

erwanp:
Approved, I let you merge!

PR #343, opened 2021-08-12
Add predict_time(),  make broadening_method = "voigt" as default , add default truncation of 50 cm-1 

### LINKED ISSUE #340: Add a 50cm-1 lineshape truncation by default

Since 0.9.20 (Aug. 2019  https://github.com/radis/radis/pull/39  RADIS has been using the DIT  (or DLM , we're not set on the name yet!) method as the default method to compute spectra, with a FFT algorithm to resolve the convolutions. The DIT+FFT approach has no lineshape truncatino, which is a good thing for small spectra, where legacy LBL codes usually rely on a truncation of 5, 10 or 50 cm-1 to improve computational performance at the cost of accuracy.

However, in the far wings (few hundreds of cm-1) lineshapes exhibit a sub-lorentzian behavior : therefore, for large spectra, having no truncation is less physically accurate than having a truncation (see discussion on [💬 Slack][https://radis-radiation.slack.com/archives/C01JBV54G11/p1610183178001400) )

Here, we recommend to use a default truncation of 300 cm-1. More accurate, molecule-dependant truncations may be implemented later.

It is hard to implement a truncation with a FFT algorithm (discussion with @dcmvdbekerom for details).  However, a truncatino in wavenumber space is easy. RADIS already has a DIT+analytical Voigt (in wavenumber space) implement ; it's actually the first implementation that was done #37. Until this version, this version was lagging behind DIT+FFT in terms of performance ; making it unpractical for large number of lines. Recent change #323  made DIT+Voigt competitive; and even faster than DIT+FFT in most cases (see the extensive benchmarks done by @anandxkumar in https://anandxkumar.github.io/Benchmark_Visualization_GSoC_2021 . Version post #323 is the one with `oa.convolve` ) 

The only drawback is that the analytical Voigt introduces some errors compared to real-Voigt. For small spectra, close to the line centers, we should keep the DIT+FFT approach (real Voigts). 

Suggested implementation, in `calc_spectrum` : 
- calculations with less than 300cm-1 range : keep using DIT+FFT as default
- calculations  with spectral range > 300 cm-1 : use DIT+Voigt with a 300cm-1 cutoff

### Example

A comparison of both methods using the CO2 HITEMP database (fetched automatically with #281 )   (example adapted from @anandxkumar 's benchmarks)

```python

from radis import plot_diff

# %% Condition Parameters
wmin = 500
wmax = 4000
dv = 0.01
T = 300.0 #K
p = 1 #bar
broadening_max_width=300    # lineshape broadening width, ext​

sf = SpectrumFactory(wavenum_min=wmin, wavenum_max=wmax, 
                  pressure=p,
                  isotope='1,2,3',
                  molecule='CO2',
                  wstep=dv,
                  broadening_max_width=broadening_max_width, 
                  cutoff=0, # 1e-27,
                  verbose=3,
                  )

sf.fetch_databank('hitemp')

#%% Calculate DIT + FFT
sf.params['optimization'] = 'simple'
sf.params['broadening_method'] = 'fft'

s_ditfft = sf.eq_spectrum(T, name = "DIT+FFT")

#%% Voigt 
sf.params['optimization'] = 'simple'
sf.params['broadening_method'] = 'voigt'

s_ditvoigt = sf.eq_spectrum(T, name = f"DIT+Voigt ({broadening_max_width:.0f}"+r"cm$^{-1}$ trunc)")


# %% Compare

s_ditfft.print_perf_profile()
s_ditvoigt.print_perf_profile()
fig, [ax0, ax1] = plot_diff(s_ditfft, s_ditvoigt, "abscoeff")
ax0.set_yscale("log")
ax0.set_ylim(ymin=1e-13, ymax=1e3)
ax1.set_ylim(ymin=-0.0025, ymax=0.0025)

assert s_ditfft.conditions["lines_calculated"] == s_ditvoigt.conditions["lines_calculated"]
print(s_ditfft.conditions["lines_calculated"], "lines calculated")
```

image

Black is current default ; red would be the new default 
New default is 4x faster on my laptop (1 CPU)

DIT+FFT profiler :
 <details>
<summary>   spectrum_calculation      46.249s ████████████████</summary>
        check_line_databank             0.137s 
        reinitialize                    1.055s 
            copy_database                   1.044s 
            memory_usage_warning            0.010s 
            reset_population                0.000s 
            others                          0.001s 
        scaled_eq_linestrength          0.835s 
        calc_lineshift                  0.075s 
        calc_hwhm                       0.671s 
        generate_wavenumber_arrays      0.055s 
        calc_line_broadening            43.329s ██████████████
            precompute_DLM_lineshapes       2.932s █
            DLM_Initialized_vectors         0.000s 
            DLM_closest_matching_line       1.484s 
            DLM_Distribute_lines            0.858s 
            DLM_convolve                    37.883s █████████████
            others                          0.172s 
        calc_other_spectral_quan        0.070s 
        generate_spectrum_obj           0.012s 
        others                          0.185s 
</details>

DIT+Voigt (300cm$^\mathrm{-1}$ trunc) profiler :
<details>
<summary>
    spectrum_calculation      12.433s ████████████████</summary>
        check_line_databank             0.104s 
        reinitialize                    1.147s █
            copy_database                   1.144s █
            memory_usage_warning            0.003s 
            reset_population                0.000s 
        scaled_eq_linestrength          0.772s 
        calc_lineshift                  0.077s 
        calc_hwhm                       1.265s █
        generate_wavenumber_arrays      0.038s 
        calc_line_broadening            8.930s ███████████
            precompute_DLM_lineshapes       0.606s 
            DLM_Initialized_vectors         0.000s 
            DLM_closest_matching_line       1.453s █
            DLM_Distribute_lines            0.879s █
            DLM_convolve                    5.876s ███████
            others                          0.116s 
        calc_other_spectral_quan        0.071s 
        generate_spectrum_obj           0.012s 
        others                          0.135s 
</details>

7,107,089 lines calculated             (>>> 0.5  and   2e11 points*lines/s 🥇  ) 

---

### References
- Physics for default 50 cm-1: Discussed in the `features-advanced_lineshapes` channel (https://radis-radiation.slack.com/archives/C01JBV54G11/p1629636709000900)
- some on-going work on characterizing physical truncations : https://www.lmd.jussieu.fr/~mturbet/research_spectroscopy.html

anandxkumar:
### Description

This pull request adds the `predict_time()` method in SpectrumFactory to precompute the estimated time taken by Legacy>Voigt, LDM>Voigt, and LDM>FFT methods. 

Splits `broadening_max_width` into 2 parameters - `neighbour_lines` and `truncation`.

Sets DIT>Voigt as default method with a truncation=300 cm-1 and neighbour_lines = 0. 
For LBL>Voigt, truncation=Spectral_range with neighbour_lines = None

Fixes #340 

todo : 

- [x] user error message if broadening_max_width used; suggestion with correct truncation value


radis/lbl/factory.py
```diff
@@ -1542,6 +1575,77 @@ def _generate_wavenumber_arrays(self):
 
         return
 
+    def predict_time(self):
```
erwanp, line 1578: I would refactor this as a function of two parameters:
- line optimization : DIT /  LBL 
- broadening algorithm :  FFT / AVP ("Analytical Voigt Profile", i.e. Voigt)

then add a series of 2 nested if/else  ;   and use some NotImplementedError   for any line optimization / convolution   you don't know.

The function "predict_time" will return only one float ; the optimization  will be done elsewhere by calling it several times to compare the results

radis/test/test_examples.py
```diff
@@ -24,6 +24,7 @@ def test_script_execution(script):
     plt.ion()
     # run:
     runpy.run_path(script, init_globals=locals())
+    plt.close("all")
```
erwanp, line 27: 👍

erwanp (changes requested):

radis/lbl/factory.py
```diff
@@ -1542,6 +1575,77 @@ def _generate_wavenumber_arrays(self):
...
+        # Check self broadening is here
+        if not "Tdpsel" in list(df.keys()):
+            Tdpsel = None  # if None, voigt_broadening_HWHM uses df.Tdpair
+        else:
+            Tdpsel = df.Tdpsel
+
+        molar_mass = get_molar_mass(df)
+
+        # Calculate broadening FWHM
+        _, wl, wg = voigt_broadening_HWHM(
```
erwanp, line 1607: I'd remove all calculation from predict_time()  ; it makes things complicated. 

We can call it during the calculation ; once we know most of the parameters. If some parmaeters are not known when it is called ; just raise an error, or add a default value.  Let's not mix calculation & the "package" around, i.e. prediction and analysis of results. 

 For instance, do not re-init the w_axis grids .  You can call predict() during the calculation once you know all the parameters; for instance replacing the ["Calculating. X lines expect Y s".](https://github.com/radis/radis/blob/c18b53522c2ca1b6486d6f7828aa108f58691639/radis/lbl/broadening.py#L2225)

radis/lbl/factory.py
```diff
@@ -1542,6 +1588,82 @@ def _generate_wavenumber_arrays(self):
 
         return
 
+    def predict_time(self):
+        def _is_at_equilibrium():
```
anandxkumar, line 1592: @erwanp I was using the `_is_at_equilibrium` function in `predict_time()` to check whether its non_equilibrium and if optimization = "simple" or "min-RMS" then `x 2` the  calculation time because `_apply_lineshape_DLM` is called twice which is the bottleneck step. One thing I found if we use this function it won't work if using same Spectrum factory because overpopulation gets updated to `{}` if using non_eq_spectrum. Should we modify `self.input.overpopulation = None` when using `eq_spectrum`?

radis/lbl/calc.py
```diff
@@ -418,6 +420,15 @@ def _check_molecules_are_consistent(
...
                 molecule_dict[mol][argument_name] = argument_dict[mol]
 
+    # Checking if `auto` mode
+    if optimization == "auto":
+        optimization = "min-RMS"
+        broadening_max_width = 300
+        if (wavenum_max - wavenum_min) < 300:
+            broadening_method = "fft"
+        else:
+            broadening_method = "voigt"
```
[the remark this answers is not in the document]
anandxkumar: So in case if the user gives all values that are `broadening_max_width` and `truncation` we should simple use the default truncation value and default `neighbour_lines`?
Also what will be the value of `neighbour_lines` based on different `spectral range`, `optimization` and `broadening_method`?

erwanp:
Is the calculation mode set to DIT/Voigt by default?

anandxkumar:
Yes with a broadening_max_width =300 and a broadening_max_width = 10 for LBL.

examples/plot_exomol_spectrum.py
```diff
@@ -19,6 +19,7 @@
     Tgas=1000,  # K
     mole_fraction=0.1,
     path_length=1,  # cm
+    broadening_method="fft",  # @ dev: Doesn't work with 'voigt'
```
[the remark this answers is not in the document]
anandxkumar: `AttributeError: 'DataFrame' object has no attribute 'selbrd'`
Basically `voigt` uses `df.selbrd` to calculate the `voigt_broadening_HWHM`. Seems like the `exomol` data doesn't have this particular column.

radis/lbl/factory.py
```diff
@@ -529,38 +530,29 @@ def __init__(
             if optimization is None:
                 if self.verbose >= 3:
                     printg(
-                        "LDM algorithm not used. Defaulting truncation from {0} to 10".format(
+                        "LDM algorithm not used. Defaulting truncation from {0} to 0".format(
```
erwanp, line 533: I made an error here. Default truncation should be 300 cm-1. And default neighbor likes should be 0. Algorithm shouldn't matter (DIT /Voigt and LBL/Voigt have same truncation,  only DIT/FFT should raise an error if truncation as its not implemented yet)

anandxkumar: DIT/Voigt with 300 cm-1 truncation will be feasible, but I'm not sure about LBL/Voigt. As seen from the benchmarks before, LBL is independent of spectral range i.e. neighbour_lines, but is dependent on the truncation. And for a truncation=300cm-1, the calculation time and memory will be expensive. So a default 300 cm-1 may not be feasible.

erwanp: Good point for the performance drop. But at least the physics remain the same, so it's a good reason to have 300 cm-1 always  (similar to the remark of @dcmvdbekerom  in #340  : we don't want to change the physical output without the user knowing / deciding) 

Given DIT/Voigt is the default, it should be fine in most cases. 

And anyway, if the predict_time() is given before the broadening step, the user will know. 
We can add an extra PerformanceWarning if predicted calculation time exceeds a few minutes.

examples/plot_exomol_spectrum.py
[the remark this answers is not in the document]
erwanp: @minouHub you'll be interested

Normally, the other broadening (i.e. airbrd in HITRAN, may be different in ExoMol) is used when selbrd is missing.

Issue #245, opened 2021-04-14
"emissivity_noslit" convolved by the slit "on the go"


CorentinGrimaldi:
### 🎯 Describe what you'd like

The request is to have the possibility to get (and plot) the variable "emissivity_noslit" convoluted by the slit "on the go".

According to @erwanp , in the future Radis will only store the non convoluted variables and the slit function, and will calculate the convoluted variables "on the go" when asked by the user (through "get" or "plot" methods for example).
This could be the first step toward the future handling of the slit by Radis. 

### Why "emissivity" convoluted by the slit ?
After the abel inversion of a calibrated experimental spectrum (in W/(cm2.sr.nm) representing the power coming from the source received on the surface of the camera by nanometer, we obtain an abel inverted spectrum in W/(cm3.sr.nm).
This is to be compared to the emissivity of the source.

In radis the emissivity is not convoluted by the slit, but the abel inverted spectrum has a slit applied by (re)construction. 

At this point in Radis, to compare the abel inverted spectrum to a simulation, we can ask for a spectrum convoluted by the slit (in W/(cm2.sr.nm) for a path length of 1 cm, thus virtually having the result in W/(cm3.sr.nm).
However we can not plot the 2 spectra on the same figure without using the 'force' mode...
And the method to get what we want is a bit tricky...

### 💡 Possible implementations

---
<sub>*Tip : you can vote for [New Features](https://github.com/radis/radis/issues/79) on https://feathub.com/radis/radis*</sub>

erwanp:
Note that as early as this morning, some users using RADIS for heat transfer also wanted to be able to use a rectangular slit to create a moving-average of a high-resolution spectrum, and generate a narrow-band-model spectrum. This would be very useful for them too !!

I'm labelling it as a Good First Issue, as most methods/functions are already implemented, and should just be properly re-ordered.

minouHub:
Moving to "long-term" milestone as this issue wasn't active for a long time

erwanp:
I'm even closing; we'll re-open if needed

PR #344, opened 2021-08-12
Refactor : remove separation of convolved / nonconvolved arrays in Spectrum

erwanp:
### Description

- **major (not visible) internal change of Spectrum** : remove ._q_conv attribute of Spectrum : all spectral arrays are stored together in Spectrum._q
Consequene for the user : (change in interface) arrays convolved with apply_slit keep the same length but have np.nans on the side in 'valid' mode
- new apply_slit(inplace=False) mode which only returns convolved arrays in a new spectrum
- add a method s.trim()  to remove nans on the side ; and a "trim_nan" parameter in s.get()
- plot_diff ; get_residual ; resample   updated to work as normally. **You shouldn't see any change in your user code**.
- add tests for resampling
- add convenience function to generate a test spectrum: 

So this piece of code is valid:
```
s = radis.test_spectrum()
s.apply_slit(1.5, "nm")
s.has_nan() 
>>> True
s2 = s.apply_slit(1.5, "nm", inplace=False).trim()
s2.has_nan()
>>> False
```

And also : 
- new simpler Spectrum creation format  : 
```
Spectrum({"wavelength":w, "radiance":I})
```
- new tests of spectrum creation
- add warning if interpolating low-res spectrum over a high-res

Will help implement #245 very easily

@CorentinGrimaldi @BlehMaks @minouHub this will prevent all the "if_convoluted_drop" errors you may have encountered in the past , when using "offset" ; "resample"; "crop" or "get_residual"


radis/spectrum/spectrum.py
```diff
@@ -2292,55 +2395,92 @@ def apply_slit(
...
+                I_conv = np.hstack(I_conv_slices[q])
+
+                # Store
+                self._q[q] = I_conv
+
+                # Get units
+                self.units[q] = new_units[q]
+            s_out = self
+        else:
+            w_conv = np.hstack(w_conv_slices)
```
CorentinGrimaldi, line 2447: this is done wther inplace = True or False

radis/spectrum/spectrum.py
```diff
@@ -2292,55 +2395,92 @@ def apply_slit(
...
+        if inplace:
+            w_conv = np.hstack(w_conv_slices)
+            if len(self._q["wavespace"]) != len(w_conv) or not np.allclose(
+                self._q["wavespace"], w_conv
+            ):
+                raise AssertionError(
+                    "Wavespace of convolved arrays is different, cannot store it in the same Spectrum. You can use Spectrum.apply_slit(inplace=False) to return a new spectrum with only the convolved arrays"
+                )
+            for q in I_conv_slices.keys():
+                qns = q + "_noslit"
```
CorentinGrimaldi, line 2435: useless now ?

radis/spectrum/spectrum.py
```diff
@@ -2292,55 +2395,92 @@ def apply_slit(
...
+            # make new spectrum
+            quantities = I_conv_slices.items()
+            quantities = {
+                "wavespace": w_conv
+            }  # a copy will be created in Spectrum creation
+
+            for q in I_conv_slices.keys():
+
+                # Merge all slices
+                quantities[q] = np.hstack(I_conv_slices[q])
```
CorentinGrimaldi, line 2457: this part too

CorentinGrimaldi (changes requested):
First of all, very nicely done!
A few things though:
- The simpler/quick creation of Spectrum is a bit dangerous because the units of the arrays are not specified, especially for wavelength, that could be a huge source of errors. I think a mandatory argument "wav_unit" (and maybe "I_unit" should be asked
- There are still some "_noslit" traces in the names of variables, which is irrelevant now (and false when you apply a slit with `inplace=True`)
- The fact now that the slit convoluted version of a calculated spectrum replace the original spectrum by default can be problematic because it is irreversible. I think the default behavior should be `inplace=False`
- Apply_slit() can now be applied multiple time and give nonsense (with inplace=True), this should not be allowed, for exemple with a `slit_convoluted` flag
- Saving a slit convoluted spectrum is often less interesting than the original spectrum, a warning could be print to explain to the user that saving the original spectrum is a better practice.

These last comments will be solved with the apply_slit on the go feature #245

radis/spectrum/spectrum.py
erwanp: No, this one is where we do the stacking. 
Instead I remove 
```
            quantities = I_conv_slices.items()
```
from the line above

erwanp:
All done ! 

> * The simpler/quick creation of Spectrum is a bit dangerous because the units of the arrays are not specified, especially for wavelength, that could be a huge source of errors. I think a mandatory argument "wav_unit" (and maybe "I_unit" should be asked

Done; wunit is required. Added tests.

> * There are still some "_noslit" traces in the names of variables, which is irrelevant now (and false when you apply a slit with `inplace=True`)

It's still relevant for the moment : if you do apply_slit inplace for `radiance_noslit`; it creates a new variable `radiance` ; it does not replace the existing `radiance_noslit`. 

> * The fact now that the slit convoluted version of a calculated spectrum replace the original spectrum by default can be problematic because it is irreversible. I think the default behavior should be `inplace=False`

Again, it does not. 

> * Apply_slit() can now be applied multiple time and give nonsense (with inplace=True), this should not be allowed, for exemple with a `slit_convoluted` flag

same

> * Saving a slit convoluted spectrum is often less interesting than the original spectrum, a warning could be print to explain to the user that saving the original spectrum is a better practice.

PR #348, opened 2021-08-18
Add/config file with plotstyles and global parameters

erwanp:
### Description

This PR brings more custom parameters, and finalizes the radis.json user config file  : 
- ~/radis.json updates the default parameters of /radis/radis/config.json    (similar to how config parmateers are done in other softwares : one default .json file ; one user .json file ) 
- remove all the global variables from params.py , add them in config.json
- choose your plotting style in your ~/radis.json

Default :
```
"plot":{
    "plotlib":"publib",
    "context":"paper",    # or "poster", "talk"
    "style":"origin"
    },
```
Or : 
```
"plot":{
    "plotlib":"seaborn",
    "context":"paper",
    "style":"darkgrid"       # or any other seaborn theme
    },
```
Or

```
"plot":{
    "plotlib":"none"         # to keep your current theme & style
    },
```

Publib 
image

Seaborn 
image

Todo : 
- [x] add None style 
- [x] add doc 

On a new system, after a first import: 

```
In [1]: import radis
RADIS configuration file created in C:\Users\erwan\radis.json

```

The radis.json looks like this : 
```
{
"database":{},
"spectroscopic_constants":{
    "CO2": "molecules_data.json",
    "CO": "molecules_data.json"
    },
"plot":{
    "plotlib":"publib",
    "context":"paper",
    "style":"origin"
    },
"DEBUG_MODE":false,
"AUTO_UPDATE_SPEC":false,
"OLDEST_COMPATIBLE_VERSION":"0.9.1",
"USE_CYTHON":true,
"GRIDPOINTS_PER_LINEWIDTH_WARN_THRESHOLD":3,
"GRIDPOINTS_PER_LINEWIDTH_ERROR_THRESHOLD":1
}

```


examples/plot_SpecDatabase.py
```diff
@@ -37,7 +37,7 @@
 s1 = sf.eq_spectrum(Tgas=300, path_length=1)
 
 # Creating SpecDatabase
-my_folder = os.getcwd() + "/SpecDatabase_Test"
+my_folder = Path.cwd() / "SpecDatabase_Test"
```
anandxkumar, line 40: I have updated it in my PR using `os.path.dirname(os.path.realpath(__file__)) + "/SpecDatabase_Test"` and also I'm deleting the SpecDatabase folder using `shutil` so that no issue is created locally. So you can undo the changes here to avoid any conflict or maybe the other way around?

anandxkumar:
The goals of JSON config file seem to be accomplished (Ability to add comments still remaining).  I fetched the repo and update my `radis.json` file, even on switching the `plotlib:` to `seaborn`, I was getting the same plot as `publib`, possible reason, seaborn is installed on my system?
Also, we will need to add seaborn in `setup.py`?

erwanp:
Did you add `seaborn` in your ~/radis.json file? Did you restart the console and imports? 

You're right for seaborn in setup.py, will do!

anandxkumar:
I deleted the radis.json and let the default script create the default `radis.json`; for the following radis.json file:
```
{
"database":{},
"spectroscopic_constants":{
    "CO2": "molecules_data.json",
    "CO": "molecules_data.json"
    },
"plot":{
    "plotlib":"seaborn"
    },
"DEBUG_MODE":false,
"AUTO_UPDATE_SPEC":false,
"OLDEST_COMPATIBLE_VERSION":"0.9.1",
"USE_CYTHON":true,
"GRIDPOINTS_PER_LINEWIDTH_WARN_THRESHOLD":3,
"GRIDPOINTS_PER_LINEWIDTH_ERROR_THRESHOLD":1
}
```

I was getting the following output:
FigureX_1

erwanp:
Indeed, the default values were overriding the user ones instead of the reverse. Fixed now !

erwanp:
@anandxkumar , tests pass, you can merge if you approve the changes!

anandxkumar:
Nice! Working now. Merging!

anandxkumar:
@erwanp, one issue, the default settings (except database) don't get appended when `radis.json` exists. So it will cause issue for the user who already have a `radis.json` file. What we can do is run a check, that is if `radis.json` exists but other global parameters don't exist then append those values in the `radis.json` file.

erwanp:
If the keys don't exist the default ones from defauult_radis.json should be used. Doesn't this work?

anandxkumar:
It does use the default values. The `radis.json` file doesn't get updated with the remaining Global Parameters, so if the user wants to update them, they will either have to manually copy them in the `radis.json` file or edit the `default_radis.json` file. So if we load the default parameters in the existing `radis.json` file, it will be easier for the user to update the settings.

erwanp:
The expected behavior is to copy them in the `~radis.json`  . Editing an existing file can be dangerous (what if we overwrite an existing key?), I think it's safer / faster to simply add a message in the documentation saying how to copy from "default_radis.json" (with link to GitHub's file ? ) into their user file.

anandxkumar:
Maybe better check if the global parameters (all keys) exist or not in `~/radis.json`, and if not then straight add that info in it from the `default_radis.json`, in that way we will not overwrite any existing key and also the other global_parameters/keys will get added.

erwanp:
That could be a project, but requires to append to an existing file. We currently do it to add DatabankEntries, so it's doable, but requires to be implemented with caution and therefore I don't want to spend this time right now ;) 

Current solution will do. Unless I understood something wrong, it's just about copy-pasting some text from one file to the other.

Issue #350, opened 2021-08-21
HITRAN connection error in CI-test : replace test database 


erwanp:
### 🐛 Describe the bug

test_eq_vs_noneq_isotope()  often fails on Travis CI  because of connection errors with the HITRAN server. 

```

        sf = SpectrumFactory(
            wavelength_min=4250,
            wavelength_max=4350,
            mole_fraction=1,
            path_length=1,
            cutoff=1e-25,
            molecule="CO2",
            isotope="1,2",
            verbose=verbose,
        )
        sf.warnings["MissingSelfBroadeningWarning"] = "ignore"
        sf.warnings["NegativeEnergiesWarning"] = "ignore"
        sf.warnings["HighTemperatureWarning"] = "ignore"
        sf.fetch_databank(
>           "hitran"
        ) 
```

Sometimes returns : 

```

            # sending a valid response.
>           raise RemoteDisconnected("Remote end closed connection without"
                                     " response")
E           urllib3.exceptions.ProtocolError: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))
```

Latest occurence : https://app.travis-ci.com/github/radis/radis/jobs/532986942#L3792 

### 💡 Possible solutions

We do not need to download from a remote server for this. Could replace with one of the test databases already shipped with RADIS : see test/files   (just make sure it has non-eq : i.e. CO or CO2 ) and multiple isotopes on the given range.

anandxkumar:
Yes, totally agree with it :+1:, a local test database would be a lot better for this.

erwanp:
Same connection error in latest commit bbd27b5   https://app.travis-ci.com/github/radis/radis/jobs/533559080#L3802

minouHub:
What is needed in my opinion:

- [ ] at least **a single** molecule for **each** molecular database (HITRAN, HITEMP, ExoMol, GEISA) and atomic database (Kurucz, NIST) 
- [ ] local test database for the other molecules

This is partially done for HITEMP in #739

PR #363, opened 2021-08-28
Update info messages & docs - Fixes: #351 

### LINKED ISSUE #351: missing value in Default() info message
### 🐛 Describe the bug

Tiny bug in print, verbose=3. It shows" Default" instead of the value. :
```
LDM algorithm used. Defaulting broadening method from <radis.misc.utils.Default object at 0x0000018C126EA070> to Voigt
```

### 💡 Possible solutions

Could be corrected by setting __repr__ for the Default class

anandxkumar:
Fixes #351 and shifts `predict_time()` to `broaden_lines` and `broaden_lines_noneq`

### DIFF
--- a/radis/lbl/factory.py
+++ b/radis/lbl/factory.py
@@ -97,7 +97,6 @@ except ImportError:  # if ran from here
     from radis.lbl.base import get_waverange
 
 from radis.misc.basics import flatten, is_float, list_if_float, round_off
-from radis.misc.printer import printg
 from radis.misc.utils import Default
 from radis.phys.constants import k_b
 from radis.phys.convert import conv2
@@ -390,7 +389,7 @@ class SpectrumFactory(BandFactory):
         optimization="simple",
         folding_thresh=1e-6,
         zero_padding=-1,
-        broadening_method=Default("voigt"),
+        broadening_method="voigt",
         cutoff=0,
         parsum_mode="full summation",
         verbose=True,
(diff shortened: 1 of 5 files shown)


radis/lbl/factory.py
```diff
@@ -528,15 +528,18 @@ def __init__(
...
         if isinstance(broadening_method, Default):
-            if optimization in ("simple", "min-RMS") and broadening_method != "voigt":
+            if (
+                optimization in ("simple", "min-RMS")
+                and broadening_method.value != "voigt"
+            ):
                 if self.verbose >= 3:
                     printg(
-                        "LDM algorithm used. Defaulting broadening method from {0} to Voigt".format(
+                        "LDM algorithm used. Defaulting broadening method from {0} to 'voigt'".format(
```
erwanp, line 537: I think we can simplify the code : in both cases (DIT/not DIT) we use "voigt" anyway. Why should we even compare / print a message ? 

All of this block could be reduced to 
```
if isinstance(broadening_method, Default):
    broadening_method = broadening_method.value
```
and maybe we could even avoid using Default ?

radis/lbl/factory.py
```diff
@@ -1716,7 +1719,7 @@ def _is_at_equilibrium():
                 raise NotImplementedError("broadening_method not implemented")
         elif optimization is None:
             if broadening_method == "voigt":
-                estimated_time = 6.6487e-08 * n_lines * truncation / wstep * factor
+                estimated_time = 6.6487e-08 * n_lines * truncation / wstep
```
erwanp, line 1722: is "factor" from L1677 still used anywhere else?

erwanp (changes requested):

radis/lbl/factory.py
anandxkumar: No. It is being used by optimization =`simple` and `min-RMS`. For non-eq cases factor = 2 so we can simply refactor it as :
```
if not  _is_at_equilibrium:
   factor = 2
```
and remove rest of the code

radis/lbl/factory.py
anandxkumar: Yes, Default was used for the purpose that we had different `broadening_methods` for DIT and LBL, but now we are using `voigt` as default so there is no need for `Default`.

erwanp (commented):
Last thing : in def predict_time(self): docstrings, can you mentino what this function is doing ; and also reference some of your graphs (ideally the Tableau ones were you derived the predict function ), and your https://anandxkumar.github.io/Benchmark_Visualization_GSoC_2021  project ? 

Will be very helpful in the long run. 

You should also copy a similar text in the docs under [Performance](https://radis.readthedocs.io/en/latest/lbl/lbl.html#precompute-spectra) ; for instance create a section "Benchmark". We want people to be able to see your graphs !

anandxkumar:
You can merge, the test has passed along linting! (Travis had connection issue as reported before)

erwanp (approved):
All good, I let you merge!

erwanp:
Test failure is yet again a connection error with `test_eq_vs_noneq_isotope`.      see #350
You can ignore;

PR #369, opened 2021-09-08
 Fix of get_baseline() function of Spectrum class #368 

### LINKED ISSUE #368: Bug of get_baseline() function of Spectrum class
### 🐛 Describe the bug
1) radis.spectrum.operations.get_baseline() function has a `peakutils` module dependency which is not included in RADIS dependencies.  

2) There is a missing parameter in radis.spectrum.operations.get_baseline() function call. There also should be a wunit parameter for `Spectrum.from_array()` correct work. 

### 💡 Possible solutions
1) The first can be solved  by: `conda install -c conda-forge peakutils`, suggested by @erwanp

2) The second can be solved by adding wunit parameter to the line 1049 and 1076 of 
```
1049: def get_baseline(s, var="radiance", wunit="nm", Iunit=None):
1076:         w1, baseline, var, wunit=wunit, unit=Iunit, name=s.get_name() + "_baseline"
```

BlehMaks:
This pull request fixes the second point of the issue #368 on missing wunit parameter in get_baseline() function of Spectrum class.

Fixes #368

### DIFF
--- a/radis/spectrum/operations.py
+++ b/radis/spectrum/operations.py
@@ -1046,32 +1046,67 @@ def offset(s, offset, unit, name=None, inplace=False):
     return s
 
 
-def get_baseline(s, var="radiance", wunit="nm", Iunit=None):
+def get_baseline(
+    s, var="radiance", algorithm="als", wunit="default", Iunit="default", **kwargs
+):
     """Calculate and returns a baseline
 
     Parameters
     ----------
     s: Spectrum
         Spectrum which needs a baseline
-
     var: str
         on which spectral quantity to read the baseline. Default ``'radiance'``.
         See :py:data:`~radis.spectrum.utils.SPECTRAL_QUANTITIES`
+    algorithm: 'als', 'polynomial'
+        Asymmetric least square or Polynomial fit
+    **kwargs: dict
+        additional parameters to send to the algorithm. By default,
+        for 'polynomial':
+
+            **kwargs = **{"deg": 1, "max_it":500}
+
+        for 'als':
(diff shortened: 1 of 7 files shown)


erwanp:
Hello! You can fix the first by adding peakutils in the environment.yml file. 
If peakutils is also on Pypi or on Github, it can also be added in setup.py

erwanp:
I think get_baseline was not used in the tests, else we would have spotted this bug before. You could :
- quick fix : add a call to get_baseline within one of the test functions in test/spectrum/test_*.py
- better but takes a bit more time : do the same, and check that the result matches what's expected. For instance check intensity value of both ends of the baseline spectrum ?

radis/spectrum/operations.py
```diff
@@ -1046,7 +1046,7 @@ def offset(s, offset, unit, name=None, inplace=False):
     return s
 
 
-def get_baseline(s, var="radiance", Iunit=None):
+def get_baseline(s, var="radiance", wunit="nm", Iunit=None):
```
erwanp, line 1049: btw, here I think the default for wunit should be `"default"`, and if `"default"` overwrite with the Spectrum default waveunit 
```
s.get_waveunit()
```

erwanp (changes requested):

erwanp:
Fixed, and added a new get_baseline() function based on asymmetric least squares

```python
sb = s_exp.get_baseline(algorithm="als")
s_exp.plot()
sb.plot(nfig="same", lw=3)
```

image

Issue #321, opened 2021-07-21
refactor : MolParams


erwanp:
### 🎯 Todo

MolParams (molecular parameters) is currently reinitialized at several places through a calculation. A same object should be used. 
Will also make it possible to use #301 

### 💡 Possible implementations

- Store as an attribute of SpectrumFactory ? 

---
<sub>*Tip : you can vote for [New Features](https://github.com/radis/radis/issues/79) on https://feathub.com/radis/radis*</sub>

PR #374, opened 2021-09-25
hard coded isotopologue name and mass of CN

minouHub:
### Description
This pull request is to address the spectrum of CN which listed in the "known molecules" of RADIS.

After the merge, the following code should be working for you:

```python

T = 3000
s = calc_spectrum(
    wavelength_min=700,     #nm
    wavelength_max=1300,    #nm
    molecule="CN",
    isotope="1",
    pressure=0.1,  # bar
    Tgas=T,  # K
    mole_fraction=0.01,
    path_length=1,  # cm
    broadening_method="fft",  # @ dev: Doesn't work with 'voigt'
    databank="exomol", 
)
s.plot("absorbance", wunit='nm')
```

### Long term
@erwanp mentioned that the mass of molecules could be directly retrieved from ExoMol definition file, but this issue #321 is closed. I am not sure that #342 fixed it completely.

### DIFF
--- a/radis/db/classes.py
+++ b/radis/db/classes.py
@@ -281,6 +281,7 @@ def get_molecule(molecule_id):
 KNOWN_EXOMOL_ISOTOPES_NAMES = {
     ("FeH", 1): "56Fe-1H",
     ("SiO", 1): "28Si-16O",  # Placeholder until all molecules parsed from website TODO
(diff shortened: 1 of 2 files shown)


erwanp:
Tests pass, only Black-linting fails. I suggest you [install pre-commit](https://radis.readthedocs.io/en/latest/dev/developer.html#code-linting) so that the linting is always done automatically on commit

PR #390, opened 2021-11-04
Improve SpecDatabase (bug correction and update)

CorentinGrimaldi:
### Description

This pull request is to address several bugs using specdatabase features. 

- Using a spectrumfactory connected to a specdatabase, the first spectrum was recalculated even if present in the database;
- Some parameters were not correctly parsed and checked due to string-boolean errors. (`"0.0"` or `"False"` (if stored as text) are evaluated as `True` in Python  )

Additionaly, a convenient option was implemented: 

- `update_register_only` to regenerate ths .csv database file without keeping the spectra in memeory.

### DIFF
--- a/radis/tools/database.py
+++ b/radis/tools/database.py
@@ -1997,7 +1997,7 @@ class SpecDatabase(SpecList):
             if ``True``, reloads files already in database. Default ``False``
         filt: str
             only consider files ending with ``filt``. Default ``.spec``
-        
+
         Other Parameters
         ----------------
         update_register_only: bool
@@ -2392,7 +2392,7 @@ class SpecDatabase(SpecList):
         update_register_only: bool
             if ``True``, load files and update csv but do not keep the Spectrum in memory.
             Default ``False``
-            
+
         Returns
         -------
         db: pandas.DataFrame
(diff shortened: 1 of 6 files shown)


radis/tools/database.py
```diff
@@ -2497,9 +2531,34 @@ def read_conditions_file(path):
...
+        if boolvar in df:
+            if df.dtypes[boolvar] == object:
+                from radis.misc.basics import str2bool
+
+                if verbose:
+                    print(
+                        f"Reading {path} : casting column {boolvar} from string type to boolean"
+                    )
+                df[boolvar] = df[boolvar].map(str2bool)
+
```
erwanp, line 2561: Works, but limited to columns we already know of. I recently had a similar problem ("0.0" detected as True) after declaring a new column name. 
There must be a better way. 
I found this : 

`read_csv` directly has a `true_values` and `false_values` optional list arguments. We could replace with 
```
pd.read_csv(path, float_precision="round_trip", true_values=["true", "True", "1.0", "1"], false_values=["false", "False", "0.0", "0"])
```

erwanp: @CorentinGrimaldi if you can try ?

erwanp:
Merging, i'll let you update in the next version

PR #395, opened 2021-11-12
Dimensioned Spectrum.max() function

erwanp:
.max now returns dimensioned quantity. Allows : 

```
s /= s.max()
```

as another way to normalize a spectrum

(and makes it much easier to multiply / subtract a Spectrum based on min/max values from another Spectrum, without having to worry about units ! )


erwanp:
@CorentinGrimaldi you can approve & merge directly

PR #397, opened 2021-11-13
Fixes #356 

### LINKED ISSUE #356: fetch_hitemp fails if isotope=int
### 🖥 Steps To Reproduce

```python
from radis.io.hitemp import fetch_hitemp
fetch_hitemp("OH", isotope=1)
```

Fails with :  
>     where.append(f'iso in {isotope.split(",")}')
> AttributeError: 'int' object has no attribute 'split'

### 💡 Possible solutions

For the moment, one can still use 
```python
fetch_hitemp("OH", isotope="1")
```
But `int` should be supported out of the box

anandxkumar:
### Description

This pull request converts `int` value isotope into `string`  for `fetch_hitemp`.

Fixes #356

### DIFF
--- a/radis/io/hitemp.py
+++ b/radis/io/hitemp.py
@@ -447,7 +447,7 @@ def fetch_hitemp(
     databank_name: str
         name of the databank in RADIS :ref:`Configuration file <label_lbl_config_file>`
         Default ``"HITEMP-{molecule}"``
-    isotope: str
+    isotope: str, int or None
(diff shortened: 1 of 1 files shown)


anandxkumar:
Should we add `list` support for isotopes like `isotope=[1,2,3]`?

erwanp (approved):
All good, you can merge!
I think list is not needed ; unless users end up asking for it !

PR #410, opened 2021-11-28
fix isotope='all' syntax when fetching HITRAN files

erwanp:
### Description

Will fix https://github.com/suzil/radis-app/issues/71  when 0.11.2 is released


suzil (approved):
great!

PR #386, opened 2021-10-24
Add/sparse waverange optim

erwanp:
### Description

Implement an major optimization to improve performances when the waverange grid is sparse, i.e., lines or groups of lines are very separated from one another   (related to #222 ) 

Optimization is so far only implemetned in DIT/Voigt mode, but could be easily added in DIT/FFT too   @dcmvdbekerom 

Basically, we mask the waverange to remove regions one truncatino away from any line, before applying the convolution. Note that in the current implementation each (wG, wL) point in the DIT grid can have its own truncation; which is something we want to keep in the future  (if we end up having truncations that are a multiple of (wG, wL) instead of having a common value. 

```

import astropy.units as u
from radis import SpectrumFactory, config, plot_diff

print("Without Sparse Range Optimization")
print("=================================")
print("\n"*2)

config["SPARSE_WAVERANGE"] = False
sf = SpectrumFactory(1, 3000, # 26777,         # cm-1
                  molecule='NO',
                  isotope='1,2,3',
                  pressure=760 * u.torr,
                  verbose=False,
                  wstep=0.01,
                  mole_fraction=1)

sf.fetch_databank('hitemp')
s = sf.eq_spectrum(700)
s.print_perf_profile()


print("With Sparse Range Optimization")
print("==============================")
print("\n"*2)

config["SPARSE_WAVERANGE"] = True
s2 = sf.eq_spectrum(700)
s2.print_perf_profile()

# Plot
s.name = f'0.10.3 ({s.conditions["calculation_time"]:.0f}s)'
s2.name = f'Sparse optim ({s2.conditions["calculation_time"]:.0f}s)'
_, [ax0, ax1] = plot_diff(s, s2)
ax0.set_yscale('log')
```

Output : 

```
Without Sparse Range Optimization
=================================
NOspectrum1533317963200 profiler :
    spectrum_calculation      22.242s ████████████████
        check_line_databank             0.001s 
        reinitialize                    0.007s 
            copy_database                   0.005s 
            memory_usage_warning            0.002s 
            reset_population                0.000s 
        scaled_eq_linestrength          0.026s 
        calc_lineshift                  0.003s 
        calc_hwhm                       0.058s 
        generate_wavenumber_arrays      0.001s 
        calc_line_broadening            22.100s ███████████████
            precompute_DLM_lineshapes       1.070s 
            DLM_Initialized_vectors         0.000s 
            DLM_closest_matching_line       0.042s 
            DLM_Distribute_lines            0.540s 
[...]

PR #416, opened 2021-12-19
major error in sparse algorithm if lines were not sorted

erwanp:
Sparse algorithm introduced in #386 , #388  [Radis 0.11](https://github.com/radis/radis/pull/400)  introduced an error in cases where the line database was not sorted in wavenumber. 

This happened in real-life scenario when : 
- downloading from HITRAN , where each isotope is downloaded one after another 
- and large spectral ranges where the sparse-mode is activated 
Typically, CO spectra. 

- or computing from ExoMol with multiple isotopes. 

No problems for HITEMP  calculations (all isotopes are stored together and already sorted). 

Fix : 

- [x] assert database is always sorted at runtime
- [x] sort lines after fetch_databank
- [x] sort lines when downloading HITRAN with vaex

Lines missing in 0.11, fixed in 0.12 : 

image

### DIFF
--- a/radis/io/hdf5.py
+++ b/radis/io/hdf5.py
@@ -171,7 +171,7 @@ class HDF5Manager(object):
             raise NotImplementedError(self.engine)
             # h5py is not designed to write Pandas DataFrames
 
-    def combine_temp_batch_files(self, file, key="default"):
+    def combine_temp_batch_files(self, file, key="default", sort_values=None):
         """Combine all batch files in ``self._temp_batch_files`` into one.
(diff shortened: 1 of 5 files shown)


dcmvdbekerom:
Nice fix! Is re-sorting at runtime due to unequal pressure shifts not dragging down performance too much?

erwanp:
I haven't compared head-to-head but we'll see in the profiles if it becomes significant.

We could improve it by having pandas use a mergesort / insertion sort, which perform best for almost sorted arrays

Issue #418, opened 2021-12-20
GPU examples not clickable


erwanp:
### 🐛 Describe the bug

Small issue introduced post 0.12 (see also  #417  )

- [ ] GPU examples have been generated (and ran), [but are not clickable on ReadTheDocs](https://radis.readthedocs.io/en/latest/auto_examples/index.html) : I think it's because they're missing a title in the header. See other examples in examples/ 

![image](https://user-images.githubusercontent.com/16088743/146754341-a2f34efc-32eb-4fd6-85f0-7dbf50223ac9.png)

@dcmvdbekerom if you wanna add a title

erwanp:
Minor comment : some formatting is broken in the README https://github.com/radis/radis#gpu-acceleration

PR #419, opened 2022-01-03
Gpu widget bug fix

dcmvdbekerom:
Fixes https://github.com/radis/radis/issues/418

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -97,7 +97,7 @@ Refer to the `Examples <https://radis.readthedocs.io/en/latest/examples/examples
 GPU Acceleration
 ----------------
 
-RADIS supports GPU acceleration for super-fast computation of spectra:
+RADIS supports GPU acceleration for super-fast computation of spectra::
 
     from radis import SpectrumFactory
     from radis.tools.plot_tools import ParamRange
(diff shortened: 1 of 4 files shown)


erwanp:
Test failure is unrelated 

https://app.travis-ci.com/github/radis/radis/jobs/554024187

>     #cython: language_level=3
> E   ValueError: numpy.ufunc size changed, may indicate binary incompatibility. Expected 232 from C header, got 216 from PyObject

Seems to be related to a Numpy version https://stackoverflow.com/questions/60323366/valueerror-numpy-ufunc-size-changed-may-indicate-binary-incompatibility-expec



We could ensure that Numpy version is always compatible, but this sounds like a lot of maintenance work eventually.  We could also decide to escape this error and switch back to numpy add_at . Would be done by adding `except (ValueError, ModuleNotFoundError)`  in [arrays.py L477](https://github.com/radis/radis/blob/develop/radis/misc/arrays.py#L477)

dcmvdbekerom:
I escaped it for now, didn't we have a flag that reports which add_at is used? It would have to be set in the code block you highlighted. I'm worried that this will result in always using numpy's add at. Did we ever try jit compilation?

erwanp:
It's in the Spectrum properties. "add_at_used". 

We could add a "check_performance" function that ensures that everything is properly activated, and return tips on how to improve performances.

erwanp (approved):
can merge!

Issue #211, opened 2021-03-11
Fail safely for lines without identifiers (ex : Error computing HITEMP H2O )


erwanp:
*Error reported by Clément Mirat at EM2C Laboratory*. 

### 🐛 Describe the bug

```
from radis import calc_spectrum
s = calc_spectrum(11000, 30000,
                  molecule='H2O',
                  isotope='1,2,3',
                  pressure=1.01325,   # bar
                  Tgas=2000,           # K
                  mole_fraction=0.1,
                  path_length=1,      # cm
                  databank='MY-HITEMP-H2O',
                  )
s.plot('emissivity_noslit')
```

with
```
[MY-HITEMP-H2O]
path = /Users/miratc/Documents/Applications/HITEMP_database/H2O/hitemp_*
format = hitran
parfuncfmt = hapi
```

Raises the error: 

```
Traceback (most recent call last):
  File “pandas/_libs/lib.pyx”, line 2062, in pandas._libs.lib.maybe_convert_numeric
ValueError: Unable to parse string ”   ”
```

### 💡 Possible solutions

The error is due to **empty lines** for the local spectroscopic parameters within HITEMP (`locu`)  

Exemple : line 600022 of [01_11000-30000_HITEMP2010.zip](https://hitran.org/hitemp/data/HITEMP-2010/H2O_line_list/) 

This fails the [parse_local_quanta](https://github.com/radis/radis/blob/develop/radis/io/hitran.py#L219) of RADIS HITRAN parser. 

Solutions : 

- label all lines in the HITEMP database (not within our control)  
- as a workaround, **fail safely if database cannot be parsed properly**. This will prevent nonequilibrium calculations, but still allow equilibrium calculations. 

Possible implementation:  

- [ ] Add a Try/except on the parse_local_quanta above ?

erwanp:
Currently circumvented by commenting the [parse_local_quanta](https://github.com/radis/radis/blob/develop/radis/io/hitran.py#L219]) and [parse_global_quanta](https://github.com/radis/radis/blob/develop/radis/io/hitran.py#L222) lines directly in io/hitran.py. Works for equilibrium calculations only !

PR #420, opened 2022-02-14
fix HCN import 

minouHub:
Hello,
The following was not working  before:

```python
s = calc_spectrum(500, 2300,         # cm-1
                  molecule='HCN',
                  isotope='1',
                  pressure=1.01325,   # bar
                  Tgas=300,
                  mole_fraction=0.1,
                  )
s.plot('absorbance')
```

I just extended the cases where the vibrational numbers are replaced by "-1". Let me know if that's ok.

Related to the kind of error seen in https://github.com/radis/radis/issues/211

Happy valentine's day <3

### DIFF
--- a/radis/io/hitran.py
+++ b/radis/io/hitran.py
@@ -94,7 +94,7 @@ def cast_to_int64_with_missing_values(dg, keys):
     """replace missing values of int64 columns with -1"""
     for c in keys:
         if dg.dtypes[c] != int64:
-            dg.loc[dg[c] == "  "] = -1  # replace empty cells by -1, e.g. HCN
+            dg.loc[dg[c] == "  ", c] = -1  # replace empty cells by -1, e.g. HCN
             dg[c] = dg[c].fillna(-1).astype(int64)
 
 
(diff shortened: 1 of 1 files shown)


radis/io/hitran.py
```diff
@@ -94,6 +94,7 @@ def cast_to_int64_with_missing_values(dg, keys):
     """replace missing values of int64 columns with -1"""
     for c in keys:
         if dg.dtypes[c] != int64:
+            dg.loc[dg[c] == "  "] = -1  # replace empty cells by -1, e.g. HCN
```
erwanp, line 97: I think you're replacing for all columns while you should only be replacing the "C" one here. Right syntax is dg.loc[dg[C]=="  ", C] 

Or you can use replace : https://stackoverflow.com/a/52851249

erwanp (changes requested):

radis/io/hitran.py
minouHub: alright, done

minouHub: done  :)

erwanp:
Tests pass except from an connection error to HITRAN sefvers, which is not related. @minouHub  you can merge !

minouHub:
Thanks @erwanp and @Supriya1702 for your work in https://github.com/radis/radis/issues/423!

Issue #275, opened 2021-06-01
Add Automatic tests (CI) on Windows


erwanp:
### 💡 Possible implementations

---
<sub>*Tip : you can vote for [New Features](https://github.com/radis/radis/issues/79) on https://feathub.com/radis/radis*</sub>

arunavabasu-03:
i would like work on this issue

erwanp:
Great, adding you here !

minouHub:
Done as of 0.16.4, see #873

PR #425, opened 2022-03-14
windows test support #275

arunavabasu-03:
## **Automate windows test** 
👉 python is not supported in travis;  so as a language I use `generic `and chocolaty to install python on the runner and the run tests, we can use references and anchors to gracefully compose our stages.

Hey @erwanp, please check that this approach is right or we have to go with another approach. 😀


anandxkumar:
Hi @arunavabasu-03, great work. The Travis build has failed: https://app.travis-ci.com/github/radis/radis/jobs/563238622.
I would suggest you to build travis locally before pushing it to repo just to ensure it's not the syntax error or any other prominent issue causing it.
Reference: https://stackoverflow.com/questions/21053657/how-to-run-travis-ci-locally

arunavabasu-03:
thanks, @anandxkumar for your feedback 😄 .

anandxkumar:
@erwanp seems like the Travis CI is not terminating, can you terminate it manually?

.travis.yml
```diff
@@ -37,7 +36,7 @@ _test: &_test
     # Update python version in environment
     - sed -i -E 's/(python=)(.*)/\1'$TRAVIS_PYTHON_VERSION'/' ./environment.yml
     # Create conda environment
-    - conda env create -n radis-env -f environment.yml
+    - conda env create -n py36 radis-env -f environment.yml
```
erwanp, line 39: default RADIS test environnment is now python 3.8, is there any specific reason you want to add py36?

arunavabasu-03: hey @erwanp , **sorry for the late reply**. I just try with another YAML file. 😀

arunavabasu-03:
hey @erwanp @anandxkumar , I add the micromamba support also.

anandxkumar:
We will first have to solve the existing errors in https://github.com/radis/radis/pull/428, so test passes and then switch it by micromamba (https://github.com/radis/radis/pull/426) to check if it working as expected and then proceed with your PR to check if it's working with resolving any conflicts and check the added window test

arunavabasu-03:
ok @anandxkumar, I am also trying to help you all  😃 .

.travis.yml
erwanp: I think the name should remain "radis-env" here

erwanp:
@arunavabasu-03 we run RADIS on Travis open-source credits, which are limited. Also, we do not want to run too much calculations and do our part in saving 🌳  

Do you think it's possible to run the Windows tests only on pull-requests to the Master branch ? (that would be for every new versino, i.e. every ~2-3 months)

anandxkumar:
Hi @arunavabasu-03, make sure to pull the latest updated develop branch, resolve any existing merge conflict before pushing the code next time!

.travis.yml
```diff
@@ -15,53 +15,78 @@ stages:
...
+jobs:
+  include:
+    - <<: *_test
+      os: windows
+      language: shell
+      before_install:
+        - choco install python --version 3.9.2
+        - python --version
+        - choco install runinbash
+      env: PATH=/c/Python39:/c/Python39/Scripts:$PATH
```
anandxkumar, line 75: @arunavabasu-03 shouldn't it be Python38?

.travis.yml
```diff
@@ -15,53 +15,78 @@ stages:
...
+    - pip install codecov
+    - codecov
+
+jobs:
+  include:
+    - <<: *_test
+      os: windows
+      language: shell
+      before_install:
+        - choco install python --version 3.9.2
```
anandxkumar, line 72: Shouldn't the python version for choco installation be the same?

.travis.yml
anandxkumar, line 75: It should get updated accordingly

arunavabasu-03:
outdated !!

Issue #382, opened 2021-10-19
Add parsers for all groups and classes


dcmvdbekerom:
### 🎯 Todo
For the parsers in io.hitran, only groups 1-2 and classes 1, 4-6 are implemented.
Notably, group 6 and class 3 are missing which are needed for linear diatomic molecules.
But this is surprising to me as I never had issues with this... am I missing something?

How they should be parsed can be found in the HITRAN2004 paper:
https://hitran.org/media/refs/HITRAN-2004.pdf
tables 3 & 4.

erwanp:
It should only be a problem if trying nonequilibrium calculations. What do you get ?

minouHub:
Hello guys,
I agree that should only be a problem for non-equilibrium calculations. As non-eq calculation of other molecules is not planned ... I'm closing the issue

PR #441, opened 2022-04-03
Add parsers to HITRAN classes

TranHuuNhatHuy:
### Description
This pull request is an attempt to provide additional parsers to HITRAN classes, and later groups if possible, according to [#382](https://github.com/radis/radis/issues/382). All parsers are checked via fetching HITRAN databases with wavenumber from 1900 cm<sup>-1</sup> to 2300 cm<sup>-1</sup> on all isotopes of a molecule from a specific class/group.
### Features
- Added HITRAN parsers to class 2, 3, 8, 9 ~~and 10~~.
- Added HITRAN syntax to notations in class 7, 8, 9 and 10, according to [Table 3](https://lweb.cfa.harvard.edu/hitran/Download/HITRAN04paper.pdf#page=7).
- (Updated Apr 4) Added local quanta identification to notations in groups 3, 4, 5 and 6, according to [Table 4](https://lweb.cfa.harvard.edu/hitran/Download/HITRAN04paper.pdf#page=8).
- Modified parsers to have a unified coding style.
### New parsers that passed the check (Updated Apr 4)
- [x] Class 2 (O2)
- [ ] Class 3 (NO, OH, ClO)
- [x] Class 8 (NH3, PH3)
- [x] Class 9 (H2O2, H2CO, COF2)
### Help wanted / Discussion needed
- Class 7 not yet attempted to add parser since I'm not sure what is the key to parse for "+-" in syntax below (for more detail please refer to [Table 3](https://lweb.cfa.harvard.edu/hitran/Download/HITRAN04paper.pdf#page=7)):
```
v1 v2 v3 v4 v5 l  +- r  S
I2 I2 I2 I2 I2 I2 A1 I1 A1
```
- (Updated Apr 4) Class 3 shows some errors when parsing global quanta with either NO, OH or ClO as below:
```
Impossible to parse global quanta in C:\Users\Admin\.radisdb\hitran\downloads__can_be_deleted\OH\OH_1.data, probably an unlabelled line. Ignoring, but nonequilibrium calculations will not be possible. See details above.
invalid literal for int() with base 10: 'X'
```
which is strange, since O2 in class 2 passed the check normally with the same `X` regexp.
- (Updated Apr 4) Since class 10 includes a wide variety of polyatomic molecules whose notations are distinct from each other, I reverted it back to the original version of simply showing message.
### Next steps
- [ ] Fix parser in class 3.
- [ ] Make parser for class 7 possible.
- [ ] Add parsers for groups 3, 4, 5 and 6.


TranHuuNhatHuy:
@erwanp @gagan-aryan excuse me, I would like to ask something regarding these parsers:

1. Problem with regexp `X` in class 3, although class 2 works perfectly fine. Updated and described in the PR description above.
2. What is the key for `+-` to parse in class 7, given their syntax below, and how do you find it? (I suspect it to be `branch` , but not sure)
```
v1 v2 v3 v4 v5 l  +- r  S
I2 I2 I2 I2 I2 I2 A1 I1 A1
```
3. Currently all the HITRAN sources I can find on the Internet, most notably [HITRAN Database](https://hitran.org/), feature HITRAN2020 databases instead of HITRAN2004 used in our I/O modules. Where can I get a raw HITRAN2004 datafile?
4. Any comments about the parsers I added? I would like to receive some feedbacks before adding more class parsers and moving to group parsers.

Hope to receive your feedback soon.

erwanp:
Hello, well done for current parsers !! 


2 . for "+/-" the key is not defined yet; you may use "symmetry" 

3. Actually, we're using HITRAN 2020 already and will be using it more and more. AFAIK its format remained compatible with HITRAN 2004, but if it's not, we should favor HITRAN 2020. 

HITRAN has their own database API ; which includes a parser; although they did not necessarily extend it to non-equilibrium. 

For test cases, you may want to retrieve data using HAPI's fetch() function, as done internally by RADIS in the fetch_hitran() function here:  https://github.com/radis/radis/blob/ce3900b063b8a168c6063da1fced652069d011a9/radis/io/hitran.py#L1164


4. For current parsers can you add small tests ? i.E. input being 1 or 2 lines of the HITRAN format, and set assertions on the actual Dataframe after parsing ?

erwanp (approved):
Excellent! Do you have access to test files? I could provide this.

TranHuuNhatHuy:
> Excellent! Do you have access to test files? I could provide this.

I would be glad to have your test files, to confirm how they actually look like ^^

In addition, 

> Excellent! Do you have access to test files? I could provide this.

Nice! I am about to add some test cases for each class/group too. Could you share your test files with me?

erwanp:
Test files sent on slack

Issue #77, opened 2020-02-20
DOCUMENTATION TODO LIST


erwanp:
# List of small documentation improvements

**New contributors**: you may want to pick one of these items and implement it. That's a good start! For that, it is recommended that you first open a [New Issue](https://github.com/radis/radis/issues/new) dedicated to the chosen item! 

**New users**: if a RADIS feature is not clear, comment here and we'll improve the documentation at some point. You can also mention in on the community chat: 

[![Gitter](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/radis-radiation/community)

- [ ] [resample](https://radis.readthedocs.io/en/latest/source/radis.spectrum.spectrum.html#radis.spectrum.spectrum.Spectrum.resample) : add examples

CorentinGrimaldi:
Add more examples in Line_survey. Also the link to the online website https://radis.readthedocs.io/en/latest/tools/line_survey.html  doesnt work.

adityabharadwaj198:
The links under "see also" (at the end of the page) don't work.
[here](https://radis.readthedocs.io/en/latest/source/radis.io.query.html)

erwanp:
Added : 

> generalize the use of "opacity", which is according to ISO 80000-7:2019 the same term as our "abscoeff", and more widely used in astrophysics.

anandxkumar:
In the LBL documentation [documentation ](https://radis.readthedocs.io/en/latest/lbl/lbl.html#flow-chart)  this [link](https://radis.readthedocs.io/en/latest/lbl/lbl/RADIS_flow_chart) doesn't exist (Radis Flow Chart) .

erwanp:
Thank you @anandxkumar for reporting. The link is now https://radis.readthedocs.io/en/latest/_images/RADIS_flow_chart.svg .  Do you want to fix it directly ?

anandxkumar:
Hello @erwanp, in the [Features Use Cases section ](https://radis.readthedocs.io/en/latest/features/features.html#use-cases), the comparison image of 2 spectra is not publishing and we are getting the alternate text `https://radis.readthedocs.io/en/latest/_images/cdsd4000_vs_hitemp_3409K.svg`, should I create a new pull request solving both issues by changing `.. image:: spectrum/cdsd4000_vs_hitemp_3409K.*` to image link in `features.rst ` and similar for the LBL documentation?

erwanp:
[...]

PR #442, opened 2022-04-05
Add some improvements in fetch_*() documentations

TranHuuNhatHuy:
### Description

This pull request is to solve issues suggested by @erwanp 's comment [here](https://github.com/radis/radis/issues/77#issuecomment-1087403583)

As [this gallery example](https://radis.readthedocs.io/en/latest/auto_examples/plot_hitemp_OH_database.html#sphx-glr-auto-examples-plot-hitemp-oh-database-py) features both HITRAN and HITEMP fetching, I add it into gallery example sessions of both `fetch_hitran()` and `fetch_hitemp()`. When a gallery example specifically for HITRAN is available, we can use it for HITRAN one.

### DIFF
--- a/radis/io/__init__.py
+++ b/radis/io/__init__.py
@@ -7,8 +7,16 @@
 
 
 from .cdsd import cdsd2df
+from .exomol import fetch_exomol
 from .hitemp import fetch_hitemp
-from .hitran import hit2df
+from .hitran import fetch_hitran, hit2df
 from .query import fetch_astroquery
 
-__all__ = ["cdsd2df", "hit2df", "fetch_hitemp", "fetch_astroquery"]
+__all__ = [
+    "cdsd2df",
+    "hit2df",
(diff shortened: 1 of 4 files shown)


erwanp (changes requested):
You may need to add it in fetch_exomol() too

TranHuuNhatHuy:
> You may need to add the gallery reference in in fetch_exomol() too

@erwanp I noticed that currently in [fetch_exomol() docs](https://radis.readthedocs.io/en/latest/source/radis.io.exomol.html#radis.io.exomol.fetch_exomol) there are two gallery references already: 
image
(are gallery example and gallery reference the same, or am I misunderstanding something? My understanding comes from #136 so if I get something wrong please enlighten me)

erwanp (approved):

erwanp:
Indeed they're already here. Thanks for checking. Merging!

Issue #444, opened 2022-04-05
Use astropy units in rescale_path_length()


erwanp:
### 🎯 Todo

Would allow this syntax, to change the column length of an already computed spectrum  : 

```python
# preparing a test spectrum :
import radis
s = radis.test_spectrum()

# rescaling : 
import astropy.units as u
s.rescale_path_length(1 * u.km).plot()       # currently not valid
```

### 💡 Possible implementations

Related to #57 

---
<sub>*Tip : you can vote for [New Features](https://github.com/radis/radis/issues/79) on https://feathub.com/radis/radis*</sub>

TranHuuNhatHuy:
@erwanp sounds interesting! I would like to tackle this problem.

erwanp:
#55 introduced a convert_and_strip_units() function, it may be used directly. The core of the PR work will be to add the corresponding small unit tests

anandxkumar:
Closing, as already been resolved in PR: https://github.com/radis/radis/pull/445

PR #445, opened 2022-04-06
Implement astropy unit conversion to rescale_path_length()

TranHuuNhatHuy:
### Description

This pull request is to address #444. Now we can use astropy units for the input arguments of `rescale_path_length()`.

Beside, currently on the RADIS documentation, `rescale_path_length()` has 2 references:

1. [radis.Spectrum.rescale_path_length](https://radis.readthedocs.io/en/latest/source/radis.html#radis.Spectrum.rescale_path_length)
2. [radis.Spectrum.rescale.rescale_path_length](https://radis.readthedocs.io/en/latest/source/radis.spectrum.rescale.html#radis.spectrum.rescale.rescale_path_length)

For [1]:

- @erwanp 's [syntax](https://github.com/radis/radis/issues/444#issue-1193462213) is added in the Examples section.
- I also notice that the gallery example for [1] does not show up on documentation. Currently the referring syntax is `.. minigallery:: radis.Spectrum.rescale_path_length`.
- Link for the source is currently not working. You can see it [here](https://radis.readthedocs.io/en/latest/source/radis.html#radis.Spectrum.rescale_path_length).

For [2], it's just kinda empty here.

### DIFF
--- a/radis/test/spectrum/test_rescale.py
+++ b/radis/test/spectrum/test_rescale.py
@@ -296,19 +296,15 @@ def test_astropy_units(verbose=True, warnings=True, *args, **kwargs):
     s_km = s0.copy().rescale_path_length(1 * u.km)  # Rescale directly to 1 astropy km
 
     # Get absorbance of original spectrum
-    w0, k0 = s0.get("abscoeff", wunit="cm-1")
-    L0 = s0.get_conditions()["path_length"]
-    A0 = k0 * L0
+    A0 = s0.get("absorbance", wunit="cm-1")
 
-    # Get absorbance of 100000cm-rescaled spectrum
-    w_cm, k_cm = s_cm.get("abscoeff", wunit="cm-1")
+    # Get absorbance and path length of 100000cm-rescaled spectrum
     L_cm = s_cm.get_conditions()["path_length"]
-    A_cm = k_cm * L_cm
+    A_cm = s_cm.get("absorbance", wunit="cm-1")
 
-    # Get absorbance of 1km-rescale spectrum
-    w_km, k_km = s_km.get("abscoeff", wunit="cm-1")
+    # Get absorbance and path length of 1km-rescale spectrum
     L_km = s_km.get_conditions()["path_length"]
-    A_km = k_km * L_km
+    A_km = s_km.get("absorbance", wunit="cm-1")
 
     # ---------- ASSERTION ----------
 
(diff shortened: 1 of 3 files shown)


erwanp:
Nice ! I think it should be _radis.**spectrum**.Spectrum.rescale_path_length_

there is a script in [doc/gendocs.sh](https://github.com/radis/radis/blob/develop/docs/gendocs.sh) that allows you to generate docs locally; helps the debugging. You may need to install the [doc specific requirements](https://github.com/radis/radis/blob/develop/docs/requirements.txt)

erwanp:
What we'll need before merging is to add **tests**.

There is a [test_rescale.py](https://github.com/radis/radis/blob/develop/radis/test/spectrum/test_rescale.py) file. You can add a new function there, testing the units. 

You can use a precomputed spectrum such as
```
s = load_spec(getTestFile("CO_Tgas1500K_mole_fraction0.01.spec"), binary=True)
```

and you could, for instance, compare the `absorbance`  (in absorption spectroscopy, the absorbance is A = k * L  where k is the absorption coefficient, L the length ; so absorbance should have changed when you rescaled the spectra )  of a a spectrum rescaled with astropy units (e.g `1 * u.m` ) and with default units (`100 # cm` )

radis/test/spectrum/test_rescale.py
```diff
@@ -282,14 +283,59 @@ def test_xsections(*args, **kwargs):
...
+
+    # Get absorbance of 1km-rescale spectrum
+    w_km, k_km = s_km.get("abscoeff", wunit="cm-1")
+    L_km = s_km.get_conditions()["path_length"]
+    A_km = k_km * L_km
+
+    # ---------- ASSERTION ----------
+
+    # Compare absorbances of original and 1km. They should be DIFFERENT.
+    assert not np.array_equal(A0, A_km)
```
erwanp, line 316: good idea to ensure this ! 👍

erwanp (changes requested):

TranHuuNhatHuy:
@erwanp thank you for the reviews. Last commit features two changes according to your suggestions:

1. Fetch `absorbance` directly by using `s.get("absorbance", wunit='cm-1')`. I believe this is better than fetching `k` and `L` then calculate the absorbance `A = k * L`, in terms of computational performance and the sake of clean code.
2. Add third assertion of comparing the two `path_length` of 100000 cm and 1 u.km to make sure the astropy unit works fine. Although I doubt it is necessary if the two absorbances are already equal, I think with this we can check if something is wrong with the astropy units in the future (extremely unlikely thou).

erwanp (approved):

erwanp:
Excellent, merging!

anandxkumar:
Fixes #444

PR #447, opened 2022-04-11
Fixes Issue#443

### LINKED ISSUE #443: Add truncation=None option
### 🎯 Todo

Allow `truncation=None` in SpectrumFactory and calc_spectrum inputs; which should perform as follow : calculate each lineshape over the full spectral range (including neighbour lines)

Supriya1702:
### Description

This PR does the following:
- If `truncation=None` for `broadening_method='voigt'` for `optimization='simple' or 'min-RMS` method it raises NotImplementedError
- If `truncation` is negative it raises value error
- Allows computation for `broadening_method == "voigt"` and `truncation == None` for `optimization=None`

Fixes #443

### DIFF
--- a/radis/lbl/factory.py
+++ b/radis/lbl/factory.py
@@ -540,13 +540,13 @@ class SpectrumFactory(BandFactory):
         # Time Based variables
         self.verbose = verbose
 
-        
-
         if truncation == 0:
             raise ValueError(
                 "Lineshape truncation must be > 0. If you want no truncation (compute lineshape on the full spectral range), use `truncation=None`"
             )
-        elif (truncation is not None and not isinstance(truncation, Default)) and truncation < 0:
+        elif (
+            truncation is not None and not isinstance(truncation, Default)
+        ) and truncation < 0:
             raise ValueError(
                 "Lineshape truncation can't be negative. Truncation must be > 0 or None to compute lineshape on the full spectral range"
             )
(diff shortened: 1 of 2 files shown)


anandxkumar:
Tested for different cases, works fine on my end, Good job! 
@erwanp should `truncation=None` work for `optimization=None` for `broadening_method='voigt'`?

radis/lbl/factory.py
```diff
@@ -542,7 +542,13 @@ def __init__(
 
         if truncation == 0:
             raise ValueError(
-                "Lineshape truncation must be >0. If you want no truncation (compute lineshape on the full spectral range), use `truncation=None`"
+                "Lineshape truncation must be > 0. If you want no truncation (compute lineshape on the full spectral range), use `truncation=None`"
```
erwanp, line 545: Can you also add :

"Note (advanced) : no truncation is not physically more accurate. Most molecules exhibit a sub-lorentzian behavior far from the line centers. A truncation at around 40-50 cm-1 is a good choice"

erwanp (changes requested):
Hello, good work! 

1. are you sure optimization=None is implemented?

2. We'd need a quick test, for instance one making sure an error is raised if negative truncation is given, and one making sure truncation=None does what we want. If you need help implementing tests please reach out on Slack

Supriya1702:
Hello @erwanp, I added test for checking negative truncation and `truncation==None`. I found that `truncation=None` works for `broadening_method='voigt'` for `optimization=None` but gives error when used `fft`. Oppositely in `optimization='simple`, `truncation=None` works for `fft` but fails for `voigt`, so I added the tests accordingly

erwanp (approved):

PR #448, opened 2022-04-12
Normalize() return norm factor and units option

CorentinGrimaldi:
### Description

Add the `return_norm` option in `spectrum.normalize()` to return the norm factor and its unit.
Usefull when rescaling mole fraction to match an experimental spectrum.

### DIFF
--- a/radis/spectrum/spectrum.py
+++ b/radis/spectrum/spectrum.py
@@ -4311,6 +4311,7 @@ class Spectrum(object):
         inplace=False,
         force=False,
         verbose=False,
+        return_norm=False,
     ):
         """Normalise the Spectrum, if only one spectral quantity is available.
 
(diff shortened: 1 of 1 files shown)


erwanp:
Nice ! Can you also add : 

- [ ] a very minimal example of how you use it in the docstrings of [normalize()](https://radis.readthedocs.io/en/latest/source/radis.spectrum.spectrum.html#radis.spectrum.spectrum.Spectrum.normalize) ?
- [ ] a minimal test to check the normalize() parameter and make sure it never ever changes (whatever astropy units become). I think [here in test_spectrum.py](https://github.com/radis/radis/blob/develop/radis/test/spectrum/test_spectrum.py#L521)  is the best place (everything is already initialized)

erwanp (changes requested):
(see my comment above on Checks & tests)

Can you also add a comment in the dostring
```
Other Paramters
------------------
return_norm: bool 
if True, returns the normalization constant (with unit) . 
```

erwanp:
(merging so you can use it; please update doc one day ;)  )

PR #451, opened 2022-04-14
Added cutoff_error parameter to enable Automatic Linestrength Cutoff

### LINKED ISSUE #268: Automatic linestrength cutoff
Linestrength cutoff can be defined in calc_spectrum and SpectrumFactory with `cutoff=` 

A good improvement, similar to #250 for automatic wstep adaptation, would be that the user inputs an acceptable error (0.1%, 1%, 10%, etc.), and the cutoff would be adjusted automatically. Modifications would be done in [base.py > _cutoff_linestrength()](https://github.com/radis/radis/blob/develop/radis/lbl/base.py#L3146)
(Note : this "Automatic cutoff" should be another Issue, here we're only talking about allowing the "cutoff=None" syntax )

_Originally posted by @erwanp in https://github.com/radis/radis/issues/251#issuecomment-822563433_

sagarchotalia:
### Description
This PR is to address #268.
Added `cutoff_error` parameter that will allow users to input a cutoff error, which will then be used to keep the internally calculated error below the inputted cutoff error.
The way I am doing this currently is, decrementing the `cutoff` by `0.001` (arbitrary value assumed for now, will change it if needed) and then recalculating the bool array 'b' and then calculating the internal error again
The size of the boolean array 'b' shall decrease with each iteration until the internal error is less than the cutoff error.

### Further Changes:
I am not sure that the approach I am using now fixes #268. If there is a more efficient way to decrease the internal error, kindly suggest the same in the comments.


radis/lbl/base.py
```diff
@@ -3202,6 +3209,26 @@ def _cutoff_linestrength(self, cutoff=None):
 
             error = df.S[b].sum() / df.S.sum() * 100
 
+            if cutoff_error is not None:
+                if cutoff_error < error:
+                    # Remove additional lines such that cutoff error is less than user-inputted value
+                    while cutoff_error < error:
+                        cutoff -= 0.0001
+                        b = df.S <= cutoff
+                        error = df.S[b].sum() / df.S.sum() * 100
```
erwanp, line 3218: what about : 
1 . we sort (a copy of) the lines by intensity intensity
2. we compute the cumulated intensity on the sorted array  [last value is the total sum]
3.  we find the index in the sorted array so that cumulated-intensity is ERROR_fraction * the total

erwanp (changes requested):

erwanp:
Also, see the error raised : https://app.travis-ci.com/github/radis/radis/jobs/567095871#L1858

sagarchotalia:
> Also, see the error raised : https://app.travis-ci.com/github/radis/radis/jobs/567095871#L1858

Yes, the error was resolved in the next commit.

radis/lbl/base.py
sagarchotalia: Yes, this seems to be a much better option. I'll work on this.

erwanp: Please let me know. It will be important to also add tests.

erwanp: Hello @sagarchotalia ,did you try the approach I suggested above ?

erwanp (changes requested):

erwanp (changes requested):
All these operations exist as Numpy array methods. You should use them : it will make your code much more readable; and also much more efficient. List are pure Python objects; whereas Numpy arrays run underlying C code

Issue #486, opened 2022-05-31
Bug Report: Exomol, hdf5 tests failing on travis (buffer error)


anandxkumar:
### 🐛 Describe the bug

Travis failing: https://app.travis-ci.com/github/radis/radis/jobs/571532389

Steps to reproduce:
- create a new virtual environment using `conda create -n radis_env python=3.8`  and fetch latest develop branch locally
- `pip install -e .`  (pip install -e .[dev] -v didn't work)
- delete `radis.json` and `.radisdb`
- `pytest`

<details>
  <summary>Following is the version of dependencies in the environment:</summary>

- anyio==3.6.1
- aplus==0.11.0
- argon2-cffi==21.3.0
- argon2-cffi-bindings==21.2.0
- asgiref==3.5.2
- astropy==5.1
- astroquery==0.4.6
- asttokens==2.0.5
- attrs==21.4.0
- backcall==0.2.0
- beautifulsoup4==4.11.1
- blake3==0.3.1
- bleach==5.0.0
- bqplot==0.12.33
- cachetools==5.2.0
- certifi==2022.5.18.1
- cffi==1.15.0
- charset-normalizer==2.0.12
- click==8.1.3
- cloudpickle==2.1.0
- commonmark==0.9.1
- configparser==5.2.0
- cryptography==37.0.2
- cycler==0.11.0
- Cython==0.29.30
- dask==2022.5.2
- debugpy==1.6.0
- decorator==5.1.1
- defusedxml==0.7.1
- entrypoints==0.4
- executing==0.8.3
- fastapi==0.78.0
- fastjsonschema==2.15.3
- filelock==3.7.0
- fonttools==4.33.3
- frozendict==2.3.2
- fsspec==2022.5.0
- future==0.18.2
- h11==0.13.0
- h5py==3.7.0
- habanero==1.2.2
- hitran-api==1.2.2.0
- hjson==3.0.2
- html5lib==1.1
- httptools==0.4.0
- idna==3.3
- importlib-metadata==4.11.4
- importlib-resources==5.7.1
- iniconfig==1.1.1
- ipydatawidgets==4.3.1.post1
- ipykernel==6.13.0
- ipyleaflet==0.16.0
- ipympl==0.9.1
- ipython==8.4.0
- ipython-genutils==0.2.0
- ipyvolume==0.5.2
- ipyvue==1.7.0
- ipyvuetify==1.8.2
- ipywebrtc==0.6.0
- ipywidgets==7.7.0
[...]

PR #487, opened 2022-06-02
Fix HDF5/ api issues

erwanp:
### Description

This pull request is to address #486 

Fixed some of the problems encountered in HITRAN & HDF5 Query . 
At least one problem is not yet resolved and requires the release of Vaex 4.9.2  (see #486)


erwanp:
@anandxkumar you can also merge this one following the merge of #490. It fixes one extra possible error (not caught in tests)

PR #489, opened 2022-06-09
Chunk size implementation for DIT Algorithm loops for Broadening

### LINKED ISSUE #488: Adding a Chunk size method for broadening in Optimized loops
### 🐛 Describe the issue

This issue pertains to adding the 'chunksize' feature to the optimized selection in [_broaden_lines](https://github.com/radis/radis/blob/develop/radis/lbl/broadening.py#:~:text=def%20_broaden_lines(self%2C%20df)%3A), which is already present for non-optimized lines on line 2181.

#### My understanding of the code:
- We use the `_broaden_lines` (and `_broaden_lines_noneq`) method to calculate the wavenumber and absorption coefficients as required by the `_calc_broadening` (and `_calc_broadening_noneq`) method, in order to calculate the broadening over lines.
- When the `_broaden_lines` method is called, the `optimization` parameter is fetched. If optimization is to be applied, then the parameters are calculated using the entire dataframe at once.
- However, when the `optimization` parameter is `None`, we check for `chunksize`:
    - if `chunksize` is `None`, we calculate the line_profile parameter using the entire dataframe, which is faster but consumes a lot of memory.
    - if `chunksize` is a float-like type, then the N parameter is calculated, and a loop is run where we iterate over each row of the enumerated dataframe line values and calculate the line profile for each row. Then, `_apply_lineshape` method gives us the (wavenumber, absorption) parameters for each row, and the absorption is added to `abscoeff` at the end of each iteration. Then the progress bar is updated.

#### The problem:
- The problem with calculating the parameters with the entire dataframe is that it might be fast, but it consumes a lot of memory, which is why we want the chunksize strategy in the optimized loop as well. Implied that, implementing chunksize will probably make the code slower, but it consumes lesser memory.
#### Implementation strategy
- The implementation for the optimized loops shall be similar to the current technique being used in the unoptimized version. We can enumerate the dataframe using the same formula `N = int(len(df) * len(wavenumber) / chunksize) + 1`
- The main change would be during the calculation of lineshapes via Line Density Map method, since this method has the extra parameters ` wL, wG, wL_dat, wG_dat` which will have to be calculated in each iteration. (Help in this part would be appreciated!)

sagarchotalia:
### Objectives
This pull request is to address the chunksize implementation in DIT Algorithm(optimized) loops in the [_broaden_lines](https://github.com/radis/radis/blob/develop/radis/lbl/broadening.py#:~:text=def%20_broaden_lines(self%2C%20df)%3A) method. 
With this PR, RADIS users shall be able to fully utilize the Chunksize feature in the code. Instead of looping the entire DataFrame, we use chunks of it at a time, hence allowing for much better memory management!

### Changes Made
- I restructured the code such that the outer loops check for the chunksize variable, since I think that it'll improve code readability. I've only added the time prediction block of code twice at the start of each loop.
- Arrays are being updated according to the Chunksize.
- Chunksize is now a parameter in [calc_spectrum()](https://github.com/radis/radis/blob/b853a26b34a0b9b53670062a2214da4d134ec391/radis/lbl/calc.py#L38) and [calc_spectrum_one_molecule()](https://github.com/radis/radis/blob/b853a26b34a0b9b53670062a2214da4d134ec391/radis/lbl/calc.py#L568)

### Future Objectives
- [ ] Support for non-equilibrium loops as well (see [_broaden_lines_noneq](https://github.com/radis/radis/blob/b853a26b34a0b9b53670062a2214da4d134ec391/radis/lbl/broadening.py#L2231))
- [ ] Add an "auto" parameter for chunksize, which will calculate the best chunksize based on user-RAM.

Fixes #488

### DIFF
--- a/radis/lbl/broadening.py
+++ b/radis/lbl/broadening.py
@@ -2194,16 +2194,17 @@ class BroadenFactory(BaseFactory):
         try:
             if chunksize is None:
                 # Deal with all lines directly (usually faster)
-                # printing estimated time
-                if self.verbose >= 2:
-                    estimated_time = self.predict_time()
-                    print(
-                        "Estimated time for calculating broadening: {0:.2f}s on 1 CPU".format(
-                            estimated_time
-                        )
-                    )
-
                 if optimization is None:
+
+                    # printing estimated time
+                    if self.verbose >= 2:
+                        estimated_time = self.predict_time()
+                        print(
+                            "Estimated time for calculating broadening: {0:.2f}s on 1 CPU".format(
+                                estimated_time
+                            )
+                        )
+
                     line_profile = self._calc_lineshape(df)  # usually the bottleneck
                     (wavenumber, abscoeff) = self._apply_lineshape(
                         df.S.values, line_profile, df.shiftwav.values
@@ -2217,6 +2218,16 @@ class BroadenFactory(BaseFactory):
                         wL_dat,
                         wG_dat,
                     ) = self._calc_lineshape_LDM(df)
+
+                    # printing estimated time
+                    if self.verbose >= 2:
+                        estimated_time = self.predict_time()
+                        print(
+                            "Estimated time for calculating broadening: {0:.2f}s on 1 CPU".format(
+                                estimated_time
+                            )
+                        )
+
                     (wavenumber, abscoeff) = self._apply_lineshape_LDM(
                         df.S.values,
                         line_profile_LDM,
@@ -2240,16 +2251,17 @@ class BroadenFactory(BaseFactory):
                 abscoeff = zeros_like(self.wavenumber)
                 pb = ProgressBar(N, active=self.verbose)
 
-                # printing estimated time
-                if self.verbose >= 2:
-                    estimated_time = self.predict_time()
-                    print(
-                        "Estimated time for calculating broadening: {0:.2f}s on 1 CPU".format(
-                            estimated_time
-                        )
-                    )
-
                 if optimization is None:
+
+                    # printing estimated time
+                    if self.verbose >= 2:
+                        estimated_time = self.predict_time()
+                        print(
+                            "Estimated time for calculating broadening: {0:.2f}s on 1 CPU".format(
+                                estimated_time
+                            )
+                        )
+
                     for i, (_, dg) in enumerate(df.groupby(arange(len(df)) % N)):
                         line_profile = self._calc_lineshape(dg)
                         (wavenumber, absorption) = self._apply_lineshape(
--- a/radis/lbl/broadening.py
+++ b/radis/lbl/broadening.py
@@ -2166,7 +2166,7 @@ class BroadenFactory(BaseFactory):
                     line_profile_LDM, wL, wG, wL_dat, wG_dat = self._calc_lineshape_LDM(
                         df
                     )
-                    (wavenumber, absorption) = self._apply_lineshape_LDM(
+                    (wavenumber, abscoeff) = self._apply_lineshape_LDM(
                         df.S.values,
                         line_profile_LDM,
                         df.shiftwav.values,
--- a/radis/lbl/broadening.py
+++ b/radis/lbl/broadening.py
@@ -2126,6 +2126,53 @@ class BroadenFactory(BaseFactory):
             line dataframe
 
         See _calc_lineshape for more information
+
+        Examples
+        ----------
+        s = calc_spectrum(
+            2135,
+            2170,
+            molecule="CO",
+            isotope="1",
+            pressure=3,
+            Tgas=2000,
+            mole_fraction=0.1,
+            path_length=1,
+            databank="hitemp",
+            name="Chunksize=1e7",
+            chunksize=1e7,
+            optimization = "min-RMS",
+        )
+
+        Alternatively, you can also initialize a SpectrumFactory object and
+        include chunksize, as follows:
+
+        sf = SpectrumFactory(
+            wavelength_min=4000,
+            wavelength_max=4500,
+            cutoff=1e-27,
+            pressure=1,
+            isotope="1,2",
+            truncation=5,
+            neighbour_lines=5,
+            path_length=0.1,
+            mole_fraction=1e-3,
+            medium="vacuum",
+            optimization=None,
+            chunksize=1e7,
+            wstep=0.001,
+            verbose=False,
+        )
+        sf.load_databank("HITEMP-CO")
+
+        To plot:
+
+        s.plot("abscoeff")
+        plt.show()
+
+        To iterate over the entire dataframe at once (not recommended for large molecules
+        unless you have large RAM), just pass chunksize = None
+
         """
         # --------------------------
 
--- a/radis/lbl/broadening.py
+++ b/radis/lbl/broadening.py
@@ -2284,16 +2284,26 @@ class BroadenFactory(BaseFactory):
         optimization = self.params.optimization
 
         try:
-            if optimization in ("simple", "min-RMS"):
-                self.reftracker.add(doi["DIT-2020"], "algorithm")
-                # Use LDM
+            # Note @dev: typical results is:
+            # >>> abscoeff:
+            # ... Precomputed LDM lineshapes in 0.0s
+            # ... Initialized vectors in 0.0s
+            # ... Get closest matching line & fraction in 0.3s
+            # ... Distribute lines over LDM 2.1s
+            # ... Convolve and sum on spectral range 0.4s
+            # >>> emisscoeff
+            # ... Initialized vectors in 0.0s
+            # ... Get closest matching line & fraction in 0.2s
+            # ... Distribute lines over LDM 2.1s
+            # ... Convolve and sum on spectral range 0.3s
+            # @EP: #performance.
+            # unlike in the non LDM case, the nonequilibruum case here is ~2x
+            # the equilibrium case: only the closest matching line is common to the
+            # absorption & emission steps. The bottleneck is the distribution
+            # of the line over the LDM, which has to be done for both abscoeff & emisscoeff.
 
-                if self.misc.zero_padding < 0 or self.misc.zero_padding > len(
-                    self.wavenumber_calc
-                ):
-                    self.misc.zero_padding = len(self.wavenumber_calc)
-
-                line_profile_LDM, wL, wG, wL_dat, wG_dat = self._calc_lineshape_LDM(df)
+            if chunksize is None:
+                # Deal with all lines directly (usually faster)
                 # printing estimated time
                 if self.verbose >= 2:
                     estimated_time = self.predict_time()
@@ -2302,55 +2312,8 @@ class BroadenFactory(BaseFactory):
                             estimated_time
                         )
                     )
-                (wavenumber, abscoeff) = self._apply_lineshape_LDM(
-                    df.S.values,
-                    line_profile_LDM,
-                    df.shiftwav.values,
-                    wL,
-                    wG,
-                    wL_dat,
-                    wG_dat,
-                    optimization,
-                )
-                (_, emisscoeff) = self._apply_lineshape_LDM(
-                    df.Ei.values,
-                    line_profile_LDM,
-                    df.shiftwav.values,
-                    wL,
-                    wG,
-                    wL_dat,
-                    wG_dat,
-                    optimization,
-                )
-                # Note @dev: typical results is:
-                # >>> abscoeff:
-                # ... Precomputed LDM lineshapes in 0.0s
-                # ... Initialized vectors in 0.0s
-                # ... Get closest matching line & fraction in 0.3s
-                # ... Distribute lines over LDM 2.1s
-                # ... Convolve and sum on spectral range 0.4s
-                # >>> emisscoeff
-                # ... Initialized vectors in 0.0s
-                # ... Get closest matching line & fraction in 0.2s
-                # ... Distribute lines over LDM 2.1s
-                # ... Convolve and sum on spectral range 0.3s
-                # @EP: #performance.
-                # unlike in the non LDM case, the nonequilibruum case here is ~2x
-                # the equilibrium case: only the closest matching line is common to the
-                # absorption & emission steps. The bottleneck is the distribution
-                # of the line over the LDM, which has to be done for both abscoeff & emisscoeff.
 
-            elif optimization is None:
-                # printing estimated time
-                if self.verbose >= 2:
-                    estimated_time = self.predict_time()
-                    print(
-                        "Estimated time for calculating broadening: {0:.2f}s on 1 CPU".format(
-                            estimated_time
-                        )
-                    )
-                if chunksize is None:
-                    # Deal with all lines directly (usually faster)
+                if optimization is None:
                     line_profile = self._calc_lineshape(df)  # usually the bottleneck
                     (wavenumber, abscoeff) = self._apply_lineshape(
                         df.S.values, line_profile, df.shiftwav.values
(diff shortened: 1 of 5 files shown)


erwanp:
Hello, good first work!
Can you provide an example by testing, for instance the Radis basic example `radis.test_spectrum()` using different optimization and chunksize as parameters?

sagarchotalia:
Hello, sorry for getting back so late.
I tested the example after just commenting out the DeprecationWarning lines in calc.py;
```
# if "chunksize" in kwargs:
#     raise DeprecationWarning("use optimization= instead of chunksize=")
```
Then, I tested the basic example test_spectrum() as such: 
```
s = radis.test_spectrum(optimization = "min-RMS", chunksize = 100)
```
The spectrum is calculated, however I do receive an error for the chunksize parameter:
```
Calculating Equilibrium Spectrum
Physical Conditions
   Tgas                 700 K
   Trot                 700 K
   Tvib                 700 K
   isotope              1,2,3
   mole_fraction        0.1
   molecule             CO
   overpopulation       None
   path_length          1 cm
   self_absorption      True
   state                X
   wavenum_max          2300.0000 cm-1
   wavenum_min          1900.0000 cm-1
Computation Parameters
   Tref                 296 K
   add_at_used          
   broadening_method    voigt
   dbformat             hitran
   dbpath               /Users/sagarchotalia/.radisdb/hitran/CO.hdf5
   folding_thresh       1e-06
   memory_mapping_engine  auto
   neighbour_lines      0 cm-1
   optimization         min-RMS
   parfuncfmt           hapi
   parsum_mode          full summation
   pseudo_continuum_threshold  0
   sparse_ldm           auto
   truncation           50 cm-1
   waveunit             cm-1
   zero_padding         -1
  File "/Users/sagarchotalia/Desktop/radis temp files/example.py", line 24, in <module>
    s = radis.test_spectrum(optimization = "min-RMS", chunksize = 100)
  File "/Users/sagarchotalia/radis/radis/test/utils.py", line 187, in test_spectrum
    s = calc_spectrum(**conditions)
  File "/Users/sagarchotalia/radis/radis/lbl/calc.py", line 514, in calc_spectrum
  File "/Users/sagarchotalia/radis/radis/lbl/calc.py", line 838, in _calc_spectrum_one_molecule
    s = sf.eq_spectrum(
  File "/Users/sagarchotalia/radis/radis/lbl/factory.py", line 799, in eq_spectrum
    wavenumber, abscoeff_v = self._calc_broadening()
  File "/Users/sagarchotalia/radis/radis/lbl/broadening.py", line 2449, in _calc_broadening
    (wavenumber, abscoeff) = self._broaden_lines(df)
  File "/Users/sagarchotalia/radis/radis/lbl/broadening.py", line 2220, in _broaden_lines
    (wavenumber, absorption) = self._apply_lineshape_LDM(
  File "/Users/sagarchotalia/radis/radis/lbl/broadening.py", line 1934, in _apply_lineshape_LDM
    df = pd.DataFrame(
  File "/Users/sagarchotalia/opt/anaconda3/envs/radis-env/lib/python3.8/site-packages/pandas/core/frame.py", line 636, in __init__
    mgr = dict_to_mgr(data, index, columns, dtype=dtype, copy=copy, typ=manager)
  File "/Users/sagarchotalia/opt/anaconda3/envs/radis-env/lib/python3.8/site-packages/pandas/core/internals/construction.py", line 502, in dict_to_mgr
    return arrays_to_mgr(arrays, columns, index, dtype=dtype, typ=typ, consolidate=copy)
  File "/Users/sagarchotalia/opt/anaconda3/envs/radis-env/lib/python3.8/site-packages/pandas/core/internals/construction.py", line 120, in arrays_to_mgr
    index = _extract_index(arrays)
  File "/Users/sagarchotalia/opt/anaconda3/envs/radis-env/lib/python3.8/site-packages/pandas/core/internals/construction.py", line 674, in _extract_index
    raise ValueError("All arrays must be of the same length")
ValueError: All arrays must be of the same length
```

Of course, this error doesn't occur if `chunksize=None` is entered as an argument in the function.

To test this example, clone [my RADIS fork](https://github.com/sagarchotalia/radis) and run the test on your local machine.

sagarchotalia:
Update: the errors were being generated due to path errors in my `~/radis.json`. I fixed them, and now it's working properly.
The spectrum is being calculated upon being given a chunksize:
`s = radis.test_spectrum(optimization = "simple", chunksize = 1000)`

Output:
```
Calculating Equilibrium Spectrum
Physical Conditions
   Tgas                 700 K
   Trot                 700 K
   Tvib                 700 K
   isotope              1,2,3
   mole_fraction        0.1
   molecule             CO
   overpopulation       None
   path_length          1 cm
   self_absorption      True
   state                X
   wavenum_max          2300.0000 cm-1
   wavenum_min          1900.0000 cm-1
Computation Parameters
   Tref                 296 K
   add_at_used          
   broadening_method    voigt
   dbformat             hitran
   dbpath               /Users/sagarchotalia/.radisdb/hitran/CO.hdf5
   folding_thresh       1e-06
   memory_mapping_engine  auto
   neighbour_lines      0 cm-1
   optimization         simple
   parfuncfmt           hapi
   parsum_mode          full summation
   pseudo_continuum_threshold  0
   sparse_ldm           auto
   truncation           50 cm-1
   waveunit             cm-1
   zero_padding         -1
0.50s - Spectrum calculated
```

sagarchotalia:
Hello, I've confirmed that both the spectra are the same, with and without chunksize!
Also, I've tested the RADIS examples [calc_hitran_full_range.py](https://radis.readthedocs.io/en/latest/auto_examples/calc_hitran_full_range.html), [plot_hitemp_spectrum.py](https://radis.readthedocs.io/en/latest/auto_examples/plot_hitemp_spectrum.html) and [plot_SpecDatabase.py](https://radis.readthedocs.io/en/latest/auto_examples/plot_SpecDatabase.html)both with and without chunksize. They're yielding the same results as well.

P.S. Another thing, I'm having issues plotting through matplotlib due to a publib warning:
```
/Users/sagarchotalia/opt/anaconda3/lib/python3.8/site-packages/publib/main.py:230: UserWarning: 
Glyph 8315 (\N{SUPERSCRIPT MINUS}) missing from current font.
  plt.tight_layout()
```
Due to which I'm not able to plot some spectra. But I'll try to get it resolved asap!

erwanp:
Hello! You can change the default plotting style and library of Radis, drop publib and use something else. There is a Gallery Example showing how to customize plots

sagarchotalia:
> ```
> ----------------------------------------
> Traceback (most recent call last):
>   File "/Users/sagarchotalia/Desktop/radis temp files/example.py", line 24, in <module>
>     s = radis.test_spectrum(optimization = "min-RMS", chunksize = 100)
>   File "/Users/sagarchotalia/radis/radis/test/utils.py", line 187, in test_spectrum
>     s = calc_spectrum(**conditions)
>   File "/Users/sagarchotalia/radis/radis/lbl/calc.py", line 514, in calc_spectrum
>     generated_spectrum = _calc_spectrum_one_molecule(
>   File "/Users/sagarchotalia/radis/radis/lbl/calc.py", line 838, in _calc_spectrum_one_molecule
>     s = sf.eq_spectrum(
>   File "/Users/sagarchotalia/radis/radis/lbl/factory.py", line 799, in eq_spectrum
>     wavenumber, abscoeff_v = self._calc_broadening()
>   File "/Users/sagarchotalia/radis/radis/lbl/broadening.py", line 2449, in _calc_broadening
>     (wavenumber, abscoeff) = self._broaden_lines(df)
>   File "/Users/sagarchotalia/radis/radis/lbl/broadening.py", line 2220, in _broaden_lines
>     (wavenumber, absorption) = self._apply_lineshape_LDM(
>   File "/Users/sagarchotalia/radis/radis/lbl/broadening.py", line 1934, in _apply_lineshape_LDM
>     df = pd.DataFrame(
>   File "/Users/sagarchotalia/opt/anaconda3/envs/radis-env/lib/python3.8/site-packages/pandas/core/frame.py", line 636, in __init__
>     mgr = dict_to_mgr(data, index, columns, dtype=dtype, copy=copy, typ=manager)
>   File "/Users/sagarchotalia/opt/anaconda3/envs/radis-env/lib/python3.8/site-packages/pandas/core/internals/construction.py", line 502, in dict_to_mgr
>     return arrays_to_mgr(arrays, columns, index, dtype=dtype, typ=typ, consolidate=copy)
>   File "/Users/sagarchotalia/opt/anaconda3/envs/radis-env/lib/python3.8/site-packages/pandas/core/internals/construction.py", line 120, in arrays_to_mgr
>     index = _extract_index(arrays)
>   File "/Users/sagarchotalia/opt/anaconda3/envs/radis-env/lib/python3.8/site-packages/pandas/core/internals/construction.py", line 674, in _extract_index
>     raise ValueError("All arrays must be of the same length")
> ValueError: All arrays must be of the same length
> ```


It seems that this error is persisting even now. It's occurring on [this line](https://github.com/radis/radis/blob/b853a26b34a0b9b53670062a2214da4d134ec391/radis/lbl/broadening.py#L1934) of _apply_lineshape_LDM(), giving a ValueError that all arrays should be of the same length.


Also, here are the Pytest results after chunksize implementation:
Screenshot 2022-07-09 at 1 26 00 PM

radis/lbl/broadening.py
```diff
@@ -2178,23 +2197,8 @@ def _broaden_lines(self, df):
                             estimated_time
                         )
                     )
-                if chunksize is None:
-
-                    # Deal with all lines directly (usually faster)
```
erwanp, line 2183: did you remove this comment from the code? if so, add it back

radis/lbl/broadening.py
```diff
@@ -2203,14 +2207,37 @@ def _broaden_lines(self, df):
...
+                    for i, (_, dg) in enumerate(df.groupby(arange(len(df)) % N)):
+                        (
+                            line_profile_LDM,
+                            wL,
+                            wG,
+                            wL_dat,
+                            wG_dat,
+                        ) = self._calc_lineshape_LDM(dg)
+                        (wavenumber, absorption) = self._apply_lineshape_LDM(
+                            df.S.values,
```
erwanp, line 2221: shouldn't you use `dg.S.values` here?

erwanp (changes requested):
See comments in code

also 
- [ ] make sure you add in your PR objectives to also make the changes for nonequilibrium mode (AFAIK, there is a "broaden_lines_noneq" function somewhere) . Do this only after equilibrium works.

radis/lbl/broadening.py
sagarchotalia: Sorry, I might have removed it by mistake. Adding it back.

radis/lbl/broadening.py
sagarchotalia: Yes you're right, I actually changed that a while back. I'll commit the latest version soon.

sagarchotalia:
Here are the plot_diff() results for a HITEMP CO spectra, with the same parameters(optimization="simple"), the only difference being the chunksize.
image

What I understand is happening:
- The values of `line_profile_LDM`, `wL` and `wG` are not the same when calculated over one chunk of the database vs. over the entire database. These values later go in [_apply_lineshape_LDM()](https://github.com/radis/radis/blob/b853a26b34a0b9b53670062a2214da4d134ec391/radis/lbl/broadening.py#L1746) to compute the coefficients such as li0, li1, mi0, mi1 etc., which affects the overall spectrum, hence the overall result is not the same. Hence these values need to be the same with or without Chunksize.

sagarchotalia:
Hello! Turns out my thinking was in the wrong direction, I finally fixed the error being caused by my implementation by taking `shifted_wavenum` over the chunks instead of the entire df.
Here's some spectra outputs:
image
**N2O Spectrum**
image
**CO2 Spectrum**
image
**CO Spectrum**

If these are alright, I can go ahead with the chunksize implementation for non-equilibrium spectra.

radis/lbl/broadening.py
```diff
@@ -2203,14 +2207,41 @@ def _broaden_lines(self, df):
...
                     pb.done()
+
+                elif optimization in ("simple", "min-RMS"):
+
+                    self.reftracker.add(doi["DIT-2020"], "algorithm")
+                    # Iterating over the chunks of the line database
+                    # Using DIT Algorithm calculations for optimized loops
+                    for i, (_, dg) in enumerate(df.groupby(arange(len(df)) % N)):
+                        (
+                            line_profile_LDM,
```
erwanp, line 2218: Tips : rename variables wL_i etc. to remember they only apply within the loop

erwanp:
Hello @sagarchotalia These results seems great ! 

Before moving to nonequilibrium we'll have to : 
- [ ] implement tests in `radis/test/lbl`, with one function computing a spectrum with/without chunksize optimization; and making sure the residual (get_residual) is very small (which it seems to be; right now) ; and remain small forever.

We also have to make sure **it is more efficient**. Computing all lines together is more efficient if all lines hold in RAM. It becomes problematic when you saturate RAM. With Chunksize, you have an implementation which will ensure you never saturate RAM. 
- [ ] Compute a large spectrum, with lots of lines. For instance HITEMP H2O or HITEMP CH4. Ensure it fills your RAM memory. If it does not, find a larger molecule (ExoMol has many extremely large molecules), or find a weaker computer :)        Record calculation times
- [ ] Then, compare with the same calculation using chunksize, and compare calculation times.   You may have to adjust chunksize parameter to optimize it. 
- [ ] @anandxkumar had created a very Performance graph showing calculation time vs number of lines.   You'll want to produce the same kind of graph : without chunksize, I expect that beyond a certain threshold, calculation time skyrockets because of RAM saturation. With chunksize, it may be slightly slower at first, but never skyrockets. 

After implementation you'll also be able to : 
- [ ] add an "auto" mode which automatically compute the chunksize based on user available RAM (and setting it to None if it seems all lines will fit in RAM); This requires that you have some kind of correlation to know the memory size of a line chunk.

radis/test/lbl/test_broadening.py
```diff
@@ -938,6 +938,57 @@ def test_noneq_continuum(plot=False, verbose=2, warnings=True, *args, **kwargs):
     assert res < 5.2e-6
 
 
+def test_broadening_chunksize_eq(verbose=True, plot=False, *args, **kwargs):
+    """
+    Test an equilibrium spectrum with and without chunksize,
+    using different optimization strategies
+    and ensure that the residual is very small.
```
erwanp, line 945: Can you mention this PR (where chunksize was implemented) in the docstring ? Helps future devs

radis/test/lbl/test_broadening.py
```diff
@@ -938,6 +938,57 @@ def test_noneq_continuum(plot=False, verbose=2, warnings=True, *args, **kwargs):
...
+        neighbour_lines=5,
+        path_length=0.1,
+        mole_fraction=1e-3,
+        medium="vacuum",
+        optimization=None,
+        chunksize=None,
+        wstep=0.001,
+        verbose=False,
+    )
+    sf.load_databank("HITEMP-CO")
```
erwanp, line 964: Use `sf.fetch_databank("hitemp")`;  adding `molecule='CO'` in SpectrumFactory parameters
--> it will ensure it downloads the database if not there already

radis/test/lbl/test_broadening.py
```diff
@@ -938,6 +938,57 @@ def test_noneq_continuum(plot=False, verbose=2, warnings=True, *args, **kwargs):
...
+    # Testing all 6 cases of chunksize with optimization
+    for chunksize in [None, chunk_loop]:
+        sf.misc["chunksize"] = chunksize  # Setting chunksize
+        s_new = sf.eq_spectrum(Tgas=Tgas)
+
+        for optimization in [None, "simple", "min-RMS"]:
+            sf.params["optimization"] = optimization  # Setting optimization
+            s = sf.eq_spectrum(Tgas=Tgas)
+            # Residual calculation
+            res = get_residual(s, s_new, "abscoeff")
```
erwanp, line 978: /!\ Warning.

You are not testing Chunksize=None Vs chunksize!= None here, are you ? 

Seems both s_new and s are computed with the same chunksize parameter.

Can you check this, and also use more specific names for spectra (like s_no_chunksize; s_chunksize)

radis/test/lbl/test_broadening.py
```diff
@@ -938,6 +938,57 @@ def test_noneq_continuum(plot=False, verbose=2, warnings=True, *args, **kwargs):
...
+            assert res < 10e-5
+            print(
+                "Res for chunksize = {0}, optimization = {1} : {2}".format(
+                    chunksize, optimization, res
+                )
+            )
+            if plot:
+                plot_diff_list.append([s, s_new])
```
erwanp, line 986: suggestion : if plot, plot directly there instead of delaying to the end ?

erwanp (changes requested):
See my Warning. The rest are minor comments

radis/test/lbl/test_broadening.py
sagarchotalia: My thinking was that, if we plot after every calculation process it will hinder the rest of the calculations since the user will have to close the window each time, and if we plot at the end then they'll appear in succession, and it'll be good for comparison as well. But I can shift it above as well, if needed😄

radis/test/lbl/test_broadening.py
sagarchotalia: Hello, I tested this separately on my system, and the chunksize param being printed is correct

radis/test/lbl/test_broadening.py
erwanp: As I always run tests in interactive mode i don't have this problem

radis/test/lbl/test_broadening.py
erwanp: You do not compare Chunksize vs No-chunksize in this loop. 
s_new and s have the same chunksize value.

radis/test/lbl/test_broadening.py
sagarchotalia: Alright, then I'll move the plot lines to the loop. Thanks for the suggestion!

radis/test/lbl/test_broadening.py
sagarchotalia: Okay, I'll fix it. Thanks!

sagarchotalia: Hello @erwanp, can you check? I think it should work now.

radis/test/lbl/test_broadening.py
```diff
@@ -938,6 +938,59 @@ def test_noneq_continuum(plot=False, verbose=2, warnings=True, *args, **kwargs):
...
+    # Testing chunksizes with different optimization strategies
+    for optimization in [None, "simple", "min-RMS"]:
+        sf.params["optimization"] = optimization  # Setting optimization
+        sf.misc["chunksize"] = None
+
+        s_chunk = sf.eq_spectrum(Tgas=Tgas)
+
+        sf.misc["chunksize"] = 1e5  # Setting chunksize
+        s_no_chunk = sf.eq_spectrum(Tgas=Tgas)
+
```
erwanp, line 981:
```suggestion
        s_no_chunk = sf.eq_spectrum(Tgas=Tgas)
        assert s_no_chunk.c["chunksize"] is None  #make sure it was taken into account in Spectrum conditions
        
        sf.misc["chunksize"] = 1e5  # Setting chunksize
        s_chunk = sf.eq_spectrum(Tgas=Tgas)
        assert s_chunk.c["chunksize"] == 1e5

```

erwanp (changes requested):

radis/test/lbl/test_broadening.py
erwanp: Looks good

erwanp:
About the failing errors 

- for https://app.travis-ci.com/github/radis/radis/jobs/578898407#L2079  , `pinv2` is deprecated and removed in Scipy 1.9, replace with `scipy.linalg.pinv`

- for https://app.travis-ci.com/github/radis/radis/jobs/578898407#L1920, I think this was created by your changes.

erwanp (approved):

erwanp:
Looks good to me, merging this first major part of the GSOC project !

arunavabasu-03:
great @sagarchotalia 🎉

anandxkumar:
Congrats @sagarchotalia on the first major PR merge!

PR #490, opened 2022-06-09
Updated numpy<=1.22.3 and fixed linting

anandxkumar:
### Description
This pull request deprecates NumPy dependency to v 1.22.3 to Fix issue #486

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -15,7 +15,7 @@ dependencies:
 - h5py   # load HDF5
 - joblib  # for parallel loading of SpecDatabase
 - matplotlib
-- numpy
+- numpy<=1.22.3
 - numba  # just-in-time compiler
 - pandas>=1.0.5
 - plotly>=2.5.1  # for line survey HTML output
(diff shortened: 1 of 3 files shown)


anandxkumar:
@erwanp, the good thing is that the test passed, the bad thing is that linting failed (something similar to the old one), could you look into the lining issue?

erwanp:
Excellent! I have a fix. You can use the following comment on the problematic lines in GPU.py :
```
# noqa  # isort: skip
```

anandxkumar:
@erwanp still failing, I tried different versions of `isort` but no good, they organize the code in a different format which causes `pytest` to fail.

erwanp:
It's the `from radis_cython_extensions  ` like thats problematic and that you should also comment with isort :skip

(the reason is that it's either considered or not considered a part of the library by the local precommit  Travis-CI and therefore the ordering changes)

anandxkumar:
@erwanp used a different approach because `isort: skip` was not working. `isort: off` and `isort:on` skip all the lining for all the lines in between. I guess we can merge it now

erwanp (approved):
Congratulations ! I let you merge.

HajimeKawahara:
Hello! Do you still need [to limit numpy to <=1.22.3](https://github.com/radis/radis/blob/c9e7ff484f309e6fde8b2c80a72c0673af564254/setup.py#L256)?

erwanp:
Haven't tested recently, but this comment from one month ago seems to suggest that latest vaex is still incompatible with numpy >1.22.3   https://github.com/vaexio/vaex/issues/2062#issuecomment-1243072089

HajimeKawahara:
I see. Ah, this was from `vaex`. Thanks!

PR #492, opened 2022-07-02
Added code to add, retrieve and store extra params

Supriya1702:
This pull request is to address the task to add, retrieve and store the extra params for all (9) isotopes of a molecule(CO)

- I checked if extra_params is not empty then `Parameters = extra_params`  will be passed in fetch() method
(`extra_params = ['gamma_CO2',n_CO2']`  ----> we will pass this via fetch_hitran() only...for time being I've hard coded it)
 This will allow addition of these extra_params like gamma_CO2, n_CO2 etc in the `.data file` for that respective isotope (CO_1.data, CO_2.data....etc)
-  Then I am retrieving these column values using `getColumn` of hapi.
-  After that I stored this data into a global dataframe(temp_df) with the column name as  `CO_1_gamma_CO2`, `CO_1_n_CO2`..(molecule_isotope_param) to distinguish the isotopes columns.

Similarly, for the rest of the isotopes this `temp_df` will keep on appending column values.
 
- After retrieving and adding these columns to the dataframe, I have deleted these extra parameters from the `.data file`

 It helps in not hindering the normal workflow and to bypass the parsing methods which won't work for these extra_params added in `.data` anyways.It behaves like the normal workflow without passing extra_params in fetch() method. 
 
 Hence, we will get a consolidated dataframe with these extra parameters for all
isotopes for a molecule. I've attached a csv file of the final dataframe created.
 some values are empty since not all isotopes have exact same number of lines,
These empty values appear as NaN in dataframe, we can replace them with 0 or avg accordingly
[extra_params_CO.csv](https://github.com/radis/radis/files/9033738/extra_params_CO.csv)

 
Now, I need to add these extra_params in the final `CO.hdf5` file.
 
but first I want to know the approach to append these columns to every isotope hdf5 accordingly which is later on consolidated into a single .hdf5 file.
Below is a code snippet of hitran.py where I believe we need to integrate the 'temp_df' columns(final df having all columns)  
 
	    # Create HDF5 cache file for all isotopes
        Nlines = 0
        for iso, data_file in zip(isotope_list, data_file_list):
            df = hit2df(
                join(tempdir, data_file),
                cache=False,  # do not generate cache yet
                parse_quanta=parse_quanta,
            )
            wmin_final = min(wmin_final, df.wav.min())
            wmax_final = max(wmax_final, df.wav.max())
            Nlines += len(df)

            writer.write(
                local_file, df, append=True
            )  # create temporary files if required


Supriya1702:
Hello @erwanp @anandxkumar ,
this time I modified the code to create a consolidated `.hdf5` file for extra_params of a molecule. In this hdf5 file, all values are appended  isotope wise. For eg- if we're adding `gamma_CO2` and `n_CO2` for `CO` molecule then the `.hdf5 `will initially have values of `gamma_CO2` and `n_CO2` for `CO_1` (isotope 1), then values of `gamma_CO2` and `n_CO2` for `CO_2` (isotope 2) will be appended and so on.

Now, if we want to merge this `extra_params.hdf5` with the `molecule.hdf5`.....then the sequence of data should be same in both right? In my `extra_params.hdf5` I have appended values sequentially as per isotope. I wonder if the `molecule.hdf5` file is also created by sequentially appending every isotope data in the same order like isotope1,isotope2....so on.
If so, then we can merge these two. 
a couple of points here:
- would the sequence/order of these `extra_params (broadening coefficients)` in .hdf5 would affect the processing?
- Also, the count of rows is equal in both `.hdf5` files
- should I replace the `NaN` values of extra_params with `0`? would that affect the physics/formula?
- If I need to merge these two `.hdf5` files then how can I do it with `vaex` (tried searching a lot but couldn't find the solution)

radis/io/hitran.py
```diff
@@ -1160,7 +1166,49 @@ def download_all_hitran_isotopes(molecule, directory):
...
+                        
+                        ######delete the extra params comma separated from .data file
+                        import fileinput 
+                    
+                        #local_path = join(directory, file + ".data")
+                        with fileinput.FileInput(join(directory,file +".data"), inplace=True) as temp_file:
+                            for line in temp_file:
+                                split_string = line.split(",",1)
+                                line = split_string[0]
+                                print(line)
```
erwanp, line 1208: what does this block do ?

radis/io/hitran.py
```diff
@@ -1437,6 +1488,14 @@ def fetch_hitran(
         output=output,
     )
 
+    #creating consolidated .hdf5 file for extra_params of all isotopes of molecule 
+    if(len(extra_params) != 0):
+        import vaex
+        vaex_df = vaex.from_pandas(temp_df, copy_index=False)
+        vaex_df.export_hdf5(join(local_databases, molecule+"extra_params.hdf5"))
+
```
erwanp, line 1496: this is not the right place to create a file. Can't you make the file creation happen in `download_and_parse` ? 

Note 2 : line 1483 :  you'll probably need to ask to load extra columns , i.e. add the required parameters ("gamma_CO2", "n_CO2" )  to `columns`

radis/io/hitran.py
```diff
@@ -1160,7 +1166,49 @@ def download_all_hitran_isotopes(molecule, directory):
...
+                        
+                        from hapi import getColumn
+                        import numpy.ma as ma
+                        import pandas as pd
+                    
+                    
+                        file_params = {}  
+                        temp_dict = {}
+                        local_df = pd.DataFrame()
+                        global temp_df
```
erwanp, line 1181: global parameter is very bad practice in Python, find another way ?

erwanp (changes requested):

erwanp:
Answering your latest question : 

> "If I need to merge these two .hdf5 files then how can I do it with vaex (tried searching a lot but couldn't find the solution"

In Vaex you can simply open multiple files all together. See `vaex.open_many`

radis/io/hitran.py
Supriya1702: the delete block -
So, whenever call the fetch method with Parameters= ['gamma_CO2','n_CO2'] the `.data` file contains basic attributes along with the extra parameters passed in a comma-separated manner

when simply fetch() is called 
image


**when fetch() with Parameters** we get extra values comma separated 
image


So, these extra parameters are extracted using getColumn of HAPI and initially stored in a dataframe which is then converted in a .hdf5 file, this `delete block` basically removes the comma-separated values (which are the extra parameters) so that it does not hinder the normal workflow of parsing the basic attributes/columns.
I went ahead and created the .hdf5 file of extra parameters in this manner so that it doesn't hinder the normal processing of code. Otherwise, we would have to define a different way of getting these extra parameters

radis/io/hitran.py
Supriya1702: In today's commit, I have created the consolidated `.hdf5` file for `extra_params` of all isotopes of the molecule  in `download_all_isotopes` since we are downloading extra_params for all isotopes.

Also,  rather than loading  required parameters in `columns` 
I've added a new parameter in `extra_params` in `fetch_hitran()` method  which I also had to add to `class DatabaseManager` in `dbmanager.py`.
So, through `extra_params` a user can pass the broadening coefficients like 'gamma_CO2, n_CO2' in the `fetch_hitran() `method and a `.HDF5` file for those attributes will be created in the same location.
For instance,
if user is calling -
````
from radis.io.hitran import fetch_hitran
fetch_hitran("CO",["gamma_CO2","n_CO2"])
````
then a file `CO.hdf5 `will be created with the basic attributes and a file `CO_gamma_CO2_n_CO2.hdf5` will be created that has only `gamma_CO2` and `n_CO2` values for CO.

File: radis/io/hitran.py (hunk 3 above)
Comment (Supriya1702) [reply radis/io/hitran.py]: Yes, I've removed the use of global df in my latest commit

File: radis/io/hitran.py [hunk 4]
```diff
@@ -1078,6 +1073,7 @@ def __init__(
         self,
         name,
         molecule,
+        extra_params,
```
erwanp, line 1076: Shouldn't it has default value of None?

Supriya1702:
Hello @erwanp ,
Thanks for `vaex.open_many` method .....I verified using below code and it worked.
````
import vaex

df_vaex = vaex.open_many(['C:\\Users\\Supriya\\.radisdb\\hitran\\CO.hdf5','C:\\Users\\Supriya\\.radisdb\\hitran\\CO_gamma_CO2_n_CO2.hdf5'])

print(df_vaex)
`````
The final df contained all columns  `basic+extra`
image


Also, one more thing I need to confirm.....last time I had asked `would the sequence/order of these extra_params (broadening coefficients) in .hdf5 would affect the processing?` to which you had answered `No`.
I was actually referring to the rows wise sequence of  data of these extra_params  .hdf5 file.

I checked the basic attributes `CO.hdf5` file and in that the rows don't seem to be ordered - isotope 1 data rows and then isotope 2 data rows and so on, it's basically random like.....first 5 rows have isotope 5 data and then suddenly a single row of isotope 6 appears and somewhere in the middle isotope 1, 2, 3 are present but unordered.

But the `.hdf5` file that I created  for extra parameters has `ordered data isotope wise` (row wise) like initial rows are of isotope 1 data , then we have isotope 2 data and so on.

This eventually means that the 1st row of `CO_gamma_CO2_n_CO2.hdf5` file should align with `CO.hdf5` file 
but `CO.hdf5` 1st row has data of isotope 5 while `CO_gamma_CO2_n_CO2.hdf5`  1st row has data of isotope 1

would these row-wise sequences matter for both files or it would work this way?
If row-wise data matters then how exactly is the data in `CO.hdf5` ordered?

erwanp:
Hello ! Row-data (= line data) should definitly align. 

There is no easy way to identify a line; isotope won't be enough ; a good indication maybe the transition wavenumber `wav`  although many lines may have the same transition wavenumber (the only accurate way is to have parsed the local & global parameters giving infos such as Upper & lower vibrational & rotational quanta). 

CO.hdf5 is probably ordered by `wav`. 
do you at least have as many rows in both files CO.hdf5 and CO_gamma_CO2_n_CO2.hdf5 ?

Supriya1702:
outdated PR

PR #495, opened 2022-07-18
Non air diluent 2

Supriya1702:
This PR is to address the refactoring of  `radis.io.hitran.py` to use the `db_begin + read LOCAL_TABLE_CACHE `approach rather than use `hit2df` and update formula of pressure broadening to calculate `non-air diluent` broadening.
Following is the difference between lineshapes of CO molecule for different diluents:
<p align="center">

</p>

- I've replaced hit2df  with pd.DataFrame(LOCAL_TABLE_CACHE[data_file_name[0]]["data"]) and renamed the columns as you had mentioned in the proof of concept of how to get a Pandas dictionary from the LOCAL_TABLE_CACHE. 
- It works just fine!
- Also, by passing Parameters = ['gamma_CO2','n_CO2'] in fetch() the columns are loaded in LOCAL_TABLE_CACHE and is parsed together along with all other columns.
- Hence, now just by passing these extra columns as Parameters = ['gamma_CO2','n_CO2'] in fetch() function we obtain a single CO.hdf5 file in which basic  + extra Columns are present . ( All row data / line data are now aligned)
- The final df returned by fetch_hitran() also contains basic  + extra columns
- If for a given molecule a column is not present then it simply adds `NaN`  values to df and .hdf5 file for that column


erwanp:
Ok this is great ! The fact that missing values are handled without crash is good, too. 

Now comes the practical implementation details : 
- do you have a way to know all the possible extra parameters for a given molecule ?
- if not, which ones to download ? 
Also : 
- what if the database has already been downloaded (true for most current radis users) but we want to add new parameters? If you can only download selectively the new parameters, then we could append them to another file (siilar to the *_extra.hdf5 approach you had) . If you'll need to re-download everything anyway, then maybe it's just better to replace the existing database with the new one.   
- if replacing tge database, make sure that already downloaded parameters are still downloaded. Else you'll end up with users having to redownload the full database each time they'll change a gas composition and require first `gamma_CO2` then `gamma_H2O` for instance. Avoid this.

Supriya1702:
Hello @erwanp, I've replied on Slack

Supriya1702:
Hello @erwanp 
```
s = calc_spectrum(
                wavenum_min=1900,
                wavenum_max=2300,
                Tgas=700,
                path_length=0.1,
                molecule='CO',
                mole_fraction=0.2,
                isotope=1,
                broadening_method='convolve',
                wstep=0.01,
                databank='hitran'  , # or 'hitemp'
                use_cached='regen',
                diluent={'CO2':0.1, 'He':0.2, 'H2O': 0.2},
                export_lines=True,
            )

s.plot('transmittance_noslit')
```

- when we pass diluent={'CO2':0.1, 'He':0.2, 'H2O': 0.2}, the code is now calculating spectrum accordingly
- First we calculate the total mole fraction which is `mole fraction of molecule + mole fraction of diluents`
- if it's less than `1` then we assume `1-total_mole_fraction` is of air and incorporate air diluent as well in the formula

Some improvements to be made-

-  for loading all columns i.e. setting `load_columns = 'all'`  we need pass `export_lines = True`. This will be later automatically added with a condition when diluent has elements other than air then set` export_lines = True`.
- And for setting `use_cached = 'regen'` , we  will check if diluent columns are already present in db or not....if not present then set `use_cached = 'regen`

erwanp:
Hello @Supriya1702 ; lot's of new improvements, well done ! 


> Hello @erwanp
> 
> ```
> s = calc_spectrum(
>                 wavenum_min=1900,
>                 wavenum_max=2300,
>                 Tgas=700,
>                 path_length=0.1,
>                 molecule='CO',
>                 mole_fraction=0.2,
>                 isotope=1,
>                 broadening_method='convolve',               # 👉  you can use the default (remove this; unless there's a reason ?)
>                 wstep=0.01,               # 👉  you can use the default (remove this; unless there's a reason ?)
>                 databank='hitran'  , # or 'hitemp'               # 👉  "hitran" only for the momoemtn
>                 use_cached='regen',
>                 diluent={'CO2':0.1, 'He':0.2, 'H2O': 0.2},
>                 export_lines=True,
>             )
> 
> s.plot('transmittance_noslit')
> ```
> 
> * when we pass diluent={'CO2':0.1, 'He':0.2, 'H2O': 0.2}, the code is now calculating spectrum accordingly
> * First we calculate the total mole fraction which is `mole fraction of molecule + mole fraction of diluents`
> * if it's less than `1` then we assume `1-total_mole_fraction` is of air and incorporate air diluent as well in the formula

👉  we should not assume anything at this point. What about : 
- allow diluent=single species; i.e. `diluent='CO2',   diluent='Air'`   which automatically computes the right mole fractions  
- default diluent should be `diluent='Air'` 
- if explicit mole fractions have been given, raise an error if sum is total is not 1 . Keep in mind : _"better crash explicitely than do things secretly under the hood"_  .  Warning: RADIS can have mole_fractions has a dictionary (see [Multiple species spectrum](https://radis.readthedocs.io/en/latest/lbl/lbl.html#calculate-multiple-molecules-spectrum)). You'll have to handle that too. 
- (optional) in multi-species diluent ; allow a last, filling species. For instance: `diluent={'CO2':0.1, 'H2O':'rest'}`  

👉  you'll have to create tests for each of these configurations. In particular `pytest.raises(ValueError)` test, which expect an error. If you've never done back let's discuss it on Slack. 

> 
> Some improvements to be made-
> 
> * for loading all columns i.e. setting `load_columns = 'all'`  we need pass `export_lines = True`. This will be later automatically added with a condition when diluent has elements other than air then set` export_lines = True`.

👉  Nop; load_columns='all' will saturate memory for larger databases. Other GSOC projects have been working on only loading the columns required, let's not reverse this. Can you instead try to : 
- have a mode where you load all and only the required parameters; i.e. `n_*` and `gamma_*` 


> * And for setting `use_cached = 'regen'` , we  will check if diluent columns are already present in db or not....if not present then set `use_cached = 'regen`


👉  the largest cached databases take up to hours to build; we should not regen then automatically. Again :  _"better crash explicitely than do things secretly under the hood"_  . There are two configurations : 
- we know the parameter exists; but it is not presently downloaded. Raise an error and suggest the user to (1) ignore the error (2) regen the database by setting `use_cached = 'regen'` (user should do it himself, in the script)
- we know the parameters does not exist --> just raise an warning , for instance similar to [MissingSelfBroadeningWarning](https://github.com/radis/radis/blob/develop/radis/misc/warning.py#L136)
- if we don't know --> assume 2nd case ?

radis/lbl/calc.py
```diff
@@ -601,6 +607,7 @@ def _calc_spectrum_one_molecule(
     mode,
     export_lines,
     return_factory=False,
+    diluent={},
```
erwanp, line 610: don't forget description in Docstrings. 
Also, you can move this parameter just below mole_fractions

radis/lbl/factory.py
```diff
@@ -1895,6 +1913,28 @@ def _generate_wavenumber_arrays(self):
...
+
+        # Checking mole_fraction of molecule and diluent
+        total_mole_fraction = mole_fraction + sum(list(diluents.values()))
+        if total_mole_fraction > 1:
+            raise ValueError(
+                "Total molefraction = {0} of molecule and diluents greater than 1.".format(
+                    total_mole_fraction
+                )
+            )
+        elif total_mole_fraction < 1 and "air" not in diluents.keys():
```
erwanp, line 1927: Good in principle, but it doesn't fit the Radis principle of "better raise an error than do something behind your back" 

I'd rather : 
- crash if total_mole_fractions are <  1  (even, if "air" is given) 
- use the error message to suggest to use a filler; for instance "Air"     (syntax `diluent='air'` or `diluent='{"species_1":mole fractoin1, "species_2" : mole_fraction2, "species_3": "fill"`  ) 
- set-up `diluent="air"` by default

radis/lbl/factory.py
```diff
@@ -648,7 +659,12 @@ def __init__(
     # XXX =====================================================================
 
     def eq_spectrum(
-        self, Tgas, mole_fraction=None, path_length=None, pressure=None, name=None
+        self,
+        Tgas,
+        mole_fraction=None,
+        path_length=None,
```
erwanp, line 665: you'll want `diluent` here too   (and in non_eq_spectrum() )  too

radis/lbl/broadening.py
```diff
@@ -248,7 +248,16 @@ def gaussian_FT(w_centered, hwhm):
 
 
 def pressure_broadening_HWHM(
-    airbrd, selbrd, Tdpair, Tdpsel, pressure_atm, mole_fraction, Tgas, Tref
+    airbrd,
```
erwanp, line 251: it's good if this function can be independant of a DataFrame; so it can be used & tested independantly from the rest of the RADIS code
(e.g., to examplify how Radis works / or to validate it,  we sometimes have a step-by-step calculations, where people can call the  `pressure_broadening_HWHM` function on simple numpy arrays)

Can you think of a way to implement this ? 

An idea could be to use a dictionary to combine `diluent` and the columns you need from `df` (making the dictionary would be done in Radis main code by the function calling pressure_broadening_HWHM() ) 

The dictionary could even include "self" and "air" (at this point there is no difference between them or other collisioniers)

Default behavior :
```
pressure_broadening_HWHM(pressure_atm, Tgas, Tref, {"self": (mole_fraction, broadening_coefficient[i.e. selfbrd], temperature_dependance_coefficient[i.e. Tdpsel ) })
```
With air : 
```
pressure_broadening_HWHM(pressure_atm, Tgas, Tref, {"self": (mole_fraction, selfbrd, Tdpsel), "air":(mole_fraction, airbrd, Tdpair) })``` 
```

ARbitrary number of species: 
```
pressure_broadening_HWHM(pressure_atm, Tgas, Tref, {"self": (mole_fraction, selfbrd, Tdpsel), "air":(mole_fraction_air, airbrd, Tdpair) }, "species1":(mole_fraction_1, airbrd, Tdpair)``` 
```
Note : 
- you should handle cases where the last coeficnet, ( Tdpair / Tdpsel / Tdp_anyspecies) is None 
- you can add a new warning in this function if mole_fraction_sum != 1 , but don't raise an error or don't suggest anything like using "diluent=..." . Remember this function may be used/called independantly of RADIS 

THis function will become much simpler (therefore easier to test / understand); and the complexity of building the dictionary will be added in the calling RADIS functions.

erwanp (changes requested):
Hello, many comments; but overall you've really well understood the RADIS architecture and where to make the changes, well done !

radis/lbl/loader.py
```diff
@@ -1159,7 +1201,7 @@ def fetch_databank(
                     extra_params=extra_params,
                 )
                 self.params.dbpath = ",".join(local_paths)
-
+                print("FETCH HITRAN:", df)
```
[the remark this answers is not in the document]
Supriya1702: will remove it

radis/io/hitran.py
```diff
@@ -1177,22 +1189,41 @@ def download_all_hitran_isotopes(molecule, directory):
...
         writer = self.get_datafile_manager()
 
         # Create HDF5 cache file for all isotopes
         Nlines = 0
         for iso, data_file in zip(isotope_list, data_file_list):
-            df = hit2df(
-                join(tempdir, data_file),
-                cache=False,  # do not generate cache yet
-                parse_quanta=parse_quanta,
+            df = data_file_name = data_file.split(".")
```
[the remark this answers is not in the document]
Supriya1702: This is outdated code. It's no longer there

radis/lbl/broadening.py
Supriya1702: I tried replacing df with a dictionary diluent_broadening_coeff that contains "broadening_coeff" : value  and hence
now pressure_broadening_HWHM  can be independent of df

radis/lbl/factory.py
Supriya1702: I've added diluent in eq_spectrum() and non_eq_sprectrum() definition

anandxkumar:
@erwanp is there any test for `cache='regen'`, if yes then I guess we can go ahead with a smaller database like `OH` or `CO` and regen it and test the non-air diluent spectrum, else we can have a separate data in `.h5` format with certain lines and test the same.

radis/lbl/calc.py
```diff
@@ -48,7 +48,7 @@ def calc_spectrum(
     molecule=None,
     isotope="all",
     mole_fraction=1,
-    diluent={},
+    diluent=None,
```
erwanp, line 51: Should be "air"

erwanp (changes requested):

radis/lbl/calc.py
Supriya1702: Done. Now we can add a single diluent as string in the `diluent` argument, by default `diluent='air'`, or can also be `diluent='CO2'`, or any other molecule. In this case we will automatically calculate the molefraction of diluent using 1-molefraction. In other cases user will be required to pass on the diluent as dictionary with it's molefraction.

erwanp (changes requested):

radis/test/lbl/test_calc.py
```diff
@@ -819,6 +819,53 @@ def test_check_wavelength_range(verbose=True, warnings=True, *args, **kwargs):
...
+        wstep=0.01,
+        databank="hitran",
+        Tvib=710,
+        Trot=710,
+    )
+
+    assert s1.get_conditions()["diluents"] == {"CO2": 0.2, "H2O": 0.1, "air": 0.5}
+    assert s2.get_conditions()["diluents"] == {"air": 0.8}
+
+    plot_diff(s1, s2, method=["diff", "ratio"])
```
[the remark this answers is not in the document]
erwanp: Also, can you add a screenshot of the plot (few lines only) in the core text of this PR ? name the spectra according to the diluent , it should add it in the legend
```
calc_spectrum(..., name="diluent CO2:0.2, H2O:0.1, air:0.5"}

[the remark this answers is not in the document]
Comment (Supriya1702) [reply radis/test/lbl/test_calc.py:865]: Done!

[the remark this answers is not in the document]
Comment (erwanp) [reply radis/test/lbl/test_calc.py:865]: Hello, nice!

However the difference is not very visible under these conditions. Can you try with another mixture? Maybe half CO2 half H2O?

File: radis/test/lbl/test_calc.py [hunk 9]
```diff
@@ -819,6 +819,53 @@ def test_check_wavelength_range(verbose=True, warnings=True, *args, **kwargs):
     return True
 
 
+def test_non_air_diluent_calc(verbose=True, warnings=True, *args, **kwargs):
```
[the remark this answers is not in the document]
Supriya1702: Hello @erwanp, I have updated the tests, and add added the following:
In `test_factory.py`:

- For a SpectrumFactory I'm generating 3 spectrums for different diluents and molefraction and comparing their `hwhm_lorentz` and check whether non_air diluent increases HWHM or not `(hwhm1< hwh2).all()`
- Checking mole fraction error. I created a new class `MoleFractionError `in `warnings.py`. If sum of diluents and molecule molefractions > 1 or < 1 using pytest.raise I'm checking that error and using str(err.value) to assert the same.

In `test_calc.py`:

- First calculating spectrums for air and non air diluent, and comparing their hwhm_voigt_s2 using assert.

radis/lbl/calc.py
```diff
@@ -399,6 +413,12 @@ def calc_spectrum(
     # (dealt with separately because we cannot use them to guess what are the input molecules)
     DICT_INPUT_DICT_ARGUMENTS = {"overpopulation": overpopulation}
 
+    # Check if mole_fraction is dictionary of multiple molecules and have non air diluent, if yes then raise NotImplementedError
+    if isinstance(mole_fraction, dict) and isinstance(diluent, dict):
+        raise NotImplementedError(
+            "Spectrum calculation for multiple molecules in multiple diluents is not yet implemented."
+        )
```
anandxkumar, line 420: Good thing you took care of it!

radis/lbl/broadening.py
```diff
@@ -354,7 +354,7 @@ def pressure_broadening_HWHM(
                 (selbrd * pressure_atm * mole_fraction)
             )
         else:
-            gamma_lb += gamma_lb + ((Tref / Tgas) ** Tdpsel) * (
+            gamma_lb += ((Tref / Tgas) ** Tdpsel) * (
```
anandxkumar, line 357: This one was a scary mistake, glad you noticed it!

anandxkumar:
Great job @Supriya1702, looks like the physics and implementation is in the right place. The code looks good to me!

anandxkumar (approved):

radis/lbl/calc.py
```diff
@@ -129,6 +132,17 @@ def calc_spectrum(
 
             mole_fraction={'CO2': 0.8, 'CO':0.2}​
 
+    diluent: str or dictionary
+            can be a string of a single diluent or a dictionary containing diluent
+            name as key and its mole_fraction as value
+            For single diluent ::
+
+            diluent = 'CO2'
```
erwanp, line 140:
```suggestion
                diluent = 'CO2'
```
I think you need an offset so it appears as code

radis/lbl/calc.py
```diff
@@ -129,6 +132,17 @@ def calc_spectrum(
...
+
+            For multiple diluents ::
+
+            diluent = { 'CO2': 0.6, 'H2O':0.2}
```
erwanp, line 144:
```suggestion
                diluent = { 'CO2': 0.6, 'H2O':0.2}
```

radis/lbl/calc.py
```diff
@@ -18,6 +18,8 @@
 
 
 from copy import deepcopy
+
+# from functools import cache
```
erwanp, line 22: You can remove this

radis/lbl/broadening.py
```diff
@@ -314,16 +328,47 @@ def pressure_broadening_HWHM(
...
+                (selbrd * pressure_atm * mole_fraction)
+            )
+        else:
+            gamma_lb += ((Tref / Tgas) ** Tdpsel) * (
+                selbrd * pressure_atm * mole_fraction
+            )
+    except KeyError as err:
+        raise KeyError("Column not found {0}".format(err))
+
+    # if Tdpsel is None:
```
erwanp, line 363: You can remove this commented block

radis/lbl/broadening.py
```diff
@@ -314,16 +328,47 @@ def pressure_broadening_HWHM(
...
+                    * pressure_atm
+                    * diluent_mole_fraction
+                )
+        # Adding air coefficient
+        if "air" in diluent_molecules:
+            gamma_lb += ((Tref / Tgas) ** Tdpair) * (
+                (airbrd * pressure_atm * diluent["air"])
+            )
+        # Adding self coefficient
+        if Tdpsel is None:
```
erwanp, line 352:
```suggestion
        if Tdpsel is None:   # use Tdpair instead 
```
Adding comment

Supriya1702:
@erwanp I have incorporated you suggestions, let me know if anything remains

radis/lbl/calc.py
Supriya1702: Done in https://github.com/radis/radis/issues/525

anandxkumar:
I believe we can merge the PR, once @erwanp approves it, you can merge it yourself!

erwanp (changes requested):
Hello @Supriya1702 . Last few changes if you can find time for this. 

For all of the new test cases you've added, can you add a Docstring, describing in a few sentences of plain text what you're testing (for instance : "Here we ensure that non-air collisioner is taken into account by comparing the linewidth to a hardcoded value, and also making sure that it is different from the linewidth with air collisioner " ) 

Also add the link to this PR (https://github.com/radis/radis/pull/495) in the test, it helps trace back to our conversations here, useful for future debugging )

erwanp:
@Supriya1702 I'm merging, but you can add the docstri'gs etc in a new PR

PR #500, opened 2022-08-13
Fixing the case when the chunksize is too small

sagarchotalia:
### Description

This pull request is to address the case of small chunksize as mentioned by @erwanp. 
When the input chunksize is too small, the intermediate chunks(dg) can be of length equal to 1, resulting in a huge number of iterations.
Here, we assert that the number of groups should be less than the size of the dataframe(df), and raise a PerformanceWarning otherwise.

Related to #489

### DIFF
--- a/radis/lbl/broadening.py
+++ b/radis/lbl/broadening.py
@@ -2255,10 +2255,14 @@ class BroadenFactory(BaseFactory):
                     assert N < len(df)
                 except AssertionError:
                     self.warn(
-                        "Chunksize is currently too small according"
-                        + " to the dataframe size. Please set a larger"
-                        + " value of chunksize.",
-                        "PerformanceWarning",
+                        "There are currently more chunks ({0:.3e}) than lines ({1:.3e}),".format(
+                            N, len(df)
+                        )
+                        + " calculation times will be extremely slow. Try to have at least 10,000 - 100,000"
+                        + " lines per chunk. We suggest increasing chunksize to"
+                        + " {0:.1e} - {1:.1e}".format(
+                            10000 * len(wavenumber), 100000 * len(wavenumber)
+                        ),
                     )
 
                 abscoeff = zeros_like(self.wavenumber)
--- a/radis/lbl/broadening.py
+++ b/radis/lbl/broadening.py
@@ -2248,6 +2248,19 @@ class BroadenFactory(BaseFactory):
                 N = int(len(df) * len(wavenumber) / chunksize) + 1
                 # Too big may be faster but overload memory.
                 # See Performance for more information
+
+                # Raise performance warning if the chunks are bigger
+                # than the number of lines.
+                try:
+                    assert N < len(df)
+                except AssertionError:
+                    self.warn(
+                        "Chunksize is currently too small according"
+                        + " to the dataframe size. Please set a larger"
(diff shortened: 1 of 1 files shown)


radis/lbl/broadening.py
```diff
@@ -2248,6 +2248,19 @@ def _broaden_lines(self, df):
...
+
+                # Raise performance warning if the chunks are bigger
+                # than the number of lines.
+                try:
+                    assert N < len(df)
+                except AssertionError:
+                    self.warn(
+                        "Chunksize is currently too small according"
+                        + " to the dataframe size. Please set a larger"
+                        + " value of chunksize.",
```
erwanp, line 2260: you don't say why, it's not very clear. 
Can you try something like  
```
"There are currently more chunks ({N:.3e}) than lines ({len(df}:.3e), calculation times will be extremelly slow. Try to have at least 10,000 - 100,000 lines per chunk. We suggest increasing chunksize to {10000*len(wavenumber):.1e} - {100000*len(wavenumber):.1e}" 
```
Note : numbers are kinda guesses here. Check it works.

erwanp (changes requested):

radis/lbl/broadening.py
sagarchotalia: Hello! Apologies for not making it descriptive. I tried the above with CO, CO2, N2O and NO molecules, and it works well!

radis/lbl/broadening.py
```diff
@@ -2248,6 +2248,23 @@ def _broaden_lines(self, df):
                 N = int(len(df) * len(wavenumber) / chunksize) + 1
                 # Too big may be faster but overload memory.
                 # See Performance for more information
+
+                # Raise performance warning if the chunks are bigger
+                # than the number of lines.
+                try:
+                    assert N < len(df)
+                except AssertionError:
```
erwanp, line 2256: what do you think of an `if` syntax instead ? Seems more compact to me

erwanp (approved):

radis/lbl/broadening.py
sagarchotalia: Yes, I agree that it would be more compact. I'll change the syntax to if-raise

radis/lbl/broadening.py
```diff
@@ -2251,9 +2251,7 @@ def _broaden_lines(self, df):
 
                 # Raise performance warning if the chunks are bigger
                 # than the number of lines.
-                try:
-                    assert N < len(df)
-                except AssertionError:
+                if N >= len(df):
                     self.warn(
                         "There are currently more chunks ({0:.3e}) than lines ({1:.3e}),".format(
```
erwanp, line 2256: Here the message is not very helpful to the user. They'll say "Sure, so what?" 
Suggest to increase chunksize to a reasonable value (similar to the other comment)

erwanp (changes requested):

radis/lbl/broadening.py
sagarchotalia: Hello, the following lines have the same message as the previous commit, should I add something else too?

erwanp: yes

erwanp (approved):

PR #503, opened 2022-08-14
update chunksize test to make it faster - and RT minor perf improvement

erwanp:
@sagarchotalia  I realized the test implemented in #488  was extremelly slow on Travis-CI  (few minutes)
Here is a small update to make it faster (~10s). 

Please review.
Please also comment if needed, it's a good way to learn more about the codebase. 

I'll ask you to update this PR and also make sure that the Number of chunks in the "chunksize" mode is different from 1, else we're actually computing all lines together and not testing correctly. 

---

Other minor changes : 

- add config parameter "RESAMPLING_TOLERANCE_THRESHOLD" to deal with resampling errors globally
- add copy_array=False option in s.take(), improve MergeSlabs performance by using it

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,8 @@ jobs:
         # Your test script goes here
         - echo ">>> Run tests"
         # use XVFB to have headless display port, and still run the Matplotlib tests.
-        - xvfb-run -a pytest --cov=./
+        - xvfb-run -a pytest --cov=./ --durations=10
+        # --durations=N to print the slowest N tests
         # lookup 'addopts' in setup.cfg>[tools:pytest] for default tests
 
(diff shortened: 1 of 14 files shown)


sagarchotalia (approved):
Hello, this looks great to me! I'll add the number of chunks parameter by explicitly calculating the "N" parameter in the code and `asserting that N>1`, if that's okay

erwanp:
In production, having N=1 (i.e. all lines processed at once) is not necessarily a problem; so it should remained allowed.

But in the test it's very important we're having N>1. You could store N as a hidden variable in the SpectrumFactory;  sf._Nchunks for instance, and evaluate this in the test()

PR #505, opened 2022-08-15
Fetch hitran update

Supriya1702:
This PR addresses-

- Replacing hit2df with LOCAL_TABLE_CACHE 
- added a parameter extra_params in fetch_hitran() and dbmanager.py
- downloading all columns when extra_params = 'all'
- Adding conditions to raise AccuracyWarning when All columns are not downloaded currently.
- cleaning of NaN value columns after downloading all columns in hdf5.py

### DIFF
--- a/radis/io/dbmanager.py
+++ b/radis/io/dbmanager.py
@@ -517,25 +517,8 @@ class DatabaseManager(object):
         return nrows
 
     def get_columns(self, local_file):
-        """Get all columns (without loading all Dataframe)"""
-        engine = self.engine
-        local_file = expanduser(local_file)
-        if engine == "vaex":
-            import vaex
-
-            # by default vaex does not load everything
-            df = vaex.open(local_file)
-            columns = df.columns
-            df.close()
-        elif engine == "pytables":
-            with pd.HDFStore(local_file, "r") as store:
-                columns = store.select("df", start=1, stop=1).columns
-        elif engine in ["h5py"]:
-            raise NotImplementedError
-        else:
-            raise ValueError(engine)
-
-        return columns
+        """Get all columns using DataFileManager class get_columns function"""
+        return self.get_datafile_manager().get_columns(local_file)
 
     def add_column(self, df, key, value):
         """Create column ``key`` in DataFrame or dictionary ``df`` with value ``value``"""
--- a/radis/io/dbmanager.py
+++ b/radis/io/dbmanager.py
@@ -91,6 +91,7 @@ class DatabaseManager(object):
         molecule,
         local_databases,
         engine,
+        extra_params=None,
         verbose=False,
         parallel=True,
         nJobs=-2,
@@ -118,6 +119,7 @@ class DatabaseManager(object):
         self.name = name
         self.molecule = molecule
         self.local_databases = local_databases
+        self.extra_params = extra_params
         # create folder if needed
         if not exists(local_databases):
             from radis.misc.basics import make_folders
--- a/radis/io/dbmanager.py
+++ b/radis/io/dbmanager.py
@@ -517,7 +517,7 @@ class DatabaseManager(object):
         return nrows
 
     def get_columns(self, local_file):
-        """Get all columns (without loading all Dataframe"""
+        """Get all columns (without loading all Dataframe)"""
         engine = self.engine
         local_file = expanduser(local_file)
         if engine == "vaex":
--- a/radis/io/dbmanager.py
+++ b/radis/io/dbmanager.py
@@ -516,6 +516,27 @@ class DatabaseManager(object):
             raise ValueError(engine)
         return nrows
 
+    def get_columns(self, local_file):
+        """Get all columns (without loading all Dataframe"""
+        engine = self.engine
+        local_file = expanduser(local_file)
+        if engine == "vaex":
+            import vaex
+
+            # by default vaex does not load everything
+            df = vaex.open(local_file)
+            columns = df.columns
+            df.close()
+        elif engine == "pytables":
+            with pd.HDFStore(local_file, "r") as store:
+                columns = store.select("df", start=1, stop=1).columns
+        elif engine in ["h5py"]:
+            raise NotImplementedError
+        else:
+            raise ValueError(engine)
+
+        return columns
+
     def add_column(self, df, key, value):
         """Create column ``key`` in DataFrame or dictionary ``df`` with value ``value``"""
         from radis.misc.basics import is_number
--- a/radis/io/hdf5.py
+++ b/radis/io/hdf5.py
@@ -196,6 +196,28 @@ class DataFileManager(object):
             raise NotImplementedError(self.engine)
             # h5py is not designed to write Pandas DataFrames
 
+    def get_columns(self, local_file):
+        """Get all columns (without loading all Dataframe)"""
+        engine = self.engine
+        local_file = expanduser(local_file)
+        if engine == "vaex":
+            import vaex
+
+            # by default vaex does not load everything
+            df = vaex.open(local_file)
+            columns = df.columns
+            df.close()
+
+        elif engine == "pytables":
+            with pd.HDFStore(local_file, "r") as store:
+                columns = store.select("df", start=1, stop=1).columns
+        elif engine in ["h5py"]:
+            raise NotImplementedError
+        else:
+            raise ValueError(engine)
+
+        return columns
+
     def combine_temp_batch_files(
         self, file, key="default", sort_values=None, delete_nan_columns=True
     ):
--- a/radis/io/hdf5.py
+++ b/radis/io/hdf5.py
@@ -213,6 +213,14 @@ class DataFileManager(object):
             import vaex
 
             df = vaex.open(self._temp_batch_files, group=key)
+            # Removing Nan values columns
+            import numpy as np
+
+            for column in df.columns:
+                col = df[column].values
+                if type(col[0]) in [np.int32, np.float64] and np.isnan(np.sum(col)):
+                    del df[column]
+
             if sort_values:
                 df.sort(by=sort_values).export_hdf5(file, group=key, mode="w")
             else:
--- a/radis/io/hdf5.py
+++ b/radis/io/hdf5.py
@@ -196,7 +196,9 @@ class DataFileManager(object):
             raise NotImplementedError(self.engine)
             # h5py is not designed to write Pandas DataFrames
 
-    def combine_temp_batch_files(self, file, key="default", sort_values=None):
+    def combine_temp_batch_files(
+        self, file, key="default", sort_values=None, delete_nan_columns=True
+    ):
         """Combine all batch files in ``self._temp_batch_files`` into one.
         Removes all batch files.
         """
@@ -214,12 +216,13 @@ class DataFileManager(object):
 
             df = vaex.open(self._temp_batch_files, group=key)
             # Removing Nan values columns
-            import numpy as np
+            if delete_nan_columns:
+                import numpy as np
 
-            for column in df.columns:
-                col = df[column].values
-                if type(col[0]) in [np.int32, np.float64] and np.isnan(np.sum(col)):
-                    del df[column]
+                for column in df.columns:
+                    col = df[column].values
+                    if type(col[0]) in [np.int32, np.float64] and np.isnan(np.sum(col)):
+                        del df[column]
 
             if sort_values:
                 df.sort(by=sort_values).export_hdf5(file, group=key, mode="w")
--- a/radis/io/hitran.py
+++ b/radis/io/hitran.py
@@ -1129,7 +1129,6 @@ class HITRANDatabaseManager(DatabaseManager):
 
         from hapi import LOCAL_TABLE_CACHE, db_begin, fetch
 
-        # from radis import hit2df
         from radis.db.classes import get_molecule_identifier
 
         if isinstance(local_file, list):
@@ -1468,26 +1467,25 @@ def fetch_hitran(
         ldb.remove_local_files(local_file)
     else:
         # Raising AccuracyWarning if local_file exists and doesn't have extra columns in it
-        if ldb.get_existing_files(local_file):
-            import vaex
+        if ldb.get_existing_files(local_file) and extra_params == "all":
 
-            df = vaex.open(local_file)
-            extra = ["y_", "gamma_", "n_"]
+            columns = ldb.get_columns(local_file[0])
+            extra_columns = ["y_", "gamma_", "n_"]
             found = False
-            for key in extra:
-                for column_name in df.columns:
+            for key in extra_columns:
+                for column_name in columns:
                     if key in column_name:
                         found = True
                         break
+
             if not found:
                 import warnings
 
                 warnings.warn(
                     AccuracyWarning(
-                        "All columns are not downloaded currently, if want to compute spectrum in non-air diluent, please use cache = 'regen' and extra_params='all'"
+                        "All columns are not downloaded currently, please use cache = 'regen' and extra_params='all' to download all columns."
                     )
                 )
-            df.close()
 
     ldb.check_deprecated_files(
         ldb.get_existing_files(local_file),
--- a/radis/io/hitran.py
+++ b/radis/io/hitran.py
@@ -1115,7 +1115,7 @@ class HITRANDatabaseManager(DatabaseManager):
         opener: an opener with an .open() command
         gfile : file handler. Filename: for info"""
 
-        from hapi import db_begin, fetch, LOCAL_TABLE_CACHE
+        from hapi import LOCAL_TABLE_CACHE, db_begin, fetch
 
         from radis import hit2df
         from radis.db.classes import get_molecule_identifier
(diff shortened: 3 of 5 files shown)


radis/io/hitran.py
```diff
@@ -1113,9 +1127,9 @@ def download_and_parse(self, local_file, cache=True, parse_quanta=True):
         opener: an opener with an .open() command
         gfile : file handler. Filename: for info"""
 
-        from hapi import db_begin, fetch
+        from hapi import LOCAL_TABLE_CACHE, db_begin, fetch
 
-        from radis import hit2df
+        # from radis import hit2df
```
erwanp, line 1132: You can remove the line entirely

radis/io/hitran.py
```diff
@@ -1160,10 +1174,31 @@ def download_all_hitran_isotopes(molecule, directory):
...
+                        fetch(file, get_molecule_identifier(molecule), iso, wmin, wmax)
+                    else:
+                        raise ValueError("extra_params can only be 'all' or None ")
+                except KeyError as err:
+                    list_pattern = ["(", ",", ")"]
+                    import re
+
+                    if (
+                        set(list_pattern).issubset(set(str(err)))
+                        and len(re.findall("\d", str(err))) >= 2
```
erwanp, line 1196: Thats not a very explicit error? Do you know what's the typical error message when a isotope is not found?

radis/io/hitran.py
```diff
@@ -1285,6 +1340,7 @@ def register(self):
 # TODO: implement parallel=True for all isotopes ?
 def fetch_hitran(
     molecule,
+    extra_params=None,
```
erwanp, line 1343: Add a description in docstrings.

Also can you add a quick example showing how to use it?

radis/io/hitran.py
```diff
@@ -1409,6 +1466,29 @@ def fetch_hitran(
     # Delete files if needed:
     if cache == "regen":
         ldb.remove_local_files(local_file)
+    else:
+        # Raising AccuracyWarning if local_file exists and doesn't have extra columns in it
+        if ldb.get_existing_files(local_file):
+            import vaex
```
erwanp, line 1472: That's the right thing to do, unfortunately we do not use Vaex all the time. It's the new default but on some systems we still rely on Pandas, and it may change in the future.
That's why Radis includes a Hdf5Manager class, to abstractize the file format / reading library 

In your case I think you'll need to write a get_columns() function that will work with at least Vaex and Pandas 

For Vaex its exactly what you wrote 
For pandas it's more tricky. If you open the database it will actually load it in memory, thus take a lot of time and memory. But you can read columns without loading everything. See for instance : https://stackoverflow.com/questions/49401464/hdfstore-get-column-names/56403217#56403217

If format is not Pandas or Vaex just read a NotImplementeferror

Implementation : 

Get_columns could be a method of the DatabaseManager class (parent of HITRAN Database Manager above).
Use : 
```
Ldb.get_columns()
```
Docstring : returns list of columns in the database without reading the full database in memory . If many files, ensure that all files have the same columns names, or raise an error.




It's a new part of the codebase for you so ask questions if needed

radis/io/hitran.py
```diff
@@ -1409,6 +1466,29 @@ def fetch_hitran(
...
+
+            df = vaex.open(local_file)
+            extra = ["y_", "gamma_", "n_"]
```
erwanp, line 1475: Why don't you compare with extra_params only?

And have a special case if extra_params='all'

erwanp (changes requested):
Many suggestions above.

Also tests are failing in all nonequilibrium calculations. You can test locally by computing a Spectrum with Tvib different from Trot

radis/io/hitran.py
Supriya1702: sure

radis/io/hitran.py
Supriya1702: I tried printing the err and the typical message that comes is  => ("molecule_id","isotope_id") . for instance-> `(5,7)` then `(5,8)` and then `(5,9)` in the loop in case of CO.

radis/io/hitran.py
```diff
@@ -340,6 +340,239 @@ def hit2df(
...
+
+    See Also
+    --------
+
+    :func:`~radis.io.cdsd.cdsd2df`
+    """
+    from hapi import LOCAL_TABLE_CACHE
+
+    data_file_name = data_file.split(".")
+    df = pd.DataFrame(LOCAL_TABLE_CACHE[data_file_name[0]]["data"])
```
erwanp, line 411: isn't `data_file`   `fname` here ? 

If it's different add docstrings for data_file

Also, the line `data_file_name = data_file.split(".")` acn be better handled by os.path or pathlib.

erwanp (changes requested):

radis/io/hitran.py
erwanp: Ok . Then add this sentence in the docstrings (the regex are cryptic)  .
You can also collect the 2 integers; and check that the first one corresponds to the expected molecule_id
This we're making sure it's an isotope Key error and not something else

use `get_molecule_identifier`: 
```python
from radis.db.classes import get_molecule_identifier 

# ex : 
# >>> get_molecule_identifier ("CO")  
# 5
```

radis/io/hitran.py
```diff
@@ -407,8 +406,7 @@ def hit2df_hapi(
     """
     from hapi import LOCAL_TABLE_CACHE
 
-    data_file_name = data_file.split(".")
-    df = pd.DataFrame(LOCAL_TABLE_CACHE[data_file_name[0]]["data"])
+    df = pd.DataFrame(LOCAL_TABLE_CACHE[os.path.basename(fname).split(".")[0]]["data"])
```
erwanp, line 409: good. I was thinking of `os.path.splitext` , doesit work ?

radis/io/hitran.py
```diff
@@ -1379,6 +1377,9 @@ def download_all_hitran_isotopes(molecule, directory, extra_params):
             .. warning::
                 this won't be able to download higher isotopes (ex : isotope 10-11-12 for CO2)
                 Neglected for the moment, they're irrelevant for most calculations anyway
+            .. Isotope Missing:
+                When an isotope is missing for a particular molecule then a key error `(molecule_id, isotope_id)
+                is raised. So the regex are cryptic.
```
erwanp, line 1382: This should be a comment in the code; where your use your regex to extract the integers
The sentence "so the regex are cryptic" is not required ;)

radis/io/hitran.py
```diff
@@ -1594,6 +1596,14 @@ def fetch_hitran(
...
     columns: list of str
         list of columns to load. If ``None``, returns all columns in the file.
+    extra_params: 'all' or None
+        Downloads all additional columns available in the HAPI database for the molecule including
+        parameters like `gamma_co2`, `n_co2` that are required to calculate spectrum in co2 diluent.
+        For eg:
+        ```
+        from radis.io.hitran import fetch_hitran
+        df = fetch_hitran('CO', extra_params='all', cache='regen') # cache='regen' to regenerate new database with additional columns
+        ```
```
erwanp, line 1606: Doctrings are in RST not Markdown. Correction below (you can commit directly) 
```suggestion
        ::
        
            from radis.io.hitran import fetch_hitran
            df = fetch_hitran('CO', extra_params='all', cache='regen') # cache='regen' to regenerate new database with additional columns
        ```
```

erwanp (changes requested):
minor comments

radis/io/hitran.py
Supriya1702: Yes, have done the changes in the recent commit

radis/io/hitran.py
Supriya1702: done

radis/io/hitran.py
Supriya1702: done

radis/io/hitran.py
Supriya1702: Yes, actually I can also extract the same data from` fname `as well, so I have removed the `data_file` in the recent commit

radis/io/hitran.py
Supriya1702: It actually splits the path by delimiter `.`  so we get an output = 
('C:\\Users\\Supriya\\.radisdb\\hitran\\downloads__can_be_deleted\\CO\\CO_5',
 '.data')`, we require `CO_5` so again we will need to access the 0th index and split it using `\\` and again access is -1 index, which is more work so I believe sticking to the current implementation will be better

radis/io/dbmanager.py
```diff
@@ -514,6 +516,27 @@ def get_nrows(self, local_file):
             raise ValueError(engine)
         return nrows
 
+    def get_columns(self, local_file):
```
erwanp, line 519: nice function. 

It would be nice to test it. There is already a test set-up that uses all engines, can you include the test for Pytables (pandas) and Vaex there : 

https://github.com/Supriya1702/radis/blob/fetch_hitran_update/radis/test/io/test_hdf5.py#L45

just assert() that the columns are what they are

radis/io/hitran.py
Supriya1702: Got it

radis/io/hdf5.py
```diff
@@ -213,6 +213,14 @@ def combine_temp_batch_files(self, file, key="default", sort_values=None):
             import vaex
 
             df = vaex.open(self._temp_batch_files, group=key)
+            # Removing Nan values columns
```
erwanp, line 216: minor suggestion, but maybe add a `delete_nan_columns`  parameter  (default True)  in combine_temp_batch_files() and run these lines only if is is True

radis/io/hitran.py
Supriya1702: Okay, but line 1607  containing ` ``` ` should also not be there right?

radis/io/hitran.py
erwanp: 👍

Supriya1702:
@erwanp I was thinking of adding test cases to check if we are able to download the new parameters or not. But am not sure how exactly to implement them, because all the current tests that are being run must be downloading the original database, we can add a test by regenerating the database with extra_params  = 'all' or how would you like to proceed with it?

erwanp (changes requested):

erwanp:
> @erwanp I was thinking of adding test cases to check if we are able to download the new parameters or not. But am not sure how exactly to implement them, because all the current tests that are being run must be downloading the original database, we can add a test by regenerating the database with extra_params = 'all' or how would you like to proceed with it?

Talking about test cases, i suggested  https://github.com/radis/radis/pull/505#discussion_r947061287  

Good idea to add the new test cases. We'll need them anyway. We shouldn't load a large database, because it will be downloaded on each CI run.  But you can try with CO. Does CO have broadneing parameters by other species ? (CO2, etc. ? ) 

I'd add a `fetch_hitran()` test in [test_hitran_cdsd](https://github.com/Supriya1702/radis/blob/fetch_hitran_update/radis/test/io/test_hitran_cdsd.py) similarly to how we already test `fetch_hitemp()`  (using "regen")   in [test_hitemp.py](https://github.com/Supriya1702/radis/blob/fetch_hitran_update/radis/test/io/test_hitemp.py#L80)

radis/io/hitran.py
```diff
@@ -328,6 +340,237 @@ def hit2df(
...
+    nmol = len(df["id"].unique())
+    if nmol == 0:
+        raise ValueError("Databank looks empty")
+    elif nmol != 1:
+        # Crash, give explicity error messages
+        try:
+            secondline = df.iloc[1]
+        except IndexError:
+            secondline = ""
+        raise ValueError(
```
erwanp, line 488: Hello @Supriya1702 , I was wrong in asking to create a hit2df_hapi function.  The function does not convert a file to a DataFrame; since actually the file -> memory has been done by [fetch()  above](https://github.com/radis/radis/pull/505/files#diff-f1a2be23a785af89993778ad1bad43a66e9f8b0a2825be0060fef6434dc6b390R1411).  All the caching / metadata checks of hit2df are not required here. 

Therefore, hit2df_hapi should be removed., and your original implementation was better. 

Instead, the only thing you want to keep is the post-processing; . ie lines 475-532 (which includes the parsing). 
You can extract these lines in a new `post_process_hitran_data()` function, shared by `hit2df()`  in hitran.py; and by `download_all_hitran_isotopes` below

erwanp (changes requested):

radis/io/hitran.py
anandxkumar: Yes, remove them as well

radis/test/io/test_hdf5.py
```diff
@@ -64,6 +65,19 @@ def test_hdf5_io_engines(*args, **kwargs):
     df.close()
     assert manager.read_metadata("test_h5py.h5", key=None) == metadata0
 
+    # Test get_columns function of DatabaseManager
+    # For vaex
+    dbmanager = DatabaseManager(
+        name="test", local_databases="hitran", molecule="CO", engine="vaex"
+    )
+    assert list(dbmanager.get_columns("test_h5py.h5")) == ["a", "b"]
```
anandxkumar, line 73: Nice :+1:

radis/io/hitran.py
Supriya1702: @erwanp  I've done the changes, can you review them?

anandxkumar:
Great work @Supriya1702, tested it locally, works fine on my end. The only thing missing will be adding a test to check if the `extra_params` are loading the extra columns or not. I think `CO` will be a good example of it as it contains `gamma_*` and `n_*` for `co2, h2o, he `(can be used in future for calculating non-air broadening coeff). Maybe we can set `extra_params=True` and `db_use_cached='regen'` to check the same in `fetch_databank` and then test whether columns are loading for load_columns = `equilibrium`, `non-eq` and `all`?

radis/test/io/test_hdf5.py
erwanp: Actually @Supriya1702, can't you move get_columns() in DataFileManager?

radis/io/dbmanager.py
erwanp, line 519: As mentionned in another comment ; it would be more organized if this method was moved in [DataFileManager](https://github.com/radis/radis/blob/634a77c0da26b8d8b97741d38cbdffed5de4d649/radis/io/hdf5.py#L72)  (which exclusively deals with Files; not necessarily Database objects... that's the case for your function ) 


And keep a wrapper here : 

```python
def get_columns(self, local_file):
    return self.get_datafile_manager().get_columns(local_file)
```

radis/io/dbmanager.py
```diff
@@ -514,6 +516,27 @@ def get_nrows(self, local_file):
...
+        """Get all columns (without loading all Dataframe"""
+        engine = self.engine
+        local_file = expanduser(local_file)
+        if engine == "vaex":
+            import vaex
+
+            # by default vaex does not load everything
+            df = vaex.open(local_file)
+            columns = df.columns
+            df.close()
```
erwanp, line 529: Note : you can use a with() to avoid the close() : 
```suggestion
            with vaex.open(local_file) as df:
                columns = df.columns
```

radis/io/hitran.py
```diff
@@ -215,6 +227,104 @@ def hit2df(
...
+    # by parsing df.wav.   Completely irrelevant files are discarded in 'load_h5_cache_file'
+    # but files that have partly relevant lines are fully loaded.
+    # Note : cache file is generated with the full line list.
+
+    return df
+
+
+def post_process_hitran_data(
+    df,
+    fname,
```
erwanp, line 281: we don't need to read the filename; you only use it to guess what's the molecule. Give it the molecule directly as an input parameter

erwanp (changes requested):

radis/io/hitran.py
Supriya1702: Okay, so then we would also have to pass molecule in the `hit2df` function. I'll have to add `molecule` as a parameter in the definition of `hit2df` as well then only `post_process_hitran_data` can be easily shared by `hit2df` and  `download_all_hitran_isotopes`

erwanp: I don't think so ; hit2df reads the file so it can infer it itself

radis/io/dbmanager.py
Supriya1702: Hi @erwan seems like with statement has `open` function associated with it, thus it overrides it when we are calling it,  thus performing different function which gives rise to error:
`Attribute Error:  __enter__ is not defined`
and also using different read-write operations it gives following error:
`ValueError: Unrecognized file extension. Please use .arrow, .hdf5, .parquet, .fits, or .csv to export to the particular file format.`

So I guess sticking with current implementation will be good.

radis/io/hitran.py
Supriya1702: Okay

erwanp (approved):
LGTM

radis/io/dbmanager.py
erwanp: ok, no problem

radis/io/dbmanager.py
Supriya1702: I've done the changes, can you review them?

erwanp: perfect for me, merging as soon as tests pass

radis/test/io/test_hdf5.py
erwanp: actually @Supriya1702 last change; since the get_columns() is in DataFileManager can you call it directly from `manager` in the example above ? 

Also, make sure that you also write a test for the engine="pytables" version

erwanp (changes requested):

radis/test/io/test_hdf5.py
Supriya1702: I've done the changes

erwanp (approved):

erwanp:
Well done ; I let you merge when the pipeline succeeds!

PR #519, opened 2022-09-03
Implement all new fitting modules into RADIS

TranHuuNhatHuy:
This pull request is to implement all of my modules, developed separately in my repo [RADIS-Spectrum-Fitting-Benchmark](https://github.com/TranHuuNhatHuy/RADIS-Spectrum-Fitting-Benchmark), into RADIS codebase. This pull request features:

- Add `new_fitting.py` and `new_fitting_models.py` into `radis/tools/`. 
- Add a gallery example file, current not work due to an import issue, will describe below.
- Add an experimental spectrum into `radis/test/files`, named `synth-NH3-1-500-2000cm-P10-mf0.01-p1.spec`.

Mr. @erwanp , currently I am facing an import issue while trying to run my gallery example file. As you can see, my example file is `radis/examples/new_fitting_Tgas.py`. In line 41, I attempted to import my newly added module with:
```python
from radis.tools.new_fitting import fit_spectrum
```
However, it does not recoginze this module: 
```
ModuleNotFoundError: No module named 'radis.tools.new_fitting'
```
Although I have modified the `__init__.py` in `radis/tools/` to include my modules and their methods.

Could you kindly help me with this? Other examples will be added after this issue has been solved.


erwanp:
new_fitting and new_fitting_models both depend on each other. Make sure that one of them doesn't require the other one. Else you're creating circular imports

TranHuuNhatHuy:
> new_fitting and new_fitting_models both depend on each other. Make sure that one of them doesn't require the other one. Else you're creating circular imports

I tried to merge `new_fitting` and `new_fitting_models` so they are in the same file, and thus no need for imports. However, the `radis.tools.new_fitting` still cannot be recognized. I'm not sure how can I natively import `new_fitting.py`.

erwanp:
Then you might have different Radis installed. Check paths and files. Ex `print(radis. __file__) `

TranHuuNhatHuy:
Thank you Mr. @erwanp . Actually I've just found the solution. Seems like after I added new methods and registered them in `__init__.py` files, I need to reinstall my local RADIS environment again to update these changes. It worked! Now they can recognize my modules.

Additional gallery examples will follow up in later commits.

erwanp (approved):

examples/new_fitting_Tgas.py
```diff
@@ -0,0 +1,110 @@
+# -*- coding: utf-8 -*-
```
erwanp, line 1: File should be named "plot_something" for ReadTheDocs to generate plots

So "plot_fitting_Tgas"

TranHuuNhatHuy: Wow, new thing to learn. Is this some of Readthedoc 's rules, or our native setup? (if yes could you give me a reference? I want to know more about this automatic wiki setup later on)

erwanp: it's just a ReadTheDocs rule, don't waste time on investigating this ;)

dcmvdbekerom:
I noticed one of the Cython files wasn't included in the manifest, pushed it to this PR as it looks like this will be merged soon

erwanp (changes requested):
`radis/test/files/synth-NH3-1-500-2000cm-P10-mf0.01-p1.spec` is 20 MB, it should not be merged into the main repo. Actually, it shouldn't appear in the history neither, or it will make all future downloads slow. 

Can you  
- find another test file, crop it, or find another way to reduce its size. Hundred kbs should be the max. 
- either reopen another PR (cherry picking all changes including Dirk's, but the file); or clean this one & force push  ?

TranHuuNhatHuy:
Thank you Mr. @dcmvdbekerom for your addition <3

Mr. @erwanp , after trying to clean this PR but not successful, I decided to close this PR and re-open another one, while cherry-picking all changes.

Issue #523, opened 2022-09-09
🚀 Feature: Allow `crop()` to deal with astropy units


TranHuuNhatHuy:
### 🔖 Feature description

Currently the [crop()](https://radis.readthedocs.io/en/latest/source/radis.spectrum.operations.html#radis.spectrum.operations.crop) function only allows 3 units: `nm`, `cm-1` and `nm_vac`. It's better if other astropy units, such as `u.m`, can be accepted too.

### 👉 Why you want this feature!!

If applicable, this will provide better flexibility to the RADIS experiment. Also, this will able [new fitting module](https://radis.readthedocs.io/en/latest/source/radis.tools.new_fitting.html) to perceive astropy units at `wmin` and `wmax`.

TranHuuNhatHuy:
Close as completed in #526 .

PR #526, opened 2022-09-15
Allow astropy units to be used in crop()

TranHuuNhatHuy:
This PR is to address the issue #523 . 

Now, we can use astropy units in `crop()` by 2 ways:

1. Directly using them in `wmin` and `wmax`:
````python
s = s_exp.crop(
    4.167 * u.um,
    4.180 * u.um,
)
````
2. State the unit in `wunit`:
````python
s= s_exp.crop(
    0.004167,
    0.004180,
    wunit = "mm",
)
````

Below is the script I used to check if the conversion works fine:
````python
from radis import load_spec, plot_diff
import numpy as np

s_exp = (
    load_spec(getTestFile("CO2_measured_spectrum_4-5um.spec"))
    .sort()
)

# The original way, as shown in the old 1-temperature fit example
s_exp_1 = s_exp.crop(
    4167,
    4180,
    wunit = "nm",
)

# The 1st new proposed way of using astropy units directly in wmin and wmax
s_exp_2 = s_exp.crop(
    4.167 * u.um,
    4.180 * u.um,
)

# The 2nd new proposed way of declaring astropy units in wunit
s_exp_3 = s_exp.crop(
    0.004167,
    0.004180,
    wunit = "mm",
)

# Now check for equality

A_1 = s_exp_1.get(
    s_exp_1.get_vars()[0],
    wunit = "nm",
)
A_2 = s_exp_2.get(
    s_exp_2.get_vars()[0],
    wunit = "nm",
)
A_3 = s_exp_3.get(
    s_exp_3.get_vars()[0],
    wunit = "nm",
)

# Compare the absorbance, should be the same for all
assert np.array_equal(A_1, A_2)
assert np.array_equal(A_1, A_3)
````

Mr. @erwanp , when you approve the code, I will add this script into the function's test.


anandxkumar:
Hey @TranHuuNhatHuy, nice addition this would be. Good work. You may want to investigate why `radis/test/db/test_molecule_lists.py ` is failing, I don't see anything that should be causing the assertion error, can you verify it locally?
Also, add a test case for the same, would be good to keep the physics in check.

erwanp:
@TranHuuNhatHuy can you use inplace=False in your crop() tests. If by mistake they were true youd be comparing the same spectra 3 times.

Also, I don't think the test file is absorbance. Should be radiance. It doesn't make your test invalid, but to prevent any confusion, name it radiance.
And actually, you do not need to compare intensities. You should be comparing the waver ange. For instance use get_wavelength()

erwanp:
PS : test_molecule_list may fail if the Exomol team released a new molecule. The error message should be explicit. If that's the case you just have to update the hard-coded list with the new molecule

TranHuuNhatHuy:
Thank you @anandxkumar for your suggestion. I tried to run the test file locally, and got this error log:
````
ExoMol molecules, fetched online 
['AlCl', 'AlF', 'AlH', 'AlO', 'AsH3', 'BeH', 'C2', 'C2H2', 'C2H4', 'CH', 'CH3', 'CH3Cl', 'CH3F', 'CH4', 'CN', 'CO', 'CO2', 'CP', 'CS', 'CaF', 'CaH', 'CaO', 'CaOH', 'CrH', 'FeH', 'H2', 'H2CO', 'H2O', 'H2O2', 'H2S', 'H2_p', 'H3O_p', 'H3_p', 'HCN', 'HCl', 'HF', 'HNO3', 'HeH_p', 'KCl', 'KF', 'KOH', 'LaO', 'LiCl', 'LiF', 'LiH', 'LiH_p', 'MgF', 'MgH', 'MgO', 'N2', 'N2O', 'NH', 'NH3', 'NO', 'NS', 'NaCl', 'NaF', 'NaH', 'NaO', 'NaOH', 'NiH', 'O2', 'OCS', 'OH', 'OH_p', 'PF3', 'PH', 'PH3', 'PN', 'PO', 'PS', 'SH', 'SO2', 'SO3', 'ScH', 'SiH', 'SiH2', 'SiH4', 'SiN', 'SiO', 'SiO2', 'SiS', 'TiH', 'TiO', 'VO', 'YO', 'cis-P2H2', 'trans-P2H2']
Comparing Radis hardcoded ExoMol molecules list to the ExoMol website: 
        Radis molecules                         Fetched from ExoMol website
        ----------------------------------------------------
        N/A                     |# 27|  LaO (<class 'str'>)
        ----------------------------------------------------

  File "/Users/tranhuunhathuy/Documents/GitHub/radis/radis/test/db/test_molecule_lists.py", line 363, in <module>
    test_ExoMol_molecules_list()
  File "/Users/tranhuunhathuy/Documents/GitHub/radis/radis/test/db/test_molecule_lists.py", line 53, in test_ExoMol_molecules_list
    assert (
AssertionError
````
I don't think this has anything to do with my newly implemented `crop()`. Probably someone messed up the ExoMol, as suggested by Mr. @erwanp ?

By the way, Mr. @erwanp , regarding the `inplace`, actually in `crop()` we have `inplace = False` as default. So I think we have been making copies of the original spectrum, so it's safe!
And yes, the spectrum object has radiance as its first spectral quantity, not absorbance. My apologies for the wrong annotation (don't really remember why I wrote absorbance...)

So, latest commit features additional 2 test cases for `radis/test/soectrum/test_operations/test_crop`, including:
- Fix annotation, from `absorbance` to `radiance` (oh wait but actually I didn't mention it in the test cases, so nvm).
- Add those 2 test cases with 2 new methods for crop input.
The `test_crop()` ran well, so I suppose no problems with the unit conversion till now.

erwanp:
Do not assume default options will remain the same.

Also, fix the Exomol file by adding the missing molecule

TranHuuNhatHuy:
I found out that recently Exomol added `LaO` as a new molecule, and thus in latest commit I added it to the hardcoded list of `EXOMOL_MOLECULES`. Now waiting for Travis checks.

erwanp:
Lgtm. You can merge!

erwanp (approved):

PR #527, opened 2022-09-19
Removing completed TODO in radis/lbl/calc.py

mdrodelo:
### Description

This pull request is to address a TODO found in radis/lbl/calc.py. The change catches and raises a KeyError when a molecule is missing from a dictionary (isotope, mole_fraction, databank, or overpopulation) for calc_spectrum. The raised error message informs the user of which molecule and dictionary caused the KeyError and also tries to provide some additional context.

Here are the comments of the TODO for reference (found on lines 465 and 466 main branch's calc.py):
\# Will raise a KeyError if not defined. That's fine!
\# TODO: maybe need to catch KeyError and raise a better error message?

This is my first OSS contribution. Any help or input is appreciated.

### DIFF
--- a/radis/lbl/calc.py
+++ b/radis/lbl/calc.py
@@ -470,7 +470,7 @@ def calc_spectrum(
                             molecule,
                             argument_name,
                             molecule_reference_set,
-                            argument_dict.keys())
+                            set(argument_dict.keys()))
                     )
             else:  # argument_name is not a dictionary.
                 # Let's distribute the same value to every molecule:
--- a/radis/lbl/calc.py
+++ b/radis/lbl/calc.py
@@ -461,17 +461,7 @@ def calc_spectrum(
         for argument_name, argument_dict in DICT_INPUT_ARGUMENTS.items():
             if isinstance(argument_dict, dict):
                 # Choose the correspond value
-                try:
-                    molecule_dict[molecule][argument_name] = argument_dict[molecule]
-                except KeyError:
-                    # KeyError is raised if a molecule is missing from the dictionary
-                    raise KeyError(
-                        "'{0}'' missing in '{1}=' argument. For reference, here are the molecules provided in the following arguments: 'molecule=' {2} and '{1}=' {3}".format(
-                            molecule,
-                            argument_name,
-                            molecule_reference_set,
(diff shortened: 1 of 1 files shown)


radis/lbl/calc.py
```diff
@@ -461,9 +461,17 @@ def _check_molecules_are_consistent(
...
+                try:
+                    molecule_dict[molecule][argument_name] = argument_dict[molecule]
+                except KeyError:
+                    # KeyError is raised if a molecule is missing from the dictionary
+                    raise KeyError(
+                        "'{0}'' missing in '{1}=' argument. For reference, here are the molecules provided in the following arguments: 'molecule=' {2} and '{1}=' {3}".format(
+                            molecule,
+                            argument_name,
+                            molecule_reference_set,
+                            set(argument_dict.keys()))
```
erwanp, line 473: Thanks for contributing. 

The `f""` syntax is more readable than `"".format()`; can you use it ? I.e `f"'{molecule}' missing in '{argument_name}=' argument (...) `

Can you also provide an example of how it would look like in real-error ?

erwanp (changes requested):

radis/lbl/calc.py
mdrodelo: I was working to set up an example error and I could use some clarification on the TODO. I think this TODO may already taken care of earlier in the calc_spectrum function. 

I understood the comment as saying a KeyError occurs in the following line: `molecule_dict[molecule][argument_name] = argument_dict[molecule]`. Specifically because of the `argument_dict[molecule]` part. For example, if the isotope= argument was provided a dict missing a molecule that was provided in molecule=.

Any inconsistencies seem to be pretty well covered by the `_check_molecules_are_consistent `function defined earlier in calc_spectrum, starting on line 402 at the moment. I don't see a line where any of the dictionaries could be modified between the `_check_molecules_are_consistent `function and the line were the TODO was listed.

I think the TODO can be removed and my pull request dropped, but let me know if I missed something.

erwanp: Ok, sounds good ! Then you can adjust your PR to at least remove the TODO , and we'll merge it

erwanp (approved):

PR #537, opened 2023-01-15
 added double axis and updated exomol molecules list and get_molecule method of DataFileManager class

1someshverma:
This is to solve issuse https://github.com/radis/radis/issues/446 on Radis repository
Also, I have updated Exomol Molecules list and added ZrO, Na and K in it the list.
Plot with double axis (nm, cm-1) https://github.com/radis/radis/issues/446

It provides the secondary or double x axis at the top of plot
-When the plot is between wavenumber(cm-1 )vs radiance , it adds a secondary axis for wavelength(nm)
-When the plot is between radiance vs wavelenght , it also adds a secondary axis for wavenumber(cm-1)

Fixes https://github.com/radis/radis/issues/446

 
Screenshot from 2023-01-16 01-09-20

Edit : Updated code to use functions defined in .pysics.convert and now wavelength is obtained depending upon whether medium is air or vaccum

PR #538, opened 2023-01-16
prevent imax index from exceeding the range

ykawashima:
### Description

I am Yui Kawashima, and involved in the development of exojax. I have just realized that when the highest value of the wavenumber grid exceeds the range of the molecular line list, imax is set outside the range [here](https://github.com/radis/radis/blob/0ea4672952a459393ef3384effe5abdc1de862fa/radis/api/exomolapi.py#L1266). So I would like to send a pull request to avoid this. An example script you can check is as follows. Thank you in advance.
```
from exojax.spec import api
from exojax.spec import defmol

db_dir = defmol.search_molfile("ExoMol", "H2CO")
nus, wav, res = wavenumber_grid(990,1010,10000,unit="nm")
mdb = api.MdbExomol(db_dir, nus, gpu_transfer=True)
```

### DIFF
--- a/radis/api/exomolapi.py
+++ b/radis/api/exomolapi.py
@@ -1267,6 +1267,8 @@ class MdbExomol(DatabaseManager):
                 np.searchsorted(dic_def["numinf"], nurange[1] + margin, side="right")
                 - 1
             )
+            # not to exceed index out of the range
+            imax = np.min([imax, len(dic_def["numinf"])-2])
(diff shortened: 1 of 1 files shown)


erwanp (approved):
All good let's merge!

erwanp:
Test failures are unrelated and will be fixed by #537

ykawashima:
Thank you!

PR #541, opened 2023-01-24
Features/exomol loading optional fields, such as quantum states, when kip_optional_data=False in api.read_states

### LINKED ISSUE #540: 🚀 Feature: Loading fields other than mandatory fields(i, E, g, J), Lande, lifetime in ExoMol.
### 🔖 Feature description

Some ExoMol .states provides fields other than mandatory fields, such as vibration states (v), electronic states(eS, X2Sigma+ etc). Related to [Issue 330](https://github.com/HajimeKawahara/exojax/issues/330) in ExoJAX, we would make `exomolapi` to load these additional fields in ExoMol. 

Currently, I am implementing this feature in `features/exomol_qstates` branch.

In my case, I'd like to compute cross sections for each vib state separately. Also, I would know eS of each line in some case.

HajimeKawahara:
### Description

This pull request is to address the loading optional fileds in ExoMol when `skip_optional_data=False` in `api.read_states`.

The loading a .states file based on the def file is a bit tricky, but the following rules work well for most species:

- if `unc `(uncertainty of line center),`lifetime`, `Lande` exists, additional column(s) should be placed before `quantum_labels` in this order
- some additional columns (maybe not described in def) can exist after  `quantum_labels` (I found e.f. Marvelized/Calculated, Calculated energy)

Note that sometime the def file looks inconsistent with states in terms of the description of  `quantum_labels`. It should be handled as an exception in `read_def`. For instance, HITRAN-HCl, see [here](https://github.com/HajimeKawahara/exojax/issues/330#issuecomment-1401174128) 

Fixes #540
See also https://github.com/HajimeKawahara/exojax/issues/330

### DIFF
--- a/radis/api/exomolapi.py
+++ b/radis/api/exomolapi.py
@@ -128,7 +128,7 @@ def read_def(deff):
             quantum_labels.append(dat["VAL"][i].strip(" "))
         elif "Uncertainty availability" in com:
             unc = int(dat["VAL"][i]) == 1
-
+    
     # SOME DEF FILES CONTAINS ERRORS. THESE ARE THE EXCEPTIONS
     if deff.stem == "12C-16O2__UCL-4000":
         ntransf = 20
@@ -137,6 +137,9 @@ def read_def(deff):
         ntransf = 100
     if deff.stem == "14N-1H3__CoYuTe":
         maxnu = 20000.0
+    if deff.stem == "40Ca-16O-1H__OYT6":
+        ntransf = 18
+        maxnu = 36000.0
     if deff.stem == "12C2-1H2__aCeTY":
         if molmass == 12.0:
             molmass = 26.0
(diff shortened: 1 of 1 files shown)


erwanp (approved):
All good minus minor code linting error

To fix linting error you can run black on all files or install pre-commit  https://radis.readthedocs.io/en/latest/dev/developer.html#code-linting

erwanp:
Linted with
```
 pre-commit run --all-files
```
You can merge !

HajimeKawahara:
Thanks! I will install **black** in VSCode.

erwanp:
I think you just need to setup 
```
pre-commit install
```
in the Radis folder. It will install the proper versinos of black + Flake8  + all these linting libraries that are annoying at first and start being useful when more people contribute ;)   ,

HajimeKawahara:
Got it, it's cool!

PR #546, opened 2023-02-06
Warn when branch data type is not int64

### LINKED ISSUE #544: Bug Report: fails at calculating N2 absorption spectrum
### Description

Calculating the absorption spectrum (at equilibrium) of the N2 molecule fails, raising a `AssertionError.`

### Reproducible example

```python
import radis

radis.calc_spectrum(
    molecule="N2",
    wmin=7000,  # arbitrary
    wmax=8000,  # arbitrary
    pressure=1.0,
    Tgas=300,
    databank="hitran",
    cutoff=0.0,
)
```

### 💡 Possible solutions

The exception is raised in `replace_PQR_with_m101` method from the `radis/api/tools` module, by the [following line](https://github.com/radis/radis/blob/3ba4cc564ecccd6a0718aed8a55d17a63911ad51/radis/api/tools.py#L254).
I assume the `branch` column has values other than `"P"` `"Q"` or `"R"` which are not replaced by integers, but I did not check that.
Possibly related to #65.

### 🎲 Radis version

0.14

### 💻 Operating system

Linux

nollety:
### Description

This pull request is to remove an assertion error being raised by `radis/api/tools.py:replace_PQR_with_m101
` when the data type for the `'branch'` column is other than `int64`, and instead issue a warning.

Fixes #544

### DIFF
--- a/radis/api/tools.py
+++ b/radis/api/tools.py
@@ -162,7 +162,7 @@ def _format_dtype(dtype):
         try:
             print("Data type")
             print("-" * 30)
-            for (k, c) in dtype:
+            for k, c in dtype:
                 print(str(k), "\t\t", c)
             print("-" * 30)
         finally:
@@ -269,4 +269,3 @@ def replace_PQR_with_m101(df):
             ),
             category=PerformanceWarning,
         )
-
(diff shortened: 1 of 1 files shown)


erwanp (approved):
Thank you, nice fix ! Merging after tests pass

Issue #525, opened 2022-09-11
Feature request : Allow multiple diluents calculation for multiple molecules in calc_spectrum


Supriya1702:
### 🐛 Describe the bug

Currently, we can computer spectrum for a single molecule in non-air diluents (implemented in [PR: 495](https://github.com/radis/radis/pull/495). For example: 
```
calc_spectrum(
        wavenum_min=2245,
        wavenum_max=2250,
        Tgas=700,
        path_length=0.1,
        molecule="CO",
        mole_fraction=0.2,
        isotope=1,
        databank="hitran",
        diluent={"CO2": 0.2, "H2O": 0.1, "air": 0.5}
)
```

If we want to compute spectrum for multiple molecules for multiple diluents, that is not currently implemented and will raise a `NotImplementError`:

```
calc_spectrum(
        wavenum_min=2245,
        wavenum_max=2250,
        Tgas=700,
        path_length=0.1,
        mole_fraction={"CO":0.4, "CO2":0.4},
        isotope=1,
        databank="hitran",
        export_lines=True,
)
```

### 💡 Possible solutions

Modify the [calc_spectrum](https://github.com/radis/radis/blob/develop/radis/lbl/calc.py) class, and check if molefraction is a dictionary, if yes then for every molecule in molefraction, use the diluent to create spectrum using SpectrumFactory(which is already done), and merge all the spectrums.

Note: Make sure to evaluate all the molefractions of diluents and molecules properly.

### 🎲 Radis version

0.13.1

### 💻 Operating system

Linux

erwanp:
Thanks! Tagging this as a Good First Issue for new contributors

4the4ryushin:
is this issue still open?
i would like to work on it

4the4ryushin:
i am new to radis but not  but i have contributed to other projects before

4the4ryushin:
![image](https://user-images.githubusercontent.com/74553951/194912264-575ed6ce-963e-43da-beac-4c29fe27f4a2.png)
its working?

erwanp:
Hello, welcome. Please hve a look at our [Contributing Guide](https://github.com/radis/radis/blob/develop/CONTRIBUTING.md)
 Open a PR when you feel it's ready !

4the4ryushin:
is it a issue or it's just my environment , raids dev installation is throwing an error on latest python3 environment ,3.10+
while it takes a lot of time to install in 3.8+

[...]

PR #549, opened 2023-03-07
Update molecular parameters

nollety:
### Description

This pull request updates the molecular parameters stored in [radis/db/molparams.txt](https://github.com/radis/radis/blob/f46147575fdaeac5e0a5a44e958e88d82467a6e3/radis/db/molparam.txt) with the latest version of https://hitran.org/media/molparam.txt.

PR #555, opened 2023-04-02
Multiple diluents calculation for multiple molecules in calc_spectrum

1someshverma:
This PR is to solve #525 

To calculate the Spectrum of multiple diluents and multiple molecules, I have used the following approach.
-First, the spectrum of individual molecules is calculated, considering other molecules as diluents, in case the broadening coefficient of some molecules is not present then the broadening coefficient of air is used and a warning message is raised.
-Then Spectrum of individual molecules is merged to calculate the final Spectrum.

If there is anything else I need to do let me know.

Screenshot from 2023-04-02 13-01-08

### DIFF
--- a/radis/lbl/broadening.py
+++ b/radis/lbl/broadening.py
@@ -899,7 +899,6 @@ class BroadenFactory(BaseFactory):
                 ]
                 diluent_broadening_coeff["n_" + key.lower()] = df["n_" + key.lower()]
             except KeyError:
-                    self.warn()
                     if key!= "air":
                         self.warn(
                                 message="Broadening Coefficient of "+key+" not present in database using broadening coefficient of air instead.",
--- a/radis/lbl/broadening.py
+++ b/radis/lbl/broadening.py
@@ -340,7 +340,7 @@ def pressure_broadening_HWHM(
                         * pressure_atm
                         * diluent_mole_fraction
                     )
-            except:
+            except KeyError:
                 import warnings
                 from radis.misc.warning import AccuracyWarning
                 if diluent_molecule != "air":
@@ -898,14 +898,10 @@ class BroadenFactory(BaseFactory):
                     "gamma_" + key.lower()
                 ]
                 diluent_broadening_coeff["n_" + key.lower()] = df["n_" + key.lower()]
-            except:
-                    import warnings
-                    from radis.misc.warning import AccuracyWarning
+            except KeyError:
                     if key!= "air":
-                        warnings.warn(
-                            AccuracyWarning(
-                                "Broadening Coefficient of "+key+" not present in database using broadening coeffiecient of air instead."
-                            )
+                        self.warn(
+                                "Broadening Coefficient of "+key+" not present in database using broadening coefficient of air instead."
                         )
 
         # Get broadenings
--- a/radis/lbl/broadening.py
+++ b/radis/lbl/broadening.py
@@ -346,9 +346,10 @@ def pressure_broadening_HWHM(
                 if diluent_molecule != "air":
                     warnings.warn(
                         AccuracyWarning(
-                            "Broadening Coefficient of "+diluent_molecule+" not present in database using broadening coefficient of air instead."
+                            "Broadening Coefficient of "+diluent_molecule+" not present in database using broadening coefficient of air instead. \n!!! Solution: Try using once `use_cached='regen'' in calc_spectrum!!!"
                         )
                     )
+                    
                 gamma_lb += ((Tref / Tgas) ** Tdpair) * (
                     (airbrd * pressure_atm * diluent["air"])
                 )
@@ -901,7 +902,7 @@ class BroadenFactory(BaseFactory):
             except KeyError:
                     if key!= "air":
                         self.warn(
-                                message="Broadening Coefficient of "+key+" not present in database using broadening coefficient of air instead.",
+                                message="Broadening Coefficient of "+key+" not present in database using broadening coefficient of air instead. \n!!! Solution: Try using once `use_cached='regen'' in calc_spectrum!!!",
                                 category="AccuracyWarning"
                         )
 
--- a/radis/lbl/broadening.py
+++ b/radis/lbl/broadening.py
@@ -341,15 +341,16 @@ def pressure_broadening_HWHM(
                         * diluent_mole_fraction
                     )
             except KeyError:
-                import warnings
-                from radis.misc.warning import AccuracyWarning
-                if diluent_molecule != "air":
-                    warnings.warn(
-                        AccuracyWarning(
-                            "Broadening Coefficient of "+diluent_molecule+" not present in database using broadening coefficient of air instead. \n!!! Solution: Try using once `use_cached='regen'' in calc_spectrum!!!"
-                        )
-                    )
-                    
+                # import warnings
+                # from radis.misc.warning import AccuracyWarning
+                # if diluent_molecule != "air":
+                #     warnings.warn(
+                #         AccuracyWarning(
+                #             "Broadening Coefficient of "+diluent_molecule+" not present in database using broadening coefficient of air instead. \n!!! Solution: Try using once `use_cached='regen'' in calc_spectrum!!!"
+                #         )
+                #     )
+                
+                ## A warning is normally raised already in '_calc_broadening_HWHM' when constructing the dict diluent_broadening_coef
                 gamma_lb += ((Tref / Tgas) ** Tdpair) * (
                     (airbrd * pressure_atm * diluent["air"])
                 )
@@ -900,11 +901,11 @@ class BroadenFactory(BaseFactory):
                 ]
                 diluent_broadening_coeff["n_" + key.lower()] = df["n_" + key.lower()]
             except KeyError:
-                    if key!= "air":
-                        self.warn(
-                                message="Broadening Coefficient of "+key+" not present in database using broadening coefficient of air instead. \n!!! Solution: Try using once `use_cached='regen'' in calc_spectrum!!!",
-                                category="AccuracyWarning"
-                        )
+                if key!= "air":
+                    self.warn(
+                            message="Broadening Coefficient of "+key+" not present in database using broadening coefficient of air instead. \n!!! Solution: Try using once `use_cached='regen'' in calc_spectrum!!!",
+                            category="AccuracyWarning"
+                    )
 
         # Get broadenings
         if broadening_method == "voigt":
--- a/radis/lbl/broadening.py
+++ b/radis/lbl/broadening.py
@@ -328,35 +328,40 @@ def pressure_broadening_HWHM(
 
     diluent_molecules = diluent.keys()
 
-    # check if gamma_diluent and n_diluent exists or not
-    try:
-        gamma_lb = 0
-        for diluent_molecule, diluent_mole_fraction in diluent.items():
-            if diluent_molecule != "air":
-                gamma_lb += (
-                    (Tref / Tgas)
-                    ** diluent_broadening_coeff["n_" + diluent_molecule.lower()]
-                ) * (
-                    diluent_broadening_coeff["gamma_" + diluent_molecule.lower()]
-                    * pressure_atm
-                    * diluent_mole_fraction
+    # check if gamma_diluent and n_diluent exists or not and in case they are not present use broadening coefficient of air instead
+    gamma_lb = 0
+    for diluent_molecule, diluent_mole_fraction in diluent.items():
+            try:
+                    gamma_lb += (
+                        (Tref / Tgas)
+                        ** diluent_broadening_coeff["n_" + diluent_molecule.lower()]
+                    ) * (
+                        diluent_broadening_coeff["gamma_" + diluent_molecule.lower()]
+                        * pressure_atm
+                        * diluent_mole_fraction
+                    )
+            except:
+                import warnings
+                from radis.misc.warning import AccuracyWarning
+                if diluent_molecule != "air":
+                    warnings.warn(
+                        AccuracyWarning(
+                            "Broadening Coefficient of "+diluent_molecule+" not present in database using broadening coefficient of air instead."
+                        )
+                    )
+                gamma_lb += ((Tref / Tgas) ** Tdpair) * (
+                    (airbrd * pressure_atm * diluent["air"])
                 )
-        # Adding air coefficient
-        if "air" in diluent_molecules:
-            gamma_lb += ((Tref / Tgas) ** Tdpair) * (
-                (airbrd * pressure_atm * diluent["air"])
-            )
-        # Adding self coefficient
-        if Tdpsel is None:  # use Tdpair instead
+    # Adding self coefficient
+    if Tdpsel is None:  # use Tdpair instead
             gamma_lb += ((Tref / Tgas) ** Tdpair) * (
                 (selbrd * pressure_atm * mole_fraction)
             )
-        else:
+    else:
             gamma_lb += ((Tref / Tgas) ** Tdpsel) * (
                 selbrd * pressure_atm * mole_fraction
             )
-    except KeyError as err:
-        raise KeyError("Column not found {0}".format(err))
+        # raise KeyError("Column not found {0}".format(err))
 
     return gamma_lb
 
@@ -888,11 +893,20 @@ class BroadenFactory(BaseFactory):
         # diluent and their broadening coeff dictionary
         diluent_broadening_coeff = {}
         for key in diluent:
-            if key != "air":
+            try:
                 diluent_broadening_coeff["gamma_" + key.lower()] = df[
                     "gamma_" + key.lower()
                 ]
                 diluent_broadening_coeff["n_" + key.lower()] = df["n_" + key.lower()]
+            except:
+                    import warnings
+                    from radis.misc.warning import AccuracyWarning
+                    if key!= "air":
+                        warnings.warn(
+                            AccuracyWarning(
+                                "Broadening Coefficient of "+key+" not present in database using broadening coeffiecient of air instead."
+                            )
+                        )
 
         # Get broadenings
         if broadening_method == "voigt":
--- a/radis/lbl/broadening.py
+++ b/radis/lbl/broadening.py
@@ -899,9 +899,11 @@ class BroadenFactory(BaseFactory):
                 ]
                 diluent_broadening_coeff["n_" + key.lower()] = df["n_" + key.lower()]
             except KeyError:
+                    self.warn()
                     if key!= "air":
                         self.warn(
-                                "Broadening Coefficient of "+key+" not present in database using broadening coefficient of air instead."
+                                message="Broadening Coefficient of "+key+" not present in database using broadening coefficient of air instead.",
+                                category="AccuracyWarning"
                         )
 
         # Get broadenings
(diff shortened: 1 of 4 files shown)


1someshverma:
@erwanp Sir, many tests are failing may be due to changes made by some of the previous pull requests. can you look into it?

1someshverma:
I think these errors are because of this pull request  [https://github.com/radis/radis/pull/549](url)  .

1someshverma:
I think this is happening because of PR #549

1someshverma:
Almost all of these are related to the pandas Series
So , it might be some issue with pandas 
Screenshot 2023-04-03 11:55:24

1someshverma:
It appears like this issue but earlier it was working fine [https://stackoverflow.com/questions/61327917/why-does-using-return-a-series-instead-of-bool-in-pandas](url)

radis/lbl/calc.py
```diff
@@ -523,6 +517,23 @@ def _check_molecules_are_consistent(
             float("inf"),
         ]  # Using a list to store minimum wstep value at 1st index
 
+    flag = False
```
erwanp, line 520: choose a more explicit name than "flag"

radis/lbl/calc.py
```diff
@@ -523,6 +517,23 @@ def _check_molecules_are_consistent(
...
+    # Check if mole_fraction is dictionary of multiple molecules and have non air diluent, if yes then set value of flag as true and update list of diluents
+    if isinstance(mole_fraction, dict) and isinstance(diluent, dict):
+        flag = True
+        for molecule, fraction in mole_fraction.items():
+            if molecule not in {"CO2", "H2O", "He", "H2"}:
+                import warnings
+
+                warnings.warn(
```
erwanp, line 528: Radis uses its own subclass of Warnings, which allows to activate/deactivate them. Here, use a [AccuracyWarning](https://radis.readthedocs.io/en/latest/source/radis.misc.warning.html#radis.misc.warning.AccuracyWarning) or subclass it into a "MissingBroadeningWarning"

radis/lbl/calc.py
```diff
@@ -531,6 +542,13 @@ def _check_molecules_are_consistent(
         # We add all of the DICT_INPUT_ARGUMENTS values:
         kwargs_molecule.update(**dict_arguments)
 
+        t = kwargs_molecule["mole_fraction"]
+        if flag:
+            if molecule not in {"CO2", "H2O", "He", "H2"}:
+                diluent["air"] -= t
```
erwanp, line 548: Depending on the database, these broadeners might or might not be available. For a given database, and depending on the molecule you're computing, broadeners might or might not be available too (i.e. it's possible that CO in HITRAN has CO2 broadening; but that N2 in HITRAN doesn't have CO2 broadneing).

Therefore:  no hardcoding is possible here. Better send all the correct diluent information; and let _calc_spectrum_one_molecule handle it (and use Air itself if not available)

erwanp (changes requested):

radis/lbl/calc.py
1someshverma: Yes, sir, I have changed it by checking in broadening functions whether the broadening coefficients are present are not, and in case they are not the present broadening coefficient of air is used.

1someshverma:
Once all tests are passed then I will commit it.

radis/lbl/calc.py
erwanp: I don't see the update, did you push?

1someshverma: Earlier there was some issue with tests so I didn't push then. I have pushed it now, let me know if it is correct.

radis/lbl/broadening.py
```diff
@@ -328,35 +328,40 @@ def pressure_broadening_HWHM(
...
+            try:
+                    gamma_lb += (
+                        (Tref / Tgas)
+                        ** diluent_broadening_coeff["n_" + diluent_molecule.lower()]
+                    ) * (
+                        diluent_broadening_coeff["gamma_" + diluent_molecule.lower()]
+                        * pressure_atm
+                        * diluent_mole_fraction
+                    )
+            except:
```
minouHub, line 343: please put a class of error to catch by "except", such as `except ValueError:`. This is to avoid that other errors are treated by this error handling

radis/lbl/broadening.py
```diff
@@ -888,11 +893,20 @@ def _calc_broadening_HWHM(self):
...
                     "gamma_" + key.lower()
                 ]
                 diluent_broadening_coeff["n_" + key.lower()] = df["n_" + key.lower()]
+            except:
+                    import warnings
+                    from radis.misc.warning import AccuracyWarning
+                    if key!= "air":
+                        warnings.warn(
+                            AccuracyWarning(
+                                "Broadening Coefficient of "+key+" not present in database using broadening coeffiecient of air instead."
```
erwanp, line 907: within SpectrumFactory I think you can use `self.warn` ; have a look at how it's done elsewhere in the code

radis/lbl/calc.py
```diff
@@ -523,6 +517,13 @@ def _check_molecules_are_consistent(
             float("inf"),
         ]  # Using a list to store minimum wstep value at 1st index
 
+    is_Multi = False
```
erwanp, line 520: make it more explicit

Is_multi what?  Multi_molecules ? Has_multiple_diluents ? 

Long words = more readable code = better maintenance

radis/lbl/calc.py
```diff
@@ -563,6 +568,10 @@ def _check_molecules_are_consistent(
             diluent=diluent,
             **kwargs_molecule,
         )
+
+        if is_Multi:
+                diluent[molecule] = fraction
```
erwanp, line 573: instead of modifying the main diluent dictionary; and re-correcting it afterwards within your loop ; why don't you just create a new add-hoc diluent dictionary specifically for this molecule ?  Seems more straightforward. And modifying input variables is always risky. 

i.E. 
``` 
diluent_for_this_molecule = {k:v for k,v in diluent.items()}  # copy 
diluent_for_this_molecule

**Review by erwanp — CHANGES_REQUESTED** [review]

File: radis/lbl/broadening.py (hunk 4 above)
Comment (1someshverma) [reply radis/lbl/broadening.py]: done

File: radis/lbl/broadening.py (hunk 5 above)
Comment (1someshverma) [reply radis/lbl/broadening.py]: done

**Review by erwanp — CHANGES_REQUESTED** [review]

File: radis/lbl/calc.py [hunk 8]
```diff
@@ -517,10 +517,10 @@ def _check_molecules_are_consistent(
             float("inf"),
         ]  # Using a list to store minimum wstep value at 1st index
 
-    is_Multi = False
+    is_multi_molecules_multi_diluents = False
```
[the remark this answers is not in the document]
1someshverma: Yes sir , it is used to popping the molecule from diluent_for_this_molecule in case multi molecules and multi diluents.

radis/lbl/broadening.py
```diff
@@ -898,14 +898,10 @@ def _calc_broadening_HWHM(self):
...
-            except:
-                    import warnings
-                    from radis.misc.warning import AccuracyWarning
+            except KeyError:
                     if key!= "air":
-                        warnings.warn(
-                            AccuracyWarning(
-                                "Broadening Coefficient of "+key+" not present in database using broadening coeffiecient of air instead."
-                            )
+                        self.warn(
```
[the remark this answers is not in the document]
1someshverma: done

radis/lbl/calc.py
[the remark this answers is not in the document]
erwanp: What about : 
```

# in the for molecule, dict_arguments in molecule_dict.items(): loop
diluent_for_this_molecule = diluent.copy() if isinstance(diluent, dict) else diluent    
if isinstance(mole_fraction, dict):
    for other_molecule, other_fraction  in mole_fraction.items():
        if other_molecule in diluent_for_this_molecule:
            diluent_for_this_molecule[other_molecule] += other_fraction
        else:
            diluent_for_this_molecule[other_molecule] = other_fraction
```
No need for tags, and no need to add items that you later pop. Seems more readable to me.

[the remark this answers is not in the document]
1someshverma: Ok sir I'll make the changes 🙂

radis/lbl/calc.py
```diff
@@ -532,13 +525,24 @@ def _check_molecules_are_consistent(
...
-            diluent_for_this_molecule=diluent
-
-        if is_multi_molecules_multi_diluents:
-            diluent_for_this_molecule.pop(molecule)
+            if isinstance(mole_fraction,dict):
+                diluent_for_this_molecule[diluent]=1-sum(list(mole_fraction.values()))
+            else:
+               diluent_for_this_molecule[diluent]=1-mole_fraction
+
+        # Getting diluents from mole_fraction
```
erwanp, line 538:
```suggestion
        # Adding the other molecules from the gas mixture as diluent for the calculation of this particular molecule 
```

erwanp (changes requested):
Hello; changes look very good. 

Last question : how do we test it, in particular the diluent part ? (and make sure it's always properly working)

One idea:  embed your diluent_for_this_molecule() dictionary in a function ; call this function from one test routine in a loop dealing with many input : 

Ex :
```
mole_fractions ={'CO2':0.2, 'CO':0.2', diluent='air'} 
# loop that simulates the calc_spectrum loop 
for molecule, mole_fraction in mole_fractions:
    diluent_for_this_molecule = get_diluent_for_this_molecule(molecule, mole_fractions, diluent)
    #in the real code this is where calc_spectrum_one_molecule() is called 
    if molecule == 'CO2':
        assert diluent_for_this_molecule = ...  # etc 
    if molecule == 'CO':
        assert diluent_for_this_molecule = ...  # etc 
```

erwanp:
For the test error : it comes from ` radis/test/lbl/test_calc.py/test_calc_spectrum_multiple_molecules_wstep_auto()` ; did you run this function locally too ?

1someshverma:
yes, I have tested it locally and it is giving the correct result.

erwanp:
OK I'll tell locally too

1someshverma:
And, for the diluent part, I will complete tests for it.

1someshverma:
@erwanp sir, if you have tested my latest PR on your system and if it is passing all the tests please let me .

minouHub (approved):

minouHub:
Sorry to add work but some cases are not working properly.

This code provides an output, but with a warning:
```python
s_dilCO2 = calc_spectrum(2000, 2010,         # cm-1
                  molecule='CO',
                  isotope='all',
                  pressure=1.01325,   # bar
                  Tgas=1000,
                  mole_fraction=0.1,
                  diluent ={'air':0, 'CO2':0.9},
                  )
s_dilCO2.plot('absorbance')
```
```
0.03s - Spectrum calculated
c:\users\nicolas minesi\python\radis\misc\warning.py:369: AccuracyWarning: Broadening Coefficient of CO2 not present in database using broadening coefficient of air instead.
  warnings.warn(WarningType(message))
c:\users\nicolas minesi\python\radis\lbl\broadening.py:347: AccuracyWarning: Broadening Coefficient of CO2 not present in database using broadening coefficient of air instead.
  warnings.warn(
```

- [ ] However, the CO2 parameters **are available in HITRAN**. Is there a problem with my database? If yes, the best thing to do is to give the command to the user to download the full database.
- [ ] Also. A single warning is necessary "Gamma_CO2 and n_CO2 are not present in the database"

This does *not* provide an output:
```python
s_dilCO2 = calc_spectrum(2000, 2010,         # cm-1
                  molecule='CO',
                  isotope='all',
                  pressure=1.01325,   # bar
                  Tgas=1000,
                  mole_fraction=0.1,
                  diluent ={'CO2':0.9},
                  )
s_dilCO2.plot('absorbance')
```
```
c:\users\nicolas minesi\python\radis\misc\warning.py:369: AccuracyWarning: Broadening Coefficient of CO2 not present in database using broadening coefficient of air instead.
  warnings.warn(WarningType(message))
c:\users\nicolas minesi\python\radis\lbl\broadening.py:347: AccuracyWarning: Broadening Coefficient of CO2 not present in database using broadening coefficient of air instead.
  warnings.warn(

  File c:\users\nicolas minesi\python\radis\lbl\broadening.py:337 in pressure_broadening_HWHM
    ** diluent_broadening_coeff["n_" + diluent_molecule.lower()]

KeyError: 'n_co2'


During handling of the above exception, another exception occurred:

  File C:\Anaconda\envs\radis_dev\lib\site-packages\spyder_kernels\py3compat.py:356 in compat_exec
    exec(code, globals, locals)

  File c:\users\nicolas minesi\python\radis\test\lbl\untitled0.py:33
    s_dilCO2 = calc_spectrum(2000, 2010,         # cm-1

  File c:\users\nicolas minesi\python\radis\lbl\calc.py:531 in calc_spectrum

  File c:\users\nicolas minesi\python\radis\lbl\calc.py:891 in _calc_spectrum_one_molecule
    s = sf.eq_spectrum(

  File c:\users\nicolas minesi\python\radis\lbl\factory.py:817 in eq_spectrum
    self._calc_broadening_HWHM()

  File c:\users\nicolas minesi\python\radis\lbl\broadening.py:911 in _calc_broadening_HWHM
    self._add_voigt_broadening_HWHM(

  File c:\users\nicolas minesi\python\radis\lbl\broadening.py:1079 in _add_voigt_broadening_HWHM
    wv, wl, wg = voigt_broadening_HWHM(

  File c:\users\nicolas minesi\python\radis\lbl\broadening.py:517 in voigt_broadening_HWHM
    gamma_lb = pressure_broadening_HWHM(

  File c:\users\nicolas minesi\python\radis\lbl\broadening.py:353 in pressure_broadening_HWHM
    (airbrd * pressure_atm * diluent["air"])

KeyError: 'air'
```

- [ ] Please also check this one after previous bullets

minouHub:
ok, so I added a warning message to debug people with incomplete databases #9e13cc74 and a small physics related test #7fe3c449

Final bullet for me: why do we need to calculate the broadening twice, and hence raise two identical warnings?

```python
s_dil = calc_spectrum(2000, 2010,         # cm-1
                  molecule='CO',
                  isotope='all',
                  pressure=1.01325,   # bar
                  Tgas=1000,
                  mole_fraction=0.1,
                  # diluent ='CO2',
                  diluent ={'air':0.5, 'H2O2':0.4},
                  )
s_dil.plot('absorbance')
```

```
0.02s - Spectrum calculated
c:\users\nicolas minesi\python\radis\misc\warning.py:369: HighTemperatureWarning: HITRAN is valid for low temperatures (typically < 700 K). For higher temperatures you may need HITEMP or CDSD. See the 'databank=' parameter
  warnings.warn(WarningType(message))
c:\users\nicolas minesi\python\radis\misc\warning.py:369: AccuracyWarning: Broadening Coefficient of H2O2 not present in database using broadening coefficient of air instead. 
!!! Solution: Try using once `use_cached='regen'' in calc_spectrum!!!
  warnings.warn(WarningType(message))
c:\users\nicolas minesi\python\radis\lbl\broadening.py:347: AccuracyWarning: Broadening Coefficient of H2O2 not present in database using broadening coefficient of air instead. 
!!! Solution: Try using once `use_cached='regen'' in calc_spectrum!!!
  warnings.warn(
```

1someshverma:
Two functions are called in calculating the spectrum and using the broadening coefficients. Though when using  calc_spectrum we can give a warning in one of the functions but in other cases when someone is using one of these functions directly then an error will be not raised.

I'll try to find out a better way for it.

minouHub:
So I rechecked the origin of these two identical warnings used to specify that `broadening Coefficient of XXX not present in database using broadening coefficient of air instead. `

- It is raised first in `_calc_broadening_HWHM` when building the dict `diluent_broadening_coef`
==> nothing is done, just a warning is raised that this molecule was not found in `df`

- It is raised in `_add_voigt_broadening_HWHM` > `voigt_broadening_HWHM` > `pressure_broadening_HWHM` 
==> gamma(T) is calculated and the function `pressure_broadening_HWHM` takes the broadening coefficient of air

In my opinion only one of them is necessary. I commented the one in `pressure_broadening_HWHM`. If @erwanp approves there is no problem for me to merge!

minouHub (approved):

minouHub:
the test_calc is not failing on my computer. Any ideas here?

1someshverma:
It's weird i also faced the same issue earlier then creating the new copy from my branch worked for me.

minouHub (commented):
@anandxkumar this assert is failing. Can you remind us why you put this test please? Especially why the value 0.013? Also, seems odd to me that the test could pass with "N/A"

minouHub:
The last review of @erwanp has been implemented. Only the "lint" tests are not passing 
==> merging the PR

anandxkumar:
> @anandxkumar this assert is failing. Can you remind us why you put this test please? Especially why the value 0.013? Also, seems odd to me that the test could pass with "N/A"

@minouHub  which test case are you referring to?

radis/test/lbl/test_calc.py
```diff
@@ -896,6 +902,20 @@ def test_diluent_invalid(verbose=True, plot=False, *args, **kwargs):
         in str(err.value)
     )
 
+def test_diluents_for_molecule():
```
erwanp, line 905: please comment this code, what's expected;  also refer to this PR

radis/lbl/broadening.py
```diff
@@ -326,37 +326,42 @@ def pressure_broadening_HWHM(
...
+                # if diluent_molecule != "air":
+                #     warnings.warn(
+                #         AccuracyWarning(
+                #             "Broadening Coefficient of "+diluent_molecule+" not present in database using broadening coefficient of air instead. \n!!! Solution: Try using once `use_cached='regen'' in calc_spectrum!!!"
+                #         )
+                #     )
+                
+                ## A warning is normally raised already in '_calc_broadening_HWHM' when constructing the dict diluent_broadening_coef
+                gamma_lb += ((Tref / Tgas) ** Tdpair) * (
+                    (airbrd * pressure_atm * diluent["air"])
```
erwanp, line 353: [IMPORTANT] 
@1someshverma  this line introduces a strong physics error. 

- you removed air_broadening (or you expect that it is added when gamma_air is not found 
- if there is a missing value ; for instance gamma_CO2 is not found, it is replaced by airbrd with the mole fraction of air. This is **wrong**. We can use gamma_air/airbrd as a placeholder for gamma_CO2; BUT it should keep the mole_fraction of CO2; not air. 
 
This is a critical error ; that's ok because it didn't make it to the public version; but it should have been picked by our test suite. 

This means  : 
- we need to fix the error 
- we need to fix the test suite   @minouHub 
- also, try not using try/except statements in general.

radis/lbl/broadening.py
```diff
@@ -888,11 +893,17 @@ def _calc_broadening_HWHM(self):
         # diluent and their broadening coeff dictionary
         diluent_broadening_coeff = {}
         for key in diluent:
-            if key != "air":
+            try:
```
erwanp, line 896: this try/except is also introducing many room for failures. Do we really need it ?

1someshverma: I think this is required as it just gives a warning in broadening coefficients are not present.

radis/lbl/broadening.py
minouHub: Test fixed by @erwanp in #589. @1someshverma, please replace `diluent["air"]` by `diluent_mole_fraction`

radis/lbl/broadening.py
erwanp: ok !

PR #557, opened 2023-04-03
Added example for examples gallery

imperorrp:
Added HITEMP H20 spectra calculations for example gallery

### Description

This pull request is to address [issue 136](https://github.com/radis/radis/issues/136), adding a HITEMP H20 spectra calculation example using the calc_spectrum function. This is already a [Radis Lab example](https://github.com/radis/radis-lab/blob/main/examples/calculate_H2O_HITEMP.ipynb) that does the same, but was not present in the main Radis repository's examples gallery yet. 

Fixes issue https://github.com/radis/radis/issues/136

### DIFF
--- a/examples/hitemp_h20_calculations.py
+++ b/examples/hitemp_h20_calculations.py
@@ -5,8 +5,10 @@
 Calculate a High Temperature H20 Spectrum
 =========================================
 
-This example uses the calc_spectrum function to calculate the H20 spectra at 3000K using the 
-HITEMP databank. Wavenumbers are set between 3125 and 3703 (corresponding to a wavelength range between 2700nm - 3200nm). 
+This example demonstrates using the calc_spectrum function to calculate the H20 spectra at 3000K using the 
+HITEMP databank. Due to HITEMP databank file size considerations, this example has been preset to using the HITRAN database instead. 
+To use HITEMP files for accurate results at high temperatures like the one in this example, modify databank='hitemp' in the code below. 
+Wavenumbers are set between 3125 and 3703 (corresponding to a wavelength range between 2700nm - 3200nm). 
 
 Warning: Please be aware that the HITEMP databank files for H20 is pretty large. Expect ~10Gb. See more under HITEMP [here](https://radis.readthedocs.io/en/latest/lbl/lbl.html#line-databases)
 """
@@ -23,7 +25,7 @@ s = calc_spectrum(wavenum_min = 3125 / u.cm,
                   mole_fraction = 1,
                   path_length = 1 * u.cm,
                   Tgas=3000, 
-                  databank='hitemp',
+                  databank='hitran',  #change to databank='hitemp' for accurate results, but be aware of size requirements (see warning above)
                   verbose=3,        # adding some extra info for the first computation
                   )
 #Without verbose=False this will show all the input parameters. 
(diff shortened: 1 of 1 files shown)


minouHub (approved):

erwanp:
Hello thanks for suggesting this. However downloading the Hitemp H2O database might download huge amount of data, and we don't want to run this on each commit (save data, save trees).

minouHub:
actually it did it for me and took some time but I thought I had a wifi problem. My bad. @imperorrp Could you restrict this calculation for a smaller waverange where there is some absorbance?

minouHub:
Indeed. I thought I had a wifi problem ==> my bad.
@imperorrp could you please restrict this calculation to a small waverange, but make sure H2O is absorbing in this region. Also, add a comment in the code to warn the user about the potential size of the data

imperorrp:
> Hello thanks for suggesting this. However downloading the Hitemp H2O database might download huge amount of data, and we don't want to run this on each commit (save data, save trees).

Ah, right. I'm not sure if this can be done, but is there a way to circumvent this? Apart from caching this data, which would still involve downloading the whole size at least once.

erwanp:
It is not possible. Even if you require a small wavenumber range the full database will be downloaded because Hitemp unlike hitran doesn't allow to retrieve partial data.

minouHub:
Alright, maybe the database to "hitran" and then add a comment "change to 'hitemp' for an accurate calculation above 300 K but be aware of the large database size"

imperorrp:
> Indeed. I thought I had a wifi problem ==> my bad. @imperorrp could you please restrict this calculation to a small waverange, but make sure H2O is absorbing in this region. Also, add a comment in the code to warn the user about the potential size of the data

Just did that- changed wavenumbers to make the wavelength range smaller and added a comment warning users about the size.

imperorrp:
> Alright, maybe the database to "hitran" and then add a comment "change to 'hitemp' for an accurate calculation above 300 K but be aware of the large database size"

Did this

erwanp:
Please weight-in the interest of doing such a thing. It's not because RADIS can compute 100+ species that the Docs & Examples should do it. 

What does it bring that's not already in, for instance, this example (HITEMP CO) : https://radis.readthedocs.io/en/latest/auto_examples/plot_hitemp_spectrum.html#sphx-glr-auto-examples-plot-hitemp-spectrum-py 


If you think it's is a valuable addition, let's do it ! 

---

Also notice we have an example using CO2 + H2O  HITRAN, but currently deactivated (https://radis.readthedocs.io/en/latest/auto_examples/calc_hitran_full_range.html#sphx-glr-auto-examples-calc-hitran-full-range-py )  i.e. calculations are not performed (Sphinx requires examples to start with `plot_*.py` so that the calculations are ran.

minouHub:
Ok, indeed this PR would just add a duplicate of currently existing examples that do more stuff:

- https://radis.readthedocs.io/en/latest/auto_examples/plot_compare_CO_geisa_hitran.html#sphx-glr-auto-examples-plot-compare-co-geisa-hitran-py
- https://radis.readthedocs.io/en/latest/auto_examples/plot_nonequilibrium_co_spectrum.html#sphx-glr-auto-examples-plot-nonequilibrium-co-spectrum-py
- https://radis.readthedocs.io/en/latest/auto_examples/plot_exomol_spectrum.html#sphx-glr-auto-examples-plot-exomol-spectrum-py

@imperorrp, if you want to repost a PR related to #136 you are welcome. A few boxes are not ticked

Issue #564, opened 2023-04-06
Linting Tests fail


erwanp:
### 🐛 Describe the bug

https://app.travis-ci.com/github/radis/radis/jobs/599466669#L249

### 💡 Possible solutions

Seems to be related to isort/poetry compatibility; upgrading to latest isort could fix the problem : 
 https://github.com/PyCQA/isort/releases/tag/5.11.5

Update https://github.com/radis/radis/blob/develop/.pre-commit-config.yaml

### 🎲 Radis version

0.14

### 💻 Operating system

Linux

minouHub:
That's indeed a good first issue but raising a lot of questions from new RADIS contributors. So I preferred to try to sort it (open for a month).

PR #568, opened 2023-05-07
updated isort for uncompatibility issue with Poetry

minouHub:
As suggested by @erwanp, the new version of `isort` should sort the problems of the liting tests.


minouHub:
to fix #564

PR #569, opened 2023-05-07
No upper bound on numpy version

### LINKED ISSUE #548: Upper bound version constraint on Numpy?
I was wondering why does `radis` sets an upper bound version constraint on Numpy in its setup file?

### 🐛 Describe the issue
When adding `radis` to a Python environment where a version of Numpy larger than `1.22.3` is required (either from the current project, or from other dependencies), the package manager is going to run into a resolution error as [`radis` requires `numpy<=1.22.3`](https://github.com/radis/radis/blob/b928f3e2f6956a2b50cc202b5968ca73cfc42f2c/setup.py#L256).

### 🎯 Expected behavior
Usually, it is [not recommended](https://iscinumpy.dev/post/bound-version-constraints/) to have upper bound version constraint for dependencies, but I don't know there are specific reasons here to avoid Numpy versions larger than `1.22.3`?

### 💡 Possible solutions
Unless `radis` is incompatible with `numpy>1.22.3`, we could simply update the constraint set by [this line in the setup.py file](https://github.com/radis/radis/blob/b928f3e2f6956a2b50cc202b5968ca73cfc42f2c/setup.py#L256)  (similarly in the [development conda environment file](https://github.com/radis/radis/blob/b928f3e2f6956a2b50cc202b5968ca73cfc42f2c/environment.yml#L18)) to:
```python
"numpy>=1.22.3"
```
(perhaps a typo changed `>` into `<`?)

If I change `numpy>=1.22.3` to `numpy`, in `environment.yml` and let conda determine what version of Numpy to install, it picks `1.23.5` (latest version at the time of writing is `1.24.2`). If I run the tests with this change, I have all tests passing.

---
<sub>*Tip : you can vote for [New Features](https://github.com/radis/radis/issues/79) on https://feathub.com/radis/radis*
*Also check for the [Long Term Todo List](https://github.com/radis/radis/issues/53)*
</sub>

minouHub:
To fix #548 

I manage to install `radis` locally with this new environment having no restriction on `numpy` version. Likely a fix from `vaex` as pointed out by @erwanp (see #548).

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -15,7 +15,7 @@ dependencies:
 - h5py   # load HDF5
 - joblib  # for parallel loading of SpecDatabase
 - matplotlib
-- numpy<=1.22.3
+- numpy
 - numba  # just-in-time compiler
 - pandas>=1.0.5
 - plotly>=2.5.1  # for line survey HTML output
--- a/radis/test/lbl/test_calc.py
+++ b/radis/test/lbl/test_calc.py
@@ -786,7 +786,7 @@ def test_calc_spectrum_multiple_molecules_wstep_auto(
         verbose=verbose,
     )
     wstep_s = s.get_conditions()["wstep"]
-    
+
     s_CO = calc_spectrum(
         wavelength_min=4165,
         wavelength_max=5000,  # cm-1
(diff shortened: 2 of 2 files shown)


minouHub:
Commit be78264095b12bdf2c41cc8c157f81d78407b88a to:

1. sort incompatibilities
2. uncomment tests that were commented on in the debugging process of #570

PR #570, opened 2023-05-08
Test failing on `test_calc`

minouHub:
**:bug: Original problem:**

The test `test_calc` has been failing for a few weeks now. This is hindering the progress of several PR and could confuse new contributors. Here is one of the [recent test log ](https://app.travis-ci.com/github/radis/radis/jobs/601699540#L1838). The culprit is [this test](https://app.travis-ci.com/github/radis/radis/jobs/601699540#L1906). From the comment there, the function has a valuable goal.

> Tests multiple molecules spectrum for wstep = 'auto' and checks that **minimum** wstep value is selected with resample = "intersect" 

There are two problems. 

1. If the two merged spectra have different resolutions, the stored conditions will be `wstep = "N/A"` and not the minimum `wstep` of the spectra. Thus, almost all the time, "N/A" will be the output which is not acceptable given that we know it should be the minimum `wstep`. 
2. For unknown reasons, this test is failing on Travis and not on local installations. 

💻  **Proposed solution**

- [ ] Rewrite the resample function so that the new `wstep` is written in the conditions of the new spectrum
- [ ] Hope for the unknown Travis automatic test to pass now


radis/spectrum/spectrum.py
```diff
@@ -3740,6 +3747,7 @@ def resample(
         # Update stored_waveunit to new unit
         if unit != stored_waveunit:
             s.conditions["waveunit"] = unit
+        s.conditions["wstep"] = wstep_new
```
anandxkumar, line 3750: Nice!

radis/test/lbl/test_calc.py
```diff
@@ -785,10 +785,22 @@ def test_calc_spectrum_multiple_molecules_wstep_auto(
...
+        path_length=1,  # cm
+        wstep="auto",
+        databank="hitran",  # or use 'hitemp'
+        verbose=verbose,
+    )
 
     # Check calculation went fine:
     assert set(s.conditions["molecule"]) == set(["CO2", "CO"])
-    assert s.get_conditions()["wstep"] in ("N/A", 0.013)
+    assert np.isclose(s.get_conditions()["wstep"], s_just_CO.get_conditions()["wstep"])
```
anandxkumar, line 803: I also believe np.isclose is a better approach for assertion since values can slightly alter over time.

minouHub: I just don't get it... The test is just not failing on my machine!

anandxkumar: The latest commit resolved all tests except this one. there are two different values we are getting, `0.021 and 0.013`. Could it be due to the order in which the molecules are processed? 
Like mole_fraction={"CO": 0.1, "CO2": 0.1}, if CO is processed first then wstep becomes `0.013` similar to what we are getting in `s_just_CO`, and if CO2 is processed first then wstep becomes `0.021`?

minouHub: pure CO2 spectrum wstep is `0.19` so that's not it

radis/test/lbl/test_calc.py
```diff
@@ -777,7 +778,7 @@ def test_calc_spectrum_multiple_molecules_wstep_auto(
         wavelength_min=4165,
         wavelength_max=5000,  # cm-1
         isotope="1,2,3",
-        pressure=1.01325,  # bar
+        pressure=10.01325,  # bar
```
minouHub, line 781: multiplying the pressure by 10 increases the wstep of each spectrum by 10. So it is effectively something happening in the calculation of the spectra

minouHub:
yes! test_calc is passing the tests now! The origin of the problem: the diluent was not updated in the CO and CO2 spectra ...

anandxkumar:
Finally @minouHub  :rocket: . Now the only thing remains is fixing the linting.

minouHub:
ahah don't worry I already did the merge of another #568 for that!

image

PR #571, opened 2023-05-21
Removed manual exceptions in exomolapi and fixed broadf option

YuiKasagi:
### Description

(This is my first pull request to radis. I am a user of radis through exojax which is developed by Hajime Kawahara. Thank you in advance.)
- This pull request is to address a wrong value in [def file of SO2 in Exomol](https://www.exomol.com/db/SO2/32S-16O2/ExoAmes/32S-16O2__ExoAmes.def). 
```
14999.68                                                                        # Maximum wavenumber (in cm-1)
```
I added a manual exception in `exomol.api` to use the actual value (8000.0).

- I also fixed the use of `broadf` option. According to the parameter description, if False, the default parameters in .def file are used. I fixed the code to apply this option not to download .def file if False.

### DIFF
--- a/radis/api/exomolapi.py
+++ b/radis/api/exomolapi.py
@@ -129,58 +129,12 @@ def read_def(deff):
             unc = int(dat["VAL"][i]) == 1
 
     # SOME DEF FILES CONTAINS ERRORS. THESE ARE THE EXCEPTIONS
-    if deff.stem == "12C-16O2__UCL-4000":
-        ntransf = 20
-    if deff.stem == "1H3-16O_p__eXeL":
-        # fixes https://github.com/radis/radis/pull/531
-        ntransf = 100
-    if deff.stem == "14N-1H3__CoYuTe":
-        maxnu = 20000.0
-    if deff.stem == "40Ca-16O-1H__OYT6":
-        ntransf = 18
-        maxnu = 36000.0
-    if deff.stem == "32S-16O2__ExoAmes":
-        maxnu = 8000.0
     if deff.stem == "1H-35Cl__HITRAN-HCl":
         quantum_labels = ["v"]
         # See https://github.com/HajimeKawahara/exojax/issues/330
     if deff.stem == "16O-1H__MoLLIST":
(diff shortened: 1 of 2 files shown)


minouHub:
Hello @YuiKasagi and thanks for your first contribution!
We could also contact the ExoMol admins to let them know that this value is incorrect, or is it something that you did already?

YuiKasagi:
Thank you for your comment @minouHub !
I have not reported this issue to the ExoMol admins yet, but I agree to do so. But I do not know how long it will take them to fix it, so I think it is better to use this manual exception for now.

YuiKasagi:
I have reported the wrong values in .def file to the ExoMol team, including those reported so far. (https://github.com/ExoMol/exomol.com/issues/20)

YuiKasagi:
New commit: 
- Exomol team has corrected the values in .def files of SO2 as well as H3O+, NH3, and CaOH.
    --> Because we no longer need a manual exception, I remove that part from `exomolapi.py`. 
    - I confirmed that the new .def files do not interrupt downloading the transition files from ExoMol.
    - (I did not delete manual exceptions for quantum_labels of HCl and OH, because I could not determine their necessity.)

So now, this pull request is for removing manual exceptions and extending the use of `broadf` option.

It would be much appreciated if you could confirm these changes.

erwanp:
@YuiKasagi , as I understand you need this to be in the public RADIS version as soon as possible (for Exojax depenedncies) ?

YuiKasagi:
Thank you @erwanp. Yes, we would appreciate it if you could include this feature to the public RADIS version before the next ExoJAX release (v1.5).

erwanp:
When is Exojax 1.5 due ?

YuiKasagi:
Kawahara-san is now trying to implement the radiative transfer including the scatter effect in ExoJAX, and he said that the release date has not been decided yet. It will probably be in about 3-4 months.

erwanp:
Thank you. We'll be deploying RADIS 0.15 in the coming weeks so it will be out in time !

PR #574, opened 2023-06-15
Allowing to use non numeric columns in HITRAN

YuiKasagi:
### Description

This pull request is to use additional columns of the HITRAN database. I wanted to use the column "ierr" (uncertainty indices of transition parameters, as described in https://hitran.org/docs/uncertainties/) in ExoJAX, but it seems to be automatically removed in the downloading process.
I would like to make the existing option `drop_non_numeric` available in the function `download_and_parse`.

(I am sorry but the commit comment was wrong. It is not HITEMP, but HITRAN.)

### DIFF
--- a/radis/api/hitranapi.py
+++ b/radis/api/hitranapi.py
@@ -1175,7 +1175,7 @@ class HITRANDatabaseManager(DatabaseManager):
         else:
             raise NotImplementedError()
 
-    def download_and_parse(self, local_file, cache=True, parse_quanta=True):
+    def download_and_parse(self, local_file, cache=True, parse_quanta=True, drop_non_numeric=True):
         """Download from HITRAN and parse into ``local_file``.
         Also add metadata
 
@@ -1320,6 +1320,7 @@ class HITRANDatabaseManager(DatabaseManager):
                 df,
                 molecule=molecule,
                 parse_quanta=parse_quanta,
+                drop_non_numeric=drop_non_numeric
             )
 
             wmin_final = min(wmin_final, df.wav.min())
--- a/radis/api/hitranapi.py
+++ b/radis/api/hitranapi.py
@@ -292,6 +292,7 @@ def post_process_hitran_data(
     verbose=True,
     drop_non_numeric=True,
     parse_quanta=True,
+    add_HITRAN_uncertainty_code=False
 ):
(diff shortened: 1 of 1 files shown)


minouHub:
This is nice to add this argument, but I'm not sure it is tackling the right problem. Seems like the uncertainty column is removed because it's not numeric, which is not the case. I would prefer an explicit argument to choose whether or not the uncertainty is downloaded (e.g. `add_HITRAN_uncertainty_code = True`)

Also, to avoid the tests to fail, please add LiOH, a new ExoMol molecule, see [11a3713](https://github.com/radis/radis/pull/571/commits/11a371325d9e78b78f9da35ffeb25d69e9daba1f)


@menasrac please take a look at this pull request. It's a good idea and you should probably implement it in other databases if available.

menasrac:
Thank you I will have a look at it. It may give me some ideas @minouHub

YuiKasagi:
Sorry, I mistakenly merged the branch. Could you tell me how I can undo this operation? I just wanted to update my branch including the updates of develop branch.

YuiKasagi:
I am sorry, but I am confusing how to push my branch. I just wanted to update my branch including the updates of develop branch. My branch seems to reflect [11a3713](https://github.com/radis/radis/pull/571/commits/11a371325d9e78b78f9da35ffeb25d69e9daba1f), but the tests failed again...
I am sorry to bother you, but could you tell me how to solve this? @minouHub 
Is it better to recreate the branch?

minouHub:
I think you imported the develop branch correctly. Check the test log of Travis, it's the linting that causes the fail. Quick explanation: Two tests are run on RADIS (1) physics/code tests and (2) linting. The linting might be annoying at first but is a very convenient tool to keep the code tidy (e.g. clear the commit logs from whitespace) and make sure there is no code change across platforms (e.g. Widnows vs Linux). You should run the pre-commit on your OS, see [this page](https://radis.readthedocs.io/en/latest/dev/developer.html#code-linting)

YuiKasagi:
Thank you very much for your kind support! I understand the system of pre-commit.

Based on your first comment, I changed the code to extract the “ierr” column by setting `add_HITRAN_uncertainty_code = True`.
I would appreciate it if you could review it.

PR #576, opened 2023-06-26
[Reduced Memory Use] calculating spectrum using vaex

### LINKED ISSUE #118: Reduce Memory usage

1someshverma:
This  PR addresses memory use for computing Spectrum, existing Pandas implementation though efficient for small databases, doesn't perform well on larger line database.
To solve this issue, vaex is used which performs exceptionally well for memory uses.
The time efficiency of vaex is also better than Pandas.

image

Fixes #118


minouHub:
Hello, 
Here is first a review of the results, I didn't check the inside of your work and the coding quality yet.

**1/** you should try to sort the following performance decrease for small spectra. On my machine, the CO spectrum simulated from 2000 to 2010 cm-1 takes 0.24 s with `vaex` and 0.04 s with `pandas`
- [ ] Improve time performance for small spectra

```
s = calc_spectrum(2000, 2010,         # cm-1
                  molecule='CO',
                  isotope='1,2,3',
                  pressure=1.01325,   # bar
                  Tgas=1000,
                  mole_fraction=0.1,
                  diluent = "air",     
                  verbose = 1,
                  engine = "vaex"
                  )
s.plot('absorbance')
```
with `pandas`:
```python
0.04s - Spectrum calculated
```
with `vaex`:
```python
0.24s - Spectrum calculated
```

**2/** When I tried to find the origin of the time difference, I ran the `calc_spectrum` function with `verbose = 3` and I got a bug. Could you please implement more tests of this new implementation and sort the bug?

- [ ] Add more tests
- [ ] Debug `verbose = 3` in `calc_spectrum`
 


```python
s = calc_spectrum(2000, 2010,         # cm-1
                  molecule='CO',
                  isotope='1,2,3',
                  pressure=1.01325,   # bar
                  Tgas=1000,
                  mole_fraction=0.1,
                  diluent = "air",     
                  verbose = 3,
                  engine = "vaex"
                  )
s.plot('absorbance')
```
```python
 File "c:\users\nicolas minesi\python\radis\api\untitled0.py", line 9, in <module>
    s = calc_spectrum(2000, 2010,         # cm-1

  File "c:\users\nicolas minesi\python\radis\lbl\calc.py", line 537, in calc_spectrum

  File "c:\users\nicolas minesi\python\radis\lbl\calc.py", line 902, in _calc_spectrum_one_molecule
    s = sf.eq_spectrum(

  File "c:\users\nicolas minesi\python\radis\lbl\factory.py", line 802, in eq_spectrum
    self._cutoff_linestrength()

  File "c:\users\nicolas minesi\python\radis\lbl\base.py", line 3265, in _cutoff_linestrength
    Nlines_cutoff / len(df.S) * 100, cutoff

TypeError: object of type 'Expression' has no len()
```

1someshverma:
For smaller data frames, I tested some simple operations like sorting and finding the minimum on a simple data frame.
Link to Kaggle notebook https://www.kaggle.com/code/someshverma/notebook2e47fa84fe
And, it clearly shows that in general for smaller data frames performance of pandas is better.

I also tested this on the local system and similar results are obtained. So, It appears that for smaller data frame performance of Pandas is better than Vaex . Though the performance of vaex improves with larger data frames.

I will try if I can improve Vaex implementation for smaller spectra.

1someshverma:
I already have some test cases , i will commit those and add more test cases.

1someshverma:
I have resolved issue related to verbose in my latest PR

minouHub:
I pushed a minor commit to show you that some improvements can still be made (in this case, how the sum is computed). I know it's a tedious task, but it is surely valuable for RADIS (but also for you to learn). So please try to improve each line of code taking some time in the computation. On my computer, the commit I pushed the time taken to calculate 25 spectra decreased from `6.0 s `to `5.4 s`. I strongly recommend you to use `verbose=3` to track where the time is taken.

My code to see the impact of a change: 
```python
import time

t0 = time.time()
for i in range(25):
    s = calc_spectrum(2000, 2010,         # cm-1
                      molecule='CO',
                      isotope='1,2,3',
                      pressure=1.01325,   # bar
                      Tgas=1000,
                      mole_fraction=0.1,
                      databank='hitemp',  # or 'hitemp'
                      diluent = "air",     
                      verbose = 0,
                      engine = "vaex"
                      )
t1 = time.time()
print(t1 -t0)
```

1someshverma:
Last commit reduces calculation time for the code below by 10 seconds

```

import time
t0=time.time()

s = calc_spectrum(1500, 2500,         # cm-1
                      molecule='H2O',
                      isotope='1,2,3',
                      pressure=1.01325,   # bar
                      Tgas=700,           # K
                      mole_fraction=0.1,
                      wstep='auto',
                      path_length=1,      # cm
                      databank='hitemp',  # or 'hitemp', 'geisa', 'exomol'
                      engine='vaex',
                      )

t1=time.time()

print('Time taken : '+str(t1 - t0))

```


And 0.5 seconds for the code 

```
import time
t0=time.time()
s = calc_spectrum(2000, 2010,         # cm-1
                  molecule='CO',
                  isotope='1,2,3',
                  pressure=1.01325,   # bar
                  Tgas=1000,
                  mole_fraction=0.1,
                  diluent = "air",
                  verbose = 3,
                  engine = "vaex"
                  )
t1=time.time()


print('Time taken : '+str(t1-t0))
```

I am working to further improve the results.

minouHub:
Very interesting last commit, we were actually talking about something similar at lunch with @erwanp.

@1someshverma, so you define a threshold when the database is either converted for the *calculations* into pandas. This threshold is 2e5 entries in the database. Please explain how you picked this threshold. A graph might be sufficient. If you used a code to determine this threshold, could you please share it?

minouHub:
Concerning your test cases: they are not passing on my machine!
If I run the test `test_vaex_and_pandas_dataframe_load_databank()`, then the test fails at `assert np.all(df1[column].to_numpy() == df2[column])` because the column `jl` is empty in the vaex format. Could you please check this?

1someshverma:
I have tested this on my laptop and it passed the test case , i will test it again and sort this issue.

1someshverma:
@minouHub After some of the improvement in the time performance for Vaex , now it perform better than better pandas after 0.2 M lines . I will share the graph in support of this soon.

1someshverma:
@minouHub  `test_vaex_and_pandas_dataframe_load_databank()`  is passing on my laptop and also on Travis. Can you share Python version and vaex version on your system so that I can check if it is related to different versions .

1someshverma:
These are the updated time graph, and it shows that the total time taken for vaex starts to perform better than pandas after 9e5 (maybe I made some mistake in computing the time earlier). 
I will update the threshold with 9e5 .

As you can see compared to the previous time graph, the time efficiency is improved for Vaex both for calculations and overall.
image
image

minouHub:
> @minouHub `test_vaex_and_pandas_dataframe_load_databank()` is passing on my laptop and also on Travis. Can you share Python version and vaex version on your system so that I can check if it is related to different versions .

You must be right, if Travis is passing I should be passing as well. I'll try to sort this on my side. 

> These are the updated time graph, and it shows that the total time taken for vaex starts to perform better than pandas after 9e5 (maybe I made some mistake in computing the time earlier).

Could you please the code to generate these figure so I can check if the same order of magnitude is obtained on other computer?

1someshverma:
This is the code that I used for obtaining the time taken for calculating the spectrum 

```

import time
t0=time.time()

s = calc_spectrum(2450, 2500,         # cm-1
                      molecule='H2O',
                      isotope='1,2,3',
                      pressure=1.01325,   # bar
                      Tgas=700,           # K
                      mole_fraction=0.1,
                      wstep='auto',
                      path_length=1,      # cm
                      databank='hitemp',  # or 'hitemp', 'geisa', 'exomol'
                      engine='vaex',
                      )

t1=time.time()

print('Time taken : '+str(t1 - t0))

```

I  started with wmin = 2450 and then decreased it in steps to 10 then increased step length.

I think it may be slightly machine dependent but the difference should be quite small.

PR #577, opened 2023-07-01
Implemented Kurucz with an example 

menasrac:
### Description

Added a new AdB Kurucz class to the API  with an atomic spectrum plotting program in the examples section and tests for the methods of the new API class.


docs/spectrum/spectrum.rst
```diff
@@ -377,6 +377,41 @@ You can also show the conditions only with
 
 .. _label_spectrum_plot:
 
+Plotting
```
minouHub: is this from your other PR?

PR #306, opened 2021-07-05
Add/vaex for hitemp

erwanp:
### Description

Adding vaex to load HDF5 files. Proof-of-concept : [test_vaex.ipynb](https://github.com/radis/radis-benchmark/blob/master/manual_benchmarks/test_vaex.ipynb)

Built on top of #281 : 

- [x] implemented different engines in HDF5 cache files & generation : vaex and h5py
- [x] added vaex filtering when loading database with wmin, wmax, or isotpes
- [ ] vaex is the default library to load HITEMP .hdf5 files  👉  "pytables" remains the default for the moment
- [x] auto update of your existing .hdF5 databases

```python
from radis.io.hitemp import fetch_hitemp
fetch_hitemp("OH", engine="pytables")       #  (if testing : make sure there is no vaex-formatted hdf5 already)
import radis; radis.AUTO_UPDATE_DATABASE = True
fetch_hitemp("OH", engine="vaex")   #  auto-converts to vaex-compatible HDF5

```

- [x] re-add comparison with already registered database (if exists)

Will be merged / rebased on #281 . 

Fixes #239 #295 

### TODO : 

- [ ] add multiple isotope support when querying vaex files
- [ ] add an auto-update test 
- [ ] write the parsed HITEMP files in Vaex-compatible / H5py format (see e3a1d734) 
- [x] Feature idea : for CO2 / H2O (multiple files) : only parse & download the files needed. Use the metadata keys wavenum_min; wavenum_max (?) in the database to see if new files should be downloaded ?  Could be extended for CDSD-4000 later.

Identified problems : 
- process sometimes stuck in a Spyder Ipython console (while converting to VAEX). File is converted properly, though, but only appears after the console is closed.   Similar to #338 and https://github.com/spyder-ide/spyder/issues/16183
- Some HDF5 files not closed properly if a crash happens : https://github.com/pandas-dev/pandas/issues/8692

### Performance improvement : 

Old (full loading and partial loading of HITEMP CO)

```python

In [2]: %timeit fetch_hitemp('CO', engine='pytables')
4.84 s ± 77 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [3]: %timeit fetch_hitemp('CO', engine='pytables', load_wavenum_min=2000, load_wavenum_max=3000)
814 ms ± 28 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

```

New:

```python
In [2]: %timeit fetch_hitemp('CO', engine='vaex')
[...]

PR #580, opened 2023-07-06
[Reduced Memory Use] calculating spectrum using vaex corrected

### LINKED ISSUE #118: Reduce Memory usage

1someshverma:
This is in continuation of PR #576 and resolve the issues in that PR.

This  PR addresses memory use for computing Spectrum, existing Pandas implementation though efficient for small databases, doesn't perform well on larger line database.
To solve this issue, vaex is used which performs exceptionally well for memory uses.
The time efficiency of vaex is also better than Pandas.
Contributions of this PR : 
- There is  a factor of 10x improvement in memory performance for vaex implementation  as compared to pandas implementation.
- vaex implementation is done for loading_databank() and its associated functions , some parts of fetch_databank() is also implemented using vaex.
- calc_spectrum() function is also written in vaex ,keeping the existing implementation of pandas.Also its associated functions are refactored .
- eq_spectrum() and as well non_eq_spectrum are rafactored for vaex .Also there associated functions are refactored .
- Thus , both spectrum under equilibrium as well non-equilibrium can be calculated for larger databases using Vaex implementation .

To use Vaex 
```python
from radis import config 
config["DATAFRAME_ENGINE"] = "vaex"
```

image

Code used to get this memory use comparison graph is 
```python
def compare_pandas_vs_vaex_memory():
    """
    Compare memory usage of `engine="vaex"` and `engine="pandas"` in calc_spectrum.
    Expected behavior is "vaex" using much less memory. This function takes tens of seconds to run.
    Returns
    -------
    None.
    """

    import tracemalloc

    for engine in ["pandas", "vaex"]:
        config["DATAFRAME_ENGINE"] = engine
        tracemalloc.start()
        s = calc_spectrum(
            1000,
            1500,  # cm-1
            molecule="H2O",
            isotope="1,2,3",
            pressure=1.01325,  # bar
            Tgas=1000,  # K
            mole_fraction=0.1,
            wstep="auto",
            databank="hitemp",  # or 'hitemp', 'geisa', 'exomol'
            verbose=0,
        )
        snapshot = tracemalloc.take_snapshot()
        memory = tracemalloc.get_traced_memory()
        tracemalloc.stop()

        # Some raw outputs
        print("\n******** Engine = {} ***********".format(engine))
        print(
            "Peak, current = {:.1e}, {:.1e} for {:} lines calculated".format(
                *memory, s.conditions["lines_calculated"]
            )
        )

        # More sophisticated
        print("*** List of biggest objects ***")
        top_stats = snapshot.statistics("lineno")
        for rank, stat in enumerate(top_stats[:3]):
            print("#{}".format(rank + 1))
            print(stat)

        # Clear for next engine in the loop
        tracemalloc.clear_traces()

```
As an example to illustrate vaex and pandas spectrum are same here is a plot showing difference between Radiance. 
diff

Code used to calculate difference between radiance 
```
from radis import config
from radis.spectrum import plot_diff
config["DATAFRAME_ENGINE"] = "vaex"
s_vaex = calc_spectrum(1900, 2300,         # cm-1
                  molecule='CO',
                  isotope='1,2,3',
                  pressure=1.01325,   # bar
                  Tgas=700,           # K
                  mole_fraction=0.1,
                  path_length=1,      # cm
                  )
config["DATAFRAME_ENGINE"] = "pandas"
s_pandas = calc_spectrum(1900, 2300,         # cm-1
                  molecule='CO',
                  isotope='1,2,3',
                  pressure=1.01325,   # bar
                  Tgas=700,           # K
                  mole_fraction=0.1,
                  path_length=1,      # cm
                  )

plot_diff(s_vaex, s_pandas)
```

Fixes #118


erwanp:
Hello @1someshverma can you add an example of spectral calculation with and without vaex, and how to switch from one to another ? Also, what are the conditions of the performance test ?

radis/lbl/loader.py
```diff
@@ -1323,21 +1336,49 @@ def fetch_databank(
...
-                for df in frames[1:]:
-                    attrs = intersect(attrs, df.attrs)
-                del attrs["iso"]  # added as a column (different for each line)
-                # Merge:
-                df = pd.concat(frames, ignore_index=True)  # reindex
-                df.attrs = attrs
-                self.params.dbpath = ",".join(local_paths)
+                if output == "pandas":
+                    for df in frames:
+                        if "iso" not in df.columns:
```
erwanp, line 1341:
```suggestion
                        # to save some memory, we have added "iso" as an attribute rather than a column if it is the same for all lines 
                        if "iso" not in df.columns:
```

radis/lbl/loader.py
```diff
@@ -1323,21 +1336,49 @@ def fetch_databank(
...
+                    for df in Frames[1:]:
+                        attrs = intersect(attrs, df.attrs)
+                    del attrs["iso"]  # added as a column (different for each line)
+                    # Merge:
+                    import vaex
+
+                    df = vaex.concat(Frames)  # reindex
+                    df.attrs = attrs
+
+                    self.params.dbpath = ",".join(local_paths)
```
erwanp, line 1379: @1someshverma @minouHub ; the "iso" as an attribute rather than a column is just a Pandas trick to save memory... Given than Vaex doesn't use any memory at all, I don't think it is necessary in the Vaex case

radis/lbl/loader.py
```diff
@@ -1392,23 +1434,39 @@ def fetch_databank(
             )
 
         # Always sort line database by wavenumber (required to SPARSE_WAVERANGE mode)
-        df.sort_values("wav", ignore_index=True, inplace=True)
+        if output == "pandas":
+            df.sort_values("wav", ignore_index=True, inplace=True)
+        elif output == "vaex":
+            try:
```
erwanp, line 1440: why is rewriting the attrs needed ? add comment

radis/test/lbl/test_factory.py
```diff
@@ -725,6 +725,57 @@ def test_diluents_molefraction(verbose=True, plot=False, *args, **kwargs):
     sf.eq_spectrum(Tgas=300, mole_fraction=0.4)
 
 
+def test_vaex_and_pandas_spectrum():
```
erwanp, line 728: add in docstrings what you're testing, also add the link to this PR (e.g. "testing Vaex as implemetned in https://github.com/radis/radis/pull/580"  )

radis/lbl/broadening.py
```diff
@@ -890,20 +892,40 @@ def _calc_broadening_HWHM(self):
...
+                    diluent_broadening_coeff["n_" + key.lower()] = df[
+                        "n_" + key.lower()
+                    ]
+                except NameError:
+                    if key != "air":
+                        self.warn(
+                            message="Broadening Coefficient of "
+                            + key
+                            + " not present in database using broadening coefficient of air instead. \n!!! Solution: Try using once `use_cached='regen'' in calc_spectrum!!!",
+                            category="AccuracyWarning",
```
erwanp, line 927: you're rewriting most of the code above,  there must be another way

radis/lbl/base.py
```diff
@@ -3245,7 +3283,14 @@ def _cutoff_linestrength(self, cutoff=None):
             ) from err
 
         # update df1:
-        self.df1 = pd.DataFrame(df[~b])
+        if self.dataframe_type == "pandas":
+            self.df1 = pd.DataFrame(df[~b])
+        elif self.dataframe_type == "vaex":
+            self.df1 = df[
+                ~b
+            ].materialize()  # If materialized is used time taken is reduced but memory efficiency is slightly reduced.
```
erwanp, line 3291: @1someshverma @minouHub to my understanding of vaex, this is a massive move : you're removing all virtual columns here, and adding them in memory.

1. Is it necessary ? 
2. If so (because CPU-time is slow for instance) we might want to have different modes; with a focus on saving memory or saving CPU-time.

radis/lbl/base.py
```diff
@@ -3341,16 +3386,16 @@ def _reinitialize(self):
...
-                    "MemoryUsageWarning",
-                )
+            if self.dataframe_type == "pandas":
+                df_size = self.df1.memory_usage(deep=False).sum()
+
+                if df_size > limit:
+                    self.warn(
+                        "Line database is large: {0:.0f} Mb".format(df_size * 1e-6)
+                        + ". Consider using save_memory "
+                        + "option, if you don't need to reuse this factory to calculate new spectra",
```
erwanp, line 3396: Maybe mention using Vaex mode to save memory ?

radis/test/lbl/test_loader.py
```diff
@@ -222,13 +222,175 @@ def test_custom_abundance(verbose=True, plot=False, *args, **kwargs):
...
+        verbose=0,
+    )
+    sf.warnings["MissingSelfBroadeningWarning"] = "ignore"
+
+    # Loading in vaex dataframe format
+    sf.load_databank("HITEMP-CO2-TEST", output="vaex", load_columns="all")
+    df1 = sf.df0
+
+    # Loadint in pandas dataframe format
+    sf.load_databank("HITEMP-CO2-TEST", load_columns="all")
```
erwanp, line 375:
```suggestion
    sf.load_databank("HITEMP-CO2-TEST", output="pandas", load_columns="all")
```
Learning : in tests never assume the default value of a parameter. Be explicit. Default might change in the future

radis/test/lbl/test_loader.py
```diff
@@ -222,13 +222,175 @@ def test_custom_abundance(verbose=True, plot=False, *args, **kwargs):
     assert s.compare_with((3 * s3.take("abscoeff")), "abscoeff", rtol=0.5e-2, plot=True)
 
 
+def test_vaex_and_pandas_dataframe_fetch_databank():
```
erwanp, line 225: Same as above, please write :  
- what are you testing 
- what results do you expect

radis/test/lbl/test_factory.py
```diff
@@ -725,6 +725,57 @@ def test_diluents_molefraction(verbose=True, plot=False, *args, **kwargs):
...
+    from radis import calc_spectrum
+    from radis.spectrum.utils import CONVOLUTED_QUANTITIES
+
+    databanks = ["hitran", "hitemp", "exomol", "geisa"]
```
erwanp, line 732: why are you testing for multiple databases ?

erwanp (changes requested):

1someshverma:
I will add the examples and resolve other issues

radis/api/dbmanager.py
```diff
@@ -516,7 +516,7 @@ def add_column(self, df, key, value):
 
             if isinstance(df, vaex.dataframe.DataFrameLocal):
```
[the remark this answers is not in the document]
1someshverma: ok,  I will change this.

radis/test/lbl/test_loader.py
1someshverma: I will  add the comments

radis/lbl/base.py
1someshverma: It is not necessary, but it improves time performance for vaex. But we can consider adding mode as saving memory or saving CPU time.

radis/lbl/loader.py
1someshverma: Because dataframe returned by sort_values doesn't have attrs, so I have to add it again.

radis/lbl/loader.py
```diff
@@ -1171,11 +1173,14 @@ def fetch_databank(
                     verbose=self.verbose,
                     return_local_path=True,
                     engine=memory_mapping_engine,
+                    output=output,
                     parallel=parallel,
                     extra_params=extra_params,
                 )
                 self.params.dbpath = ",".join(local_paths)
 
+                df.attrs = {}
```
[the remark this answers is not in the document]
1someshverma: This is because while pandas have df.attrs for vaex we need to declare explicitly. df.attrs is a dictionary, otherwise it raises an error.

radis/api/hdf5.py
```diff
@@ -367,7 +367,11 @@ def load(
             raise NotImplementedError(output)
 
         if output == "vaex":
-            df = df.extract()  # return DataFrame containing only the filtered rows
+            # df = df.extract()  # Not required # return DataFrame containing only the filtered rows
+            if columns:  # load only these columns (if they exist)
+                columns = [c for c in columns if c in df.columns]
```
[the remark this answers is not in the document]
erwanp: @1someshverma just so you don't forget

erwanp:
You can also add the script that allowed you to compute the performance comparison above ? (include which conditions, what database, etc.. )

radis/api/dbmanager.py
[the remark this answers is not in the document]
erwanp: can you push your latest changes when you've done them ?

erwanp:
Hello @1someshverma , don't forget to answer my comments above. Resolve conversations once you think we're all good

1someshverma:
I will push the latest code by the end of this week.

minouHub:
> Please push as soon as you can, as long as it runs. A pull request is not a final version, it is a draft, and is a great tool to collaborate/suggest. 

Indeed! Do not hesitate to push your latest PR even if all you to-do list is not complete. We could help you this way

radis/api/hdf5.py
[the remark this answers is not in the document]
1someshverma: When output and engine are both vaex, fetch_databank() was returning all the columns. So, I added this line. 
Can you specify of which file line ```1329``` I need to check.

radis/lbl/loader.py
1someshverma: I think that it will cost more time if we store "iso" as a column instead of an attribute.

environment.yml
```diff
@@ -25,6 +25,7 @@ dependencies:
 - termcolor  # terminal colors
 - specutils
 - vaex>=4.9.2   # load HDF5 files  (version needed to fix https://github.com/radis/radis/issues/486). #TODO : install only required sub-packages
+- pydantic=1.*
```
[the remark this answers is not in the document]
minouHub: When answering this, put a comment next to the import line (like for termcolor and vaex above) so that we know why you ask for this package

radis/lbl/broadening.py
```diff
@@ -2391,7 +2452,13 @@ def _broaden_lines(self, df):
                             )
                         )
 
-                    for i, (_, dg) in enumerate(df.groupby(arange(len(df)) % N)):
+                    if self.dataframe_type == "pandas":
+                        df = df.groupby(arange(len(df)) % N)
+                    elif self.dataframe_type == "vaex":
+                        df["idx"] = np.arange(len(df)) % N
+                        df = df.groupby(df.idx)
```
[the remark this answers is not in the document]
minouHub: And for vaex:
```python
# Create a new virtual column for group index
df['group_index'] = vaex.vrange(0, len(df))
df['group_index'] = df['group_index'] // N

# Group the DataFrame by the 'group_index' column
grouped = df.groupby('group_index')
```

minouHub (commented):
A few comments and suggested commits to consider :)

minouHub:
A few things to sort but this PR is close to being merged! Here is a list of things to not forget, feel free to tick them when done:

- [ ] Resolve/answer all comments of Erwan and Nicolas
- [ ] Add the script that allowed you to compute the performance comparison (see above Erwan comment). I suggest a file `vaex-vs-pandas_performances.py` in the `radis/test` folder. Do not add test in the filename so that Travis does not turn it on at each push. It's rather for us, to easily keep track fo the difference with time if needed.

radis/test/lbl/test_factory.py
1someshverma: To ensure that some commit has not changed vaex and pandas spectrum for some databanks and spectrum is same.

radis/api/hdf5.py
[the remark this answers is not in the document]
1someshverma: for vaex engine=output so that line ```329``` is not executed for it .

```
        if engine == output:
            df = df
        elif engine == "vaex":
            # in vaex, column selection has to happen now
            if columns:  # load only these columns (if they exist)
                columns = [c for c in columns if c in df.columns]
            if output == "pandas":
                df_pandas = df.to_pandas_df(column_names=columns)
                df.close()
                df = df_pandas
```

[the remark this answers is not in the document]
minouHub: Alright. Then I would simply add a `if engine = "vaex"` at the start. This would make the code just easier to read.

examples/calculate_spectrum_using_vaex.py
```diff
@@ -6,7 +6,7 @@
 Calculate spectrum using Vaex
 ===============================
 
-Calculating spectrum using pandas is memory intensive , that is it loads data into memory for calculating spectrum .For larger number of lines it restricts the 
+Calculating spectrum using pandas is memory intensive , that is it loads data into memory for calculating spectrum .For larger number of lines it restricts the
 number of lines that can be calculated on a device.
 When number of liens is more Vaex is recommended as engine.
 VAEX is a similar table management tool than PANDAS but uses memory mapping, a zero memory copy policy, and lazy computations for the best performance.
```
minouHub:
```suggestion
Calculating a spectrum using pandas can be memory-intensive, as it loads data into memory for the computation. This can restrict the number of lines that can be processed on a device, especially for larger datasets. In such cases, Vaex is recommended as the engine. Vaex is a table management tool similar to PANDAS, but it utilizes memory mapping, a zero-memory-copy policy, and lazy computations for optimal performance.
```

radis/lbl/base.py
```diff
@@ -3737,6 +3809,8 @@ def _reinitialize(self):
                             + "option, if you don't need to reuse this factory to calculate new spectra",
                             "MemoryUsageWarning",
                         )
+                elif self.dataframe_type == "vaex":
+                    print("Using Vaex to save memory")
```
minouHub, line 3813: add a `if verbose:` to avoid uncontrolled print

radis/test/lbl/test_factory.py
```diff
@@ -727,62 +730,66 @@ def test_vaex_and_pandas_spectrum():
...
-            mole_fraction=0.1,
-            wstep="auto",
-            path_length=1,  # cm
-            databank="hitemp",  # or 'hitemp', 'geisa', 'exomol'
-            optimization=None,
-            engine="vaex",
-            verbose=3,
-            return_factory=True,
-        )
+    from radis import calc_spectrum
```
minouHub, line 738: Some conditions are repeated. To improve readability and conciseness, you can store the redundant conditions with
```
conditions = {'molecule': 'CO', 'mole_fraction':0.1, 'path_length':1} #etc. 
s1 = calc_spectrum(**conditions, Trot=700, Tvib=710, Tgas=700, engine="vaex")
s2 = calc_spectrum(**conditions, Trot=700, Tvib=710, Tgas=700, engine="pandas")
```

radis/test/lbl/test_factory.py
```diff
@@ -727,62 +730,66 @@ def test_vaex_and_pandas_spectrum():
...
 
-        assert np.allclose(s.get("radiance"), s1.get("radiance"), equal_nan=True)
-        assert np.allclose(s.get("absorbance"), s1.get("absorbance"), equal_nan=True)
-        assert np.allclose(
-            s.get("transmittance"), s1.get("transmittance"), equal_nan=True
-        )
+    # Comparing different quantities
+    assert np.allclose(s.get("radiance"), s1.get("radiance"), equal_nan=True)
+    assert np.allclose(s.get("absorbance"), s1.get("absorbance"), equal_nan=True)
+    assert np.allclose(s.get("transmittance"), s1.get("transmittance"), equal_nan=True)
```
minouHub: no need to apply a slit, and checking only one quantity such as "absorbance" is enough :)

minouHub (commented):
A few minor comments

examples/calculate_spectrum_using_vaex.py
```diff
@@ -0,0 +1,51 @@
...
+    isotope="1,2,3",
+    pressure=1.01325,  # bar
+    Tgas=700,  # K
+    mole_fraction=0.1,
+    path_length=1,  # cm
+    databank="hitran",  # or 'hitemp', 'geisa', 'exomol'
+    engine="pandas",
+)
+s.apply_slit(0.5, "nm")  # simulate an experimental slit
+s.plot("radiance")
```
erwanp, line 51: can you easily show the memory usage in both cases n this example ?

examples/calculate_spectrum_using_vaex.py
```diff
@@ -0,0 +1,51 @@
...
+
+===============================
+Calculate spectrum using Vaex
+===============================
+
+Calculating a spectrum using pandas can be memory-intensive, as it loads data into memory
+for the computation. This can restrict the number of lines that can be processed on a
+device, especially for larger datasets. In such cases, Vaex is recommended as the engine.
+Vaex is a table management tool similar to PANDAS, but it utilizes memory mapping, a zero-
+memory-copy policy, and lazy computations for optimal performance.
```
erwanp, line 13:
```suggestion
Vaex is a table management tool similar to PANDAS, but it utilizes memory mapping, a zero-
memory-copy policy, and lazy computations for optimal performance. 

This means what whatever your RAM and whatever the size of the dataset, **you won't run out of memory**.
```

radis/test/lbl/test_factory.py
erwanp: add the comments in docstrings under the function, not above. Same for all other test functions

radis/test/lbl/test_factory.py
erwanp: Let me challenge this. I did not think the changes your made were dependant on the database, are they ?

By doing it you're increasing downloads & computation times for the test suite, which is both environmentally problematic and slightly annoying for the users => we want tests to be as minimal and as fast as possible.
However, we want even more the test suite to be robust and trustworthy.

Therefore, if there is **one** good reason to test Vaex vs Pandas for different databases (i.e. that Vaex & Pandas might be equal for HITRAN, but different for HITEMP), let's keep it. But from my point of view your work is database-indepednant. 
So just use one database, and make sure the database is already used in other tests so we do not have to download one more database. 


Tagging @minouHub too ; I think it's important to maintain a clean & efficient test suite

radis/test/lbl/test_factory.py
```diff
@@ -644,6 +644,140 @@ def test_all_spectrum_using_wstep_auto(verbose=True, plot=False, *args, **kwargs
...
+            Tgas=700,  # K
+            # mole_fraction={'CO': 0.5, 'CO2': 0.3},
+            mole_fraction=0.1,
+            path_length=1,  # cm
+            databank=databank,  # or 'hitemp', 'geisa', 'exomol'
+            # use_cached=False,
+            engine="vaex",
+            return_factory=True,
+        )
+        s.apply_slit(0.5, "nm")  # simulate an experimental slit
```
erwanp, line 673: rename `s` as `s_vaex`, `s1` as `s1_pandas`, it's clearer

erwanp: also you don't need to apply_slit()

radis/test/lbl/test_factory.py
```diff
@@ -644,6 +644,140 @@ def test_all_spectrum_using_wstep_auto(verbose=True, plot=False, *args, **kwargs
...
+
+        for spec_quantity in CONVOLUTED_QUANTITIES:
+            assert np.allclose(
+                s.get(spec_quantity)[1], s1.get(spec_quantity)[1], equal_nan=True
+            )
+
+        for column in factory_s1.df1.columns:
+            assert np.all(factory_s.df1[column].to_numpy() == factory_s1.df1[column])
+
+    #%% Additional test with other options (cutoff, ...)
```
erwanp, line 700: also mention this in the function docstrings.
e.g "Here are the things we ensure in this test : 
- that spectra calculated with Vaex & Pandas are the same 
- that spectra calculated with Vaex & Pandas using Database cutoff are the same 
- etc.
"

radis/test/lbl/test_factory.py
```diff
@@ -644,6 +644,140 @@ def test_all_spectrum_using_wstep_auto(verbose=True, plot=False, *args, **kwargs
...
+    from radis import SpectrumFactory
+
+    sf_vaex = SpectrumFactory(
+        wavelength_min=4200,
+        wavelength_max=4500,
+        cutoff=1e-23,
+        molecule="CO2",
+        dataframe_type="vaex",
```
erwanp, line 708: do you really need this new parameter "dataframe_type" ?  
Can't you get all you need later from the fetch_databank(output="vaex") value ?

radis/test/lbl/test_factory.py
```diff
@@ -644,6 +644,140 @@ def test_all_spectrum_using_wstep_auto(verbose=True, plot=False, *args, **kwargs
...
+            mole_fraction=0.1,
+            path_length=1,  # cm
+            databank=databank,  # or 'hitemp', 'geisa', 'exomol'
+            # use_cached=False,
+            engine="pandas",
+            return_factory=True,
+        )
+        s1.apply_slit(0.5, "nm")  # simulate an experimental slit
+
+        for spec_quantity in CONVOLUTED_QUANTITIES:
```
erwanp, line 692: s1.compare_with(s)  should work here ; doesn't it ?

radis/test/lbl/test_factory.py
```diff
@@ -644,6 +644,140 @@ def test_all_spectrum_using_wstep_auto(verbose=True, plot=False, *args, **kwargs
...
+
+    for spec_quantity in CONVOLUTED_QUANTITIES:
+        assert np.allclose(
+            s_vaex.get(spec_quantity)[1], s_pd.get(spec_quantity)[1], equal_nan=True
+        )
+
+
+#%%
+# Compares the Spectrum calculated under non-equilibrium conditions . Comparison is made between CONVOLUTED_QUANTITIES and dataframe df .
+def test_vaex_and_pandas_spectrum_noneq():
```
erwanp, line 736: same docstrings remark as above

radis/test/lbl/test_loader.py
```diff
@@ -222,13 +222,179 @@ def test_custom_abundance(verbose=True, plot=False, *args, **kwargs):
...
+            if_same = if_same * np.all(df_vaex[column].to_numpy() == df_pandas[column])
+
+        return if_same
+
+    # HITRAN
+    # Comparing Vaex dataframe and Pandas dataframe fetched from HITRAN
+
+    # Fetching in vaex dataframe format
+    sf.dataframe_type = "vaex"
+    sf.fetch_databank(
```
erwanp, line 253: do you need a SpectrumFactory, why not use fetch_databank() directly ?

radis/test/lbl/test_loader.py
```diff
@@ -222,13 +222,179 @@ def test_custom_abundance(verbose=True, plot=False, *args, **kwargs):
...
+        "hitran", memory_mapping_engine="vaex", output="vaex", load_columns="all"
+    )
+    df1 = sf.df0
```
erwanp, line 256: assert that it's the right output
i.e. something like 
```
assert isinstance(df1, vaex.DataFrame)
```

Same idea for df2 below

radis/test/lbl/test_loader.py
```diff
@@ -222,13 +222,179 @@ def test_custom_abundance(verbose=True, plot=False, *args, **kwargs):
...
+    sf.fetch_databank(
+        "hitran", memory_mapping_engine="vaex", output="pandas", load_columns="all"
+    )
+    df2 = sf.df0
+
+    # Comparing the dataframe fetched from HITEMP
+    assert compare_dataframe(df1, df2, df2.columns)
+
+    # HITEMP
+    # Comparing Vaex dataframe and Pandas dataframe fetched from HITEMP
```
erwanp, line 269: Same remark as above in calc_spectrum. 
I do not understand why you're testing all of this ; i.e. what new information are you getting ? From what I see in your PR, you did not change fetch_databank(). 

Having a Vaex/Pandas output in databank was done in 2021, and is already tested https://github.com/radis/radis/pull/306 (as far as I know. If they aren't, let's discuss !)

radis/test/vaex_vs_pandas_performance.py
```diff
@@ -0,0 +1,118 @@
+import time
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+from radis import calc_spectrum
+from radis.misc.progress_bar import ProgressBar
+
+
+def compare_vaex_pandas_time():
```
erwanp, line 10: I dislike that we need HITEMP H2O ; it's a very large database. 
Ok for your own performance tests, and to plot on the PR; but remember that these tests will be executed on every single commit during the next 5 years ! 
Let's be minimal. Use a database already downloaded in the other tests. 


Also, there is no assert statement anywhere here. So it's probably more a Benchmark than a Test. 
If confirmed, then add it in Radis/benchmarks : https://github.com/radis/radis-benchmark , where you can use whatever database you want as it's not ran on every new commit

(as well as in this PR text )

radis/test/vaex_vs_pandas_performance.py
```diff
@@ -0,0 +1,118 @@
...
+    plt.plot(lines_list, timeC_list, "k--", label="pandas computation")
+    plt.plot(lines_list_va, time_list_va, "r", label="vaex total")
+    plt.plot(lines_list_va, timeC_list_va, "r--", label="vaex computation")
+    plt.ylabel("Time [s]")
+    plt.xlabel("Number of lines")
+    plt.legend()
+
+
+# Compare the memory performance of Pandas and Vaex
+def compare_pandas_vs_vaex_memory():
```
erwanp, line 67: same as above

erwanp (changes requested):
Few changes regarding the CI : seems we're overkilling the test suite here, we need more minimal tests !

radis/lbl/calc.py
```diff
@@ -66,6 +66,7 @@ def calc_spectrum(
     export_lines=False,
     verbose=True,
     return_factory=False,
+    engine="pandas",
```
erwanp, line 69: Adding a new parameter makes the library more difficult to understand & use for new users; ideally we'd like all the input parameters of calc_spectrum() to be only physical parameters (pressure, temperature, species, etc.)  but not computational parameters. 

One alternative I'm trying to push is to store the "engine" key in `radis.config` , which itself is read from the radis.json file. Each user can modify their own radis.json file. The default one is copied from [radis/default_radis.json](https://github.com/radis/radis/blob/develop/radis/default_radis.json) . All of this is implemented and working quite well. 

Have a look at how `MEMORY_MAPPING_ENGINE` is implemented there and in the code. You could do a similar thing for a new key `DATAFRAME_ENGINE`, for instance (default "pandas" then "auto" or "vaex" if it works well !) 

Users can edit it either in radis.json ; or in their scripts : 
```python
import radis
radis.config["DATAFRAME_ENGINE"] = "vaex"
```


Also tagging @minouHub for reference

radis/lbl/loader.py
erwanp: sure about that ?

radis/phys/convert.py
```diff
@@ -345,12 +357,20 @@ def _assertK(E):
 
 
 def _assertcm(E):
-    if np.sum(np.abs(E)) != 0:  # check E != 0 for both floats and arrays
-        try:
-            m = _magn(E)
-            assert ((1 <= m) & (m <= 5)).all()
-        except AssertionError:
-            print(("Warning. Input values may not be in cm-1", E, "cm-1?"))
+    if isinstance(E, vaex.expression.Expression):
```
erwanp, line 360: when is _assertcm used in the code ? 

I'd like it better if we didn't touch _assertcm (which expects an array, not necessarily a Pandas column), but rather edit the function call

erwanp (changes requested):
(new comments)

radis/lbl/calc.py
1someshverma: I think as the default value of engine is pandas, everything thing should work fine for new users and all the examples given in docs can be tested without bothering about engine parameter of calc_spectrum().Just by supplying physical conditions , user can calculate spectrum easily.
Also, updating the radis.json every time user wants to switch from calculating spectrum with a low number of lines to spectrum having a high number of lines is not that flexible.

1someshverma: If you want changes i can make those.

radis/misc/arrays.py
```diff
@@ -345,13 +347,17 @@ def evenly_distributed_fast(w, rtolerance=1e-5):
     return np.isclose(w[-1] - w[-2], w[1] - w[0], rtol=rtolerance)
 
 
-def anynan(a):
+def anynan(a, dataframe_type="pandas"):
```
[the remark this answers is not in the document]
1someshverma: Ok, I will create a new function for vaex.

radis/test/lbl/test_factory.py
1someshverma: s1.compare_with works when data frames are of the same type, here data frames are of different types.

radis/test/vaex_vs_pandas_performance.py
1someshverma: These are not executed in every commit here is an example from last commit https://app.travis-ci.com/github/radis/radis/jobs/607697939. These functions were not executed in that as there name doesn't start or end with test so they are not detected by pytest and hence not executed.

The idea to use HITEMP CO2 was to have more clear idea about the performance of vaex for larger databases.

radis/lbl/calc.py
erwanp: Our target users aren't spectroscopy experts; and shouldn't even have to worry about which mode to use. Less parameters = less trouble for them, so I'd like not to have a new parameter unless there is a strong argument against it. There are already too many parameters in calc_spectrum (and we'll have to remove some)  [just have a look at [calc_spectrum](https://radis.readthedocs.io/en/latest/source/radis.lbl.calc.html#radis.lbl.calc.calc_spectrum) docs, it's frightening for a 1st time user ! :) )  

What you're suggesting (switch from one to another) is interesting and could (should !) be done under-the-hood, without the user noticing : 
- there could be an "auto" mode that automatically switch if we evaluate that Vaex might be faster when loading the database, 

Do not implement this now, as the goal is to merge the Vaex calculations (it won't be the default mode yet); but keep in mind that this is where we're heading to so that the underlying architecture is ready for this "auto" mode. 

So to wrap is up

1. for the time being, use `DATAFRAME_ENGINE = "pandas"` as default, but `DATAFRAME_ENGINE = "vaex"` should work 
2. we'll publish a new Radis release and ask people in the Slack to use DATAFRAME_ENGINE = "vaex" and see if it works for them, gather some user feedbacks
3. if it works well we can implement `DATAFRAME_ENGINE = "auto"`

radis/test/vaex_vs_pandas_performance.py
erwanp: You're right, I did not notice function names don't start with `test*` so we're good. You can resolve my comment. Maybe start with "benchmark_*"  and mention explicitely in the docstrings that they're long tests

```
.. warnings 
    This function will download the full H2O HITEMP database which requires > 10 (?) GB and many minutes if not hours of parsing. It is not executed by default in the test suite (name doesn't start with test_*)
```

radis/lbl/calc.py
1someshverma: I will make the changes.

radis/test/lbl/test_loader.py
1someshverma: fetch_databank() is a method of DatabankLoader class so it can't be used directly

radis/test/lbl/test_factory.py
```diff
@@ -746,7 +744,7 @@ def test_vaex_and_pandas_spectrum_noneq():
         "mole_fraction": 0.1,
         "wstep": "auto",
         "path_length": 1,
-        "databank": "hitemp",
+        "databank": "hitran",
```
erwanp: we actually already download HITEMP CO for other tests, so you can keep this one !

erwanp (approved):
looks good to me, minor change

radis/lbl/broadening.py
```diff
@@ -3021,9 +3089,15 @@ def project_lines_on_grid(df, wavenumber, wstep):
...
 
-    shiftwav = df.shiftwav.values  # cm-1  ,   size N (number of lines)
-    S = df.S.values  # cm/#  ~   cm-1/(#.cm-2)  ,   size N
-    wv = df.hwhm_voigt.values * 2  # HWHM > FWHM
+    if dataframe_type == "pandas":
+        shiftwav = df.shiftwav.values  # cm-1  ,   size N (number of lines)
+        S = df.S.values  # cm/#  ~   cm-1/(#.cm-2)  ,   size N
+        wv = df.hwhm_voigt.values * 2  # HWHM > FWHM
+    elif dataframe_type == "vaex":
+        # TODO : check memory usage & if there is no other implementation without numpy conversion
```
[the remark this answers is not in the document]
minouHub: From a quick look, the rest of the function would have to be refactored whether the dataframe is pandas or vaex. Please keep the comment open

minouHub:
Please re-apply the changes of 5c2c1240 on vaex_vs_pandas_performances.py.

1someshverma:
Yes, I also noticed it . I will add this in next commit.

radis/test/benchmark/vaex_vs_pandas_performance.py
```diff
@@ -71,30 +64,52 @@ def compare_vaex_pandas_time():
...
-        pressure=1.01325,  # bar
-        Tgas=1000,  # K
-        mole_fraction=0.1,
-        wstep="auto",
-        path_length=1,  # cm
-        databank="hitemp",  # or 'hitemp', 'geisa', 'exomol'
-        engine="vaex",
-    )
+    for engine in ["pandas", "vaex"]:
+        config["DATAFRAME_ENGINE"] = engine
```
erwanp, line 77: just as good practice, reset the final state of config["DATAFRAME_ENGINE"] at the end of the test

erwanp (approved):

erwanp (approved):
All good ! @1someshverma @minouHub according to me we can merge !

minouHub:
For point 2., the output of the vaex_vs_pandas_performance.py should be enough @1someshverma

1someshverma:
TIme Performance Graph
It was obtained when i tested code on Kaggle 
Screenshot at 2023-07-12 22-08-43



It was obtained by @minouHub 
image(1)

1someshverma:
For 1 point, I tested this code and it is working fine 
```
from radis import config
from radis.spectrum import plot_diff
config["DATAFRAME_ENGINE"] = "vaex"
s_vaex = calc_spectrum(1900, 2300,         # cm-1
                  molecule='CO',
                  isotope='1,2,3',
                  pressure=1.01325,   # bar
                  Tgas=700,           # K
                  mole_fraction=0.1,
                  path_length=1,      # cm
                  )
config["DATAFRAME_ENGINE"] = "pandas"
s_pandas = calc_spectrum(1900, 2300,         # cm-1
                  molecule='CO',
                  isotope='1,2,3',
                  pressure=1.01325,   # bar
                  Tgas=700,           # K
                  mole_fraction=0.1,
                  path_length=1,      # cm
                  )

plot_diff(s_vaex, s_pandas)
```

diff

1someshverma:
Thanks sir

erwanp:
Your time performances numbers (# lines) do not correspond to the Memory performance of the PR description. Can you update it, and show the code that generated the graph ?

1someshverma:
I will update it

1someshverma:
The memory graph include lines before any cutoff is applied while time graph which I obtained later includes number of lines after cutoff is applied . 
I will share the memory vs line , lines after cutoff is applied  soon.

1someshverma:
This code can be used to obtain a comparison between memory use 
```
# Compare the memory performance of Pandas and Vaex
def compare_pandas_vs_vaex_memory():
    """
    Compare memory usage of `engine="vaex"` and `engine="pandas"` in calc_spectrum.
    Expected behavior is "vaex" using much less memory. This function takes tens of seconds to run.
    Returns
    -------
    None.
    """

    import tracemalloc

    initial_engine = config[
        "DATAFRAME_ENGINE"
    ]  # To make sure dataframe engine not changed after running this test
    for engine in ["pandas", "vaex"]:
        config["DATAFRAME_ENGINE"] = engine
        tracemalloc.start()
        s = calc_spectrum(
            1000,
            1500,  # cm-1
            molecule="H2O",
            isotope="1,2,3",
            pressure=1.01325,  # bar
            Tgas=1000,  # K
            mole_fraction=0.1,
            wstep="auto",
            databank="hitemp",  # or 'hitemp', 'geisa', 'exomol'
            verbose=0,
        )
        snapshot = tracemalloc.take_snapshot()
        memory = tracemalloc.get_traced_memory()
        tracemalloc.stop()

        # Some raw outputs
        print("\n******** Engine = {} ***********".format(engine))
        print(
            "Peak, current = {:.1e}, {:.1e} for {:} lines calculated".format(
                *memory, s.conditions["lines_calculated"]
            )
        )

        # More sophisticated
        print("*** List of biggest objects ***")
        top_stats = snapshot.statistics("lineno")
        for rank, stat in enumerate(top_stats[:3]):
            print("#{}".format(rank + 1))
            print(stat)

        # Clear for next engine in the loop
        tracemalloc.clear_traces()

    config["DATAFRAME_ENGINE"] = initial_engine
```

erwanp:
Add the CPU performance too. You can get it from

```
s.c["calculation_time"]
```

1someshverma:
Memory Use graph
vaexMemory

Time Performance Graph
vaexTime

But for time performance, Graph Obtained by Nicolas give complete view , I have not been able to run it for 1000 - 4000 wave range due to RAM size of my laptop but the graph by Nicolas give more clear picture of the time performance of vaex and pandas.

image(1)

Code used 
```
from radis import  config
import matplotlib.pyplot as plt
import tracemalloc

memory_use_list_vaex = []
memory_use_list_pandas = []
computation_time_list_vaex = []
computation_time_list_pandas = []
number_lines_list = []

for engine in ["pandas","vaex"]:
    for wmax in [1050,1100,1300,1500,1800,2100]:
        tracemalloc.start()
        config["DATAFRAME_ENGINE"] = engine
        s = calc_spectrum(1000, wmax,         # cm-1
                          molecule='H2O',
                          isotope='1,2,3',
                          pressure=1.01325,   # bar
                          Tgas=700,           # K
                          mole_fraction=0.1,
                          path_length=1,      # cm
                          databank='hitemp',  # or 'hitemp', 'geisa', 'exomol'
                          )

        if engine == "vaex":
            memory_use_list_vaex.append(tracemalloc.get_traced_memory()[1])
            computation_time_list_vaex.append(tracemalloc.get_traced_memory()[1])
            number_lines_list.append(s.conditions["lines_calculated"])
        elif engine == "pandas":
            memory_use_list_pandas.append(tracemalloc.get_traced_memory()[1])
            computation_time_list_pandas.append(s.conditions["lines_calculated"])

        tracemalloc.stop()
        tracemalloc.clear_traces()

plt.figure()
plt.plot(number_lines_list, computation_time_list_pandas, "k", label="pandas computation time")
plt.plot(number_lines_list, computation_time_list_vaex, "r", label="vaex computation use")
plt.ylabel("Time [s]")
plt.xlabel("Number of lines")
plt.legend()
plt.show()

plt.figure()
plt.plot(number_lines_list, memory_use_list_pandas,"k",label="pandas memory use")
plt.plot(number_lines_list, memory_use_list_vaex,"r",label="vaex memory use")
plt.ylabel("Memory (in bytes)")
plt.xlabel("Number of lines")
plt.legend()
plt.show()
```

erwanp:
Thanks for sharing the graph ! 
Actually, you mentioned that in your case you couldn't compute with more lines due to limited RAM ... could you do it with Vaex ?

1someshverma:
Yes, it can be done for the vaex .

erwanp:
One last comment : using "wstep='auto'" is good in general ; but for the current case it means you're not only changing the number of lines but also the grid resolution therefore the wavenumber grid size. 
It's better if, for the sake of the comparison, you use a fixed wstep.  You're at atmospheric pressure so `wstep=0.01` should work

1someshverma:
Memory Performance Graph
image
For Pandas i am able to calculate 20K lines(in wav range 500-1200) but for Vaex I have calculated up to 366000(in wav range 500 - 30000) and still memory use is half of the maximum memory used by pandas.


Time Performance 
image

Code Used 
```

from radis import calc_spectrum, config
import matplotlib.pyplot as plt
import tracemalloc

memory_use_list_vaex = []
memory_use_list_pandas = []
computation_time_list_vaex = []
computation_time_list_pandas = []
number_lines_list_pandas = []
number_lines_list_vaex = []

config["DATAFRAME_ENGINE"] = "pandas"
for wmax in [600,700,800,900,1000,1200]:
    tracemalloc.start()
    s = calc_spectrum(500,wmax,         # cm-1
                      molecule='H2O',
                      isotope='1,2,3,4,5,6',
                      pressure=1.01325,   # bar
                      Tgas=700,           # K
                      mole_fraction=0.1,
                      wstep=0.01,
                      path_length=1,      # cm
                      databank='hitemp',  # or 'hitemp', 'geisa', 'exomol'
                      )
    memory_use_list_pandas.append(tracemalloc.get_traced_memory()[1])
    computation_time_list_pandas.append(s.c["calculation_time"])
    number_lines_list_pandas.append(s.conditions["lines_calculated"])
    tracemalloc.stop()
    tracemalloc.clear_traces()

config["DATAFRAME_ENGINE"] = "vaex"
for wmax in [600,700,800,900,1000,1200, 1500,2000,2500,3000,3500,4000,4500,5000,6000,7000,8000,9000,10000,11000,30000]:
    tracemalloc.start()
    s = calc_spectrum(500,wmax,         # cm-1
                      molecule='H2O',
                      isotope='1,2,3,4,5,6',
                      pressure=1.01325,   # bar
                      Tgas=700,           # K
                      mole_fraction=0.1,
                      wstep=0.01,
                      path_length=1,      # cm
                      databank='hitemp',  # or 'hitemp', 'geisa', 'exomol'
                      )
    memory_use_list_vaex.append(tracemalloc.get_traced_memory()[1])
    computation_time_list_vaex.append(s.c["calculation_time"])
    number_lines_list_vaex.append(s.conditions["lines_calculated"])
    tracemalloc.stop()
    tracemalloc.clear_traces()


plt.figure()
plt.plot(number_lines_list_pandas, computation_time_list_pandas, "k", label="pandas computation time")
plt.plot(number_lines_list_vaex, computation_time_list_vaex, "r", label="vaex computation time")
plt.ylabel("Time [s]")
plt.xlabel("Number of lines")
plt.legend()
plt.show()

plt.figure()
plt.plot(number_lines_list_pandas, memory_use_list_pandas,"k",label="pandas memory use")
plt.plot(number_lines_list_vaex, memory_use_list_vaex,"r",label="vaex memory use")
plt.ylabel("Memory (in bytes)")
plt.xlabel("Number of lines")
plt.legend()
plt.show()
```

erwanp:
Great, these are two very good graph. Seeing that the final ```wmax``` is 30,000 you're actually able to compute the full-range H2O database on a 8 GB RAM laptop . 

Just out of curiosity, do you manage to it with no-cutoff (i.e. all lines will be used even the ones that are automatically discarded at 700 K )  ? 

```python
  s = calc_spectrum(500,30000,         # cm-1    👈
                    molecule='H2O',
                    isotope='1,2,3,4,5,6',
                    pressure=1.01325,   # bar
                    cutoff=0,            # 👈
                    Tgas=700,           # K
                    mole_fraction=0.1,
                    wstep=0.01,
                    path_length=1,      # cm
                    databank='hitemp',  # or 'hitemp', 'geisa', 'exomol'
                    )
```

and what's the number of lines / calculation time then ? 
(you're supposed to find 114 million lines for HITEMP H2O : https://hitran.org/hitemp/ )

1someshverma:
cutoff is default which is 1e-27

1someshverma:
I am not able to calculate with cutoff 0 for the code that you shared.

erwanp:
Does it fail with Memory Error ? If so you can post-it as a new Issue. 
Might be the next GSOC project ! ;)  
Our long-term target is to compute anything anywhere.

1someshverma:
It doesn't seem to give any error , it just crashes.

erwanp:
Likely because of the remaining [to_numpy() call in this part](https://github.com/radis/radis/blob/f60094a6356b89eae2221063472668a4d3a3f03b/radis/lbl/broadening.py#L2019)  of the code

Issue #585, opened 2023-07-27
Bug Report: Fetch exomol


minouHub:
### 🐛 Describe the bug

The following code produces an error:
```python
from radis.io.exomol import fetch_exomol

df_exomol = fetch_exomol("CO2", output="vaex", 
                         verbose = 0, 
                         load_wavenum_min=2000, load_wavenum_max=2500,
                         )
```
output:
```
Background atmosphere:  Air
Reading C:\Users\Nicolas Minesi\.radisdb\exomol\CO2\12C-16O2\UCL-4000\12C-16O2__UCL-4000__02000-03000.trans.bz2
.broad is used.
Broadening code level= a0
Traceback (most recent call last):

  File C:\Anaconda\lib\site-packages\spyder_kernels\py3compat.py:356 in compat_exec
    exec(code, globals, locals)

  File c:\users\nicolas minesi\python\radis\io\untitled8.py:10
    df_exomol = fetch_exomol("CO2", output="vaex",

  File c:\users\nicolas minesi\python\radis\io\exomol.py:264 in fetch_exomol
    mdb.set_broadening(df, output=output)

  File c:\users\nicolas minesi\python\radis\api\exomolapi.py:1315 in set_broadening
    self.compute_broadening(

  File c:\users\nicolas minesi\python\radis\api\exomolapi.py:1360 in compute_broadening
    jlower_max=np.max(jlower),

  File <__array_function__ internals>:200 in amax

  File C:\Anaconda\lib\site-packages\numpy\core\fromnumeric.py:2820 in amax
    return _wrapreduction(a, np.maximum, 'max', axis, None, out,

  File C:\Anaconda\lib\site-packages\numpy\core\fromnumeric.py:84 in _wrapreduction
    return reduction(axis=axis, out=out, **passkwargs)

TypeError: max() got an unexpected keyword argument 'axis'
```

### 💡 Possible solutions

Inspecting the file, I found several problems, mostly relating to the calculation of the broadening parameter (gamma and N) in the [compute_broadening](https://github.com/radis/radis/blob/0b87b6f9e0541fce9f6c9e8b7be30ffefbdd3c30/radis/api/exomolapi.py#L1326) function of `exomol.api`:

- [ ] First of all, the name of the functions is confusing. The actual pressure broadening FWHM is not calculated here. Only the broadening coefficients are calculated. 
- [ ] The handling of the exception is not well organized. A very large `try` `except` is employed. This should be avoided as much as possible.
[...]

PR #586, opened 2023-07-27
Fetch exomol

minouHub:
### Description

This pull request is to address issue #585.

Inspecting the file, I found several problems, mostly relating to the calculation of the broadening parameter (gamma and N) in the [compute_broadening](https://github.com/radis/radis/blob/0b87b6f9e0541fce9f6c9e8b7be30ffefbdd3c30/radis/api/exomolapi.py#L1326) function of exomol.api:

- [x] First of all, the name of the functions is confusing. The actual pressure broadening FWHM is not calculated here. Only the broadening coefficients are calculated. 
- [x] The handling of the exception is not well organized. A very large `try` `except` is employed. This should be avoided as much as possible.
- [ ] It is not possible to ask for another broadener than "air". 
- [x] `verbose=0` has no impact on the print. 
- [x] in the specific case of CO2, the above error is raised because the Dataframe is in `vaex` format (instead of Pandas historically)


radis/lbl/loader.py
```diff
@@ -931,6 +931,7 @@ def fetch_databank(
         load_columns="equilibrium",
         parallel=True,
         extra_params=None,
+        verbose=True,
```
erwanp, line 934: There is a SpectrumFactory.verbose parameter, no need to create a new parameter in the function

Just use 
```
if self.verbose
```
or 
```
verbose = self.verbose
if verbose ... 
```

(also, not related but always good to mention, we can have multiple verbose levels like 
```
if verbose >= 2: print("some extra details")
```

erwanp (changes requested):
(see my comment on using self.verbose rather than verbose)

radis/lbl/loader.py
minouHub: self.verbose ==> indeed, done

erwanp:
Failing with https://app.travis-ci.com/github/radis/radis/jobs/607057429#L2301  : remaining `sf.fetch_databank(**conditions, verbose=verbose)`   in `_calc_spectrum_one_molecule()`   @minouHub

minouHub:
The last feature (broadening with other molecules than air) is not critical and will be sorted in #602

erwanp (approved):

PR #587, opened 2023-07-30
fixes #529

### LINKED ISSUE #529: Bug Report: ValueError in _is_at_equilibrium() due to Tvib not being float
### 🐛 Describe the bug

I’m trying to write a fitting routine for CO2 using `lmfit` and stumbled upon a ValueError being raised in `calc.py`, `_is_at_equilibrium()`.

```error
~\AppData\Local\Programs\Python\Python39\lib\site-packages\radis\lbl\calc.py in _is_at_equilibrium()
    642     def _is_at_equilibrium():
    643         try:
--> 644             assert Tvib is None or Tvib == Tgas
    645             assert Trot is None or Trot == Tgas
    646             assert overpopulation is None

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

In this line `Tvib` - a tuple of floats - is getting compared to a float hence raising this error.

### 💡 Possible solutions

 I assume that this line should not have be raised to start with but for now I avoid this error by writing 

```python
assert Tvib is None or (min(Tvib) == Tgas and max(Tvib) == Tgas)

```

therefore, checking for equilibrium within `Tvib` and to `Trot.`

### 🎲 Radis version

0.13.1

### 💻 Operating system

Windows

erwanp:
fixes #529  reported by @MrAnimaniac   when checking Tvib1,Tvib2,Tvib3 <> Tgas

### DIFF
--- a/radis/lbl/calc.py
+++ b/radis/lbl/calc.py
@@ -20,7 +20,7 @@ Routine Listing
 from copy import deepcopy
 from os.path import exists
 
-# from radis.db.classes import M
+import numpy as np
 
 try:  # Proper import
     from .base import get_wavenumber_range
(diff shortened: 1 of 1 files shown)


erwanp:
(tests are OK except from an Exomol molecule mismatch, no need to fix here will be done in #589)

Issue #545, opened 2023-02-06
Docs are broken


erwanp:
### 🐛 Describe the bug

Docs are broken : https://readthedocs.org/projects/radis/builds/19302835/ 

Apparently due to missing test file : `FileNotFoundError: Test file `trimmed_1857_VoigtCO_Minesi.mat` does not exist`

```

Extension error:
Here is a summary of the problems encountered when running the examples

Unexpected failing examples:
/home/docs/checkouts/readthedocs.org/user_builds/radis/checkouts/develop/examples/plot_compare_CO_geisa_hitran.py failed leaving traceback:
Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/radis/checkouts/develop/examples/plot_compare_CO_geisa_hitran.py", line 37, in <module>
    s_geisa = calc_spectrum(**conditions, databank="geisa", name="GEISA's CO")
  File "/home/docs/checkouts/readthedocs.org/user_builds/radis/envs/develop/lib/python3.8/site-packages/radis/lbl/calc.py", line 534, in calc_spectrum
    generated_spectrum = _calc_spectrum_one_molecule(
  File "/home/docs/checkouts/readthedocs.org/user_builds/radis/envs/develop/lib/python3.8/site-packages/radis/lbl/calc.py", line 800, in _calc_spectrum_one_molecule
    sf.fetch_databank(**conditions)
  File "/home/docs/checkouts/readthedocs.org/user_builds/radis/envs/develop/lib/python3.8/site-packages/radis/lbl/loader.py", line 1364, in fetch_databank
    df, local_paths = fetch_geisa(
  File "/home/docs/checkouts/readthedocs.org/user_builds/radis/envs/develop/lib/python3.8/site-packages/radis/io/geisa.py", line 174, in fetch_geisa
    df = ldb.load(
TypeError: load() got an unexpected keyword argument 'isotope'


/home/docs/checkouts/readthedocs.org/user_builds/radis/checkouts/develop/examples/plot_newfitting_Tgas-molfrac.py failed leaving traceback:
Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/radis/checkouts/develop/examples/plot_newfitting_Tgas-molfrac.py", line 35, in <module>
    data = scipy.io.loadmat(getTestFile(data_file), simplify_cells=True)["CO_resu_Voigt"]
[...]

PR #477, opened 2022-05-22
Compute power based on vacuum wavelengths - Resolution proposal for bug #460

AllanHOlesenBW:
### Description

This pull request is to address the wavelength problem described in bug #460.

Fixes #460

PR #591, opened 2023-07-30
Fix docs and various little things

### LINKED ISSUE #460: Integrals based on wavelength differs from integrals based on wavenum by a factor of 1.00027
Original comment from @AllanHOlesenBW  

---

**Integrals based on wavelength differs from inte

> grals based on wavenum by a factor of 1.00027.**
> At first, I thought that this was just a numerical error in the integrals. But it seems very consistent. And the error is suspiciously close to the refraction index for air.
> 
> An example of this can be seen in bug #454 where @erwanp performs a check of integrals after conversions of emisscoeff, based on wavenum and wavelength:
> 
> > Also checking that the integral is properly conserved in both cases :
> > 
> > 0.001534377339562196
> > 0.0015347955335563383
> 
> These two numbers are actually different by a factor of 1.00027
> 
> A demonstration example:
> 
> ```
> s = calc_spectrum(1900, 2300, # cm-1
>         molecule='CO',
>         isotope='1,2,3',
>         pressure=1.01325, # bar
>         Tgas=700, # K
>         mole_fraction=0.1,
>         path_length=1, # cm
>         databank='hitran', # or use 'hitemp'
>     )
> 
> print('Wave unit:    ', s.get_waveunit())
> print('Radiance unit:', s.units['radiance'])
> 
> print('Testing different methods of integrating the radiance over the full spectrum.')
> print()
> print('First the specific built-in method, Spectrum.get_power()')
> print('This uses an internal rescaling of waveunit to nm to get a meaningful integral:')
> power1 = s.get_power(unit='W/m2/sr')
> print(power1)
> 
> print()
> print('Replicate this using the generic built-in integration method, also converting to wavelength:')
> power2 = s.get_integral("radiance_noslit", wunit="nm", Iunit="W/m2/sr/nm")
> print(power2)
> print('Ratio between this result and get_power()', power2/power1)
> 
> print()
> print('Do the same, but this time maintain the wavenum based units:')
> power3 = s.get_integral("radiance_noslit", wunit="cm-1", Iunit="W/m2/sr/cm-1")
> print(power3)
> print('Ratio between this result and get_power()', power3/power1)
> 
> print()
> print('Test again with wavelength base, but this time use nm_vac as wunit:')
> power4 = s.get_integral("radiance_noslit", wunit="nm_vac", Iunit="W/m2/sr/nm")
> print(power4)
> print('Ratio between this result and get_power()', power4/power1)
> ```
> 
> I would expect that if I specify wunit="nm", Iunit="W/m2/sr/nm" in an integral, then those nm are the same type of nanometers, either vacuum based or air based. 
> 
> And if I specify wunit="cm-1", Iunit="W/m2/sr/cm-1", I expect those to be the same type of wavenums, either vacuum based or air based.
> 
> Something clearly deviates  from this expectation since I get different results, but I am not sure which of the results are correct.

_Originally posted by @AllanHOlesenBW in https://github.com/radis/radis/issues/215#issuecomment-1112985821_

erwanp:
- fix GEISA and upload .mat test file. This should fix docs #545
- fix test after power calculation correction in #477 (to fix #460)
- refactor : rename lbl/test_spectrum to avoid confusion with spectrum/test_spectrum
- fix Travis-CI badge on README


erwanp:
@minouHub I'm merging, we'll see afterwards if docs are truely fixed, please have a look

minouHub:
Concerning the docs, they are still failing on the develop branch (a 7.30 am on the 31st of July), see [bug report](https://readthedocs.org/projects/radis/builds/21453774/)

PR #622, opened 2023-09-21
Install libmamba in the base environment before creating radis-env.

pmariotto:
### Description

This pull request is to update the installation procedure: we remind the user that libmamba must be installed in the `base` environment before creating the `radis-env` environment.

### DIFF
--- a/docs/dev/_install.rst
+++ b/docs/dev/_install.rst
@@ -34,6 +34,8 @@ isolated environment. We recommend to use the faster, local-optimization ``libma
 You can create a radis environment with all dependencies with::
 
     cd radis
+    conda deactivate
+    conda install -n base conda-libmamba-solver
     conda env create --file environment.yml --solver=libmamba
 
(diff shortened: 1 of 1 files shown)


erwanp:
Anaconda is actually deploying Mamba as its default solver this month (September) so I don't think we need to update the docs (it's a matter of weeks) 

https://conda.org/blog/2023-07-05-conda-libmamba-solver-rollout/

minouHub:
Libmamba has been installed by default in conda since Dec. 2022, see [here](https://github.com/conda/conda/blame/af87ce22a53b8b29d45b2dd21c4acb509a856df1/recipe/meta.yaml#L55). Just note that everyone does not update conda every year. Worst-case scenario, the user should understand that `libmamba` must installed if it is not in his/her current version of conda.

erwanp:
Conda 23.10 now uses Libmamba by default : https://github.com/conda/conda/releases/tag/23.10.0 
This PR is no longer required

PR #624, opened 2023-10-09
Add Vulkan driver API

dcmvdbekerom:
### Description

This pull request is to add Vulkan driver API. This adds support for all vendors (Nvidia/AMD/Intel/etc.) on all platforms (Windows/Linux/Mac).

For current PR:
- [x] Add vulkan_compute_lib.py for Vulkan backend
- [x] Add VkFFT for vulkan FFT's
- [x] Rewrite GPU code in Vulkan format
- [x] Add timing back in (possibly using timers from Vulkan API)
- [x] Fix timers (times are currently underreported)
- [x] Add linux support
- [x] Recompile vkFFT for older linux versions (hard!)
- [x] Add Vulkan files to manifest
- [x] Fix travis test issues
- [x] Fix memory leaks
- [x] Make iso uint32
- [x] Make database one big array
- [x] Fix VkFFT int32/int64 issue 
- [x] Make device_id keyword accessible
- [x] ~~Fix exomol broadening issue
       The GPU code expects a 'selfbr' and 'airbr' key in the df0 dataframe, but exomol dataframes may not contain the 'selfbr' key. Moreover, the GPU code expects the 'gp' key, i.e. the upper state degeneracy, which is also not present. How to deal with this?~~ Not a GPU issue, see https://github.com/radis/radis/issues/714

- [x] Dynamically grow LDM buffer:
       Allocate some extra space, e.g. 1.5x(N_L x N_G); if it's too small reallocate with 50% extra size. This is a nice compromise between minimal wasted (unused allocated) memory and minimal number of re-allocations. It also means we don't need to know maximum T or p in advance. It also better integrates with the legacy cuda code.
  - [x] Separate buffer memory from array shape
  - [x] Implement reshape method
    - [x] Refresh FFT's
    - [x] ~~Refresh compute shaders~~ The compute shaders don't have to be refreshed.
- [x] Dynamic scaling / remove max T and p / rewrite command buffer on the fly
- [x] Bind descriptor set only once (as opposed to with every pipeline)
- [x] remove debug printfs from vkfft dll
- [x] ~~Provide access to legacy cuda code~~ CUDA backend support was dropped; see https://github.com/radis/radis/issues/758
- [x] Update docs

 

---

The following TODO items, more or less in order of urgency/importance, will not become part of this PR anymore:
- Cut out pyvkfft intermediate layer and move all code to `vulkan_copmute_lib.py` or `VkFFT (c++)`
- Use dedicated memory flags for indirect/uniform/ssbo buffers
- Perform apply lines at the end of forward FFT in registers
- Check effect of strides on shader performance
- Use VkFFT for compilation of GPU code
- Get bindings from GLSL (as opposed to hardcoded in gpu.py)
- Add pocketfft binary for multithreaded CPU FFT's


dcmvdbekerom:
Vulkan is working! It is still quite a bit slower, ~100ms (Vulkan) vs ~4ms (Cuda) on a Nvidia card, but I'm quite confident this can be solved through a better understanding of Vulkan and how to run pipelines efficiently.

Still lot's of open points, but the basic functionality is there.

You can change the GPU that's used by changing the `deviceID` keyword in line 108 of `radis.gpu.gpu.py` (changing the device will obviously be possible through keywords to spectrumfactory in the future).

minouHub:
Here is a test I tried. 
```python

sf = SpectrumFactory(
    2150,
    2450,  # cm-1
    molecule="CO2",
    isotope="1,2,3",
    wstep=0.002,
)

sf.fetch_databank("hitemp")

T = 1500.0  # K
p = 1.0  # bar
x = 0.8
l = 0.2  # cm
w_slit = 0.5  # cm-1

# s_cpu = sf.eq_spectrum(
#     name="CPU",
#     Tgas=T,
#     pressure=p,
#     mole_fraction=x,
#     path_length=l,
# )
# s_cpu.apply_slit(w_slit, unit="cm-1")

s_gpu = sf.eq_spectrum_gpu(
    name="GPU",
    Tgas=T,
    pressure=p,
    mole_fraction=x,
    path_length=l,
    backend="gpu-cuda",
)
s_gpu.apply_slit(w_slit, unit="cm-1")
# plot_diff(s_cpu, s_gpu, var="emissivity", wunit="nm", method="diff")
```

```python
In ``C:\Users\Nicolas Minesi\.radisdb\hitemp`` keep only relevant input files:
CO2-02_02125-02250_HITEMP2010.hdf5
CO2-02_02250-02500_HITEMP2010.hdf5
Selected card (deviceID=0):
[X] 0: Intel(R) UHD Graphics
[ ] 1: NVIDIA RTX A1000 Laptop GPU

  File C:\Anaconda\envs\radis-vulkan2\lib\site-packages\spyder_kernels\py3compat.py:356 in compat_exec
    exec(code, globals, locals)

  File c:\users\nicolas minesi\python\examples\plot_gpu.py:49
    s_gpu = sf.eq_spectrum_gpu(

  File c:\users\nicolas minesi\python\radis\lbl\factory.py:1144 in eq_spectrum_gpu
    gpu_init(

  File c:\users\nicolas minesi\python\radis\gpu\gpu.py:214 in gpu_init
    app.schedule_shader(

  File c:\users\nicolas minesi\python\radis\gpu\vulkan\vulkan_compute_lib.py:84 in schedule_shader
    pipeline, pipelineLayout, computeShaderModule = self.createComputePipeline(

  File c:\users\nicolas minesi\python\radis\gpu\vulkan\vulkan_compute_lib.py:426 in createComputePipeline
    if len(pipelines) == 1:

TypeError: cdata of type 'struct VkPipeline_T *' has no len()
```

I really like the list of available GPU in the prompt btw. I tried with both GPU and it gave the same output.

dcmvdbekerom:
It looks like the driver wasn't able to load (one of) the shaders (this is what kernels are called in Vulkan). This can happen because Vulkan can use extensions that may or may not be present on your particular device. In the future we can test for this at runtime, but only once we know what the problem is. Which extensions are present depend both on the hardware capabilities as well as the Vulkan version.

My suspicion is that `EXT_scalar_block_layout` is the culprit. To test this, we first need to check your capabilities with the [Vulkan capability viewer](https://www.saschawillems.de/creations/vulkan-hardware-capability-viewer/)

Then can you report back what is the value of `Properties->Core 1.0->apiVersion`? And what is the value of `Features->Core 1.2->scalarBlockLayout`? (this should be True or False). Please check these values for both of your cards (change with "Select device")

minouHub:
Integrated GPU: Intel(R) UHD Graphics
- apiVersion: 1.3.250
- scalarBlockLayout: true

Integrated GPU: NVIDIA RTX A1000

- apiVersion: 1.3.242
- scalarBlockLayout: true

dcmvdbekerom:
This all looks good, not sure what's the problem then..

dcmvdbekerom:
Actually, could you try the most recent version of the PR?
It likely won't solve your issue, but it might point to what's wrong.
I'm hoping it will say `AttributeError: 'GPUApplication' object has no attribute 'applyLineshapes'`, in which case it's a simple path issue.

minouHub:
same error message after a pull and a `pip install -e .`

minouHub:
Working with `vulkan==1.1.99.1`

dcmvdbekerom:
Now also works on linux! Unfortunately not on Travis :(

erwanp:
@dcmvdbekerom aren't the test failing just because you aren't running on a Travis-GPU enabled instance ? 

It can be activated here: 

https://docs.travis-ci.com/user/reference/overview/#gpu-vm-instance-size

dcmvdbekerom:
Tests are now working! Still failing for the moment, but at least they are now testing the actual code. The first GPU test is passes, the second (multiple GPU plots) still has issues.

dcmvdbekerom:
Tests are now passing! This means that the Vulkan driver is working on Linux and tested on the CI server.
As briefly mentioned above, in absence of GPU hardware the Linux Vulkan drivers defaul to a CPU renderer `llvm-pipe`, which is used by Travis to run the GPU tests. 

There was an issue with some erratic output that was produced about ~50% of the time. This was related to not zeroing the buffers before executing the FFT's. Currently the buffers are zeroed _during_ iteration, which may be excessive -- zeroing during initialization might suffice.

minouHub:
I tried to do a GPU computation with exomol. It failed because the self broadening is not imported correctly. @dcmvdbekerom can you add that to the to-do list. I'll take a look if you want me to

dcmvdbekerom:
> I tried to do a GPU computation with exomol. It failed because the self broadening is not imported correctly. @dcmvdbekerom can you add that to the to-do list. I'll take a look if you want me to

@minouHub I think I need your help with this!

dcmvdbekerom:
Reallocate vulkan buffer:

*destroy old buffer*:
vkDestroyBuffer
vkFreeMemory

*allocate new buffer*
vkCreateBuffer
vkAllocateMemory
vkBindBufferMemory

*update references* 
Update Vulkan objects or structure that reference the old buffer (e.g. descriptor sets, bindings)

minouHub:
> >   I tried to do a GPU computation with exomol. It failed because the self broadening is not imported correctly. @dcmvdbekerom can you add that to the to-do list. I'll take a look if you want me to

> 
> @minouHub I think I need your help with this!

This should be solved when #716 and #720 will be merged!

erwanp:
Working for me : `plot_gpu_recalc.py`

image

radis/gpu/vulkan/shaders/cmdFillLDM.comp
```diff
@@ -0,0 +1,113 @@
...
+}
+#endif
+
+void main() {
+
+	uint i = gl_GlobalInvocationID.x;
+
+	if (i >= init_d.N_lines) return;
+
+    uint iso = floatBitsToUint(database_d[i]); // iterpret data as uint
```
erwanp, line 65: maybe comment this part ?

radis/gpu/vulkan/shaders/cmdApplyLineshapes.comp
```diff
@@ -0,0 +1,75 @@
...
+layout(std430, binding = 3) readonly buffer SSBO1 {
+   complex64_t data_LDM_FT_d[];
+};
+
+layout(std430, binding = 4) writeonly buffer SSBO2 {
+   complex64_t spectrum_FT_d[];
+};
+
+
+void main() {
```
erwanp, line 47: comment here too, mention Algorithm

.travis.yml
```diff
@@ -82,6 +83,61 @@ stages: #This sequence defines the order of the stages
 
 jobs:
   include:
+# <<<<<<< HEAD
```
erwanp, line 86: pb here

radis/test/gpu/test_gpu.py
```diff
@@ -132,21 +125,57 @@ def test_multiple_gpu_calls():
...
+##    print("\n")
+##
+##    if plot:
+##        plot_diff(s_list[0], s_list[-2], wunit="nm", method="diff")
+##
+##    assert (
+##        abs(s_list[0].get_power() - s_list[-2].get_power()) / s_list[0].get_power()
+##        < 1e-5
+##    )
+##    assert s_list[0].get_power() > 0
```
erwanp, line 170:
```suggestion
```

radis/lbl/loader.py
```diff
@@ -682,6 +682,7 @@ class DatabankLoader(object):
         "dataframe_type",
         "df0",
         "df1",
+        "gpu_backend",
```
erwanp, line 685: check if needed?

erwanp (changes requested):
on frontend side check

- calc_spectrum : https://github.com/radis/radis/blob/develop/radis/lbl/calc.py#L67C1-L67C16
- add argument exit_gpu=False to https://github.com/radis/radis/blob/develop/radis/lbl/calc.py#L958

dcmvdbekerom:
We decided to drop support for the CUDA backend, see issue here: https://github.com/radis/radis/issues/758

radis/lbl/factory.py
```diff
@@ -1063,8 +1058,17 @@ def eq_spectrum_gpu(
         Other Parameters
         ----------------
         backend: str
-            if ``'cuda-gpu'``, set CUDA as backend to run code on Nvidia GPU.
-            if ``'cuda-cpu'``, execute the GPU code on the CPU (useful for development)
+            currently only ``'vulkan'`` backend is supported.
+            ``'gpu-cuda'`` and``'cpu-cuda'`` used to be available to switch to a CUDA backend,
+            but have been deprecated in favor of the Vulkan backend.
+        device_id: int, str
+            Select the Vulkan device. If ``int``, specifies the device index, which is printed when Vulkan is loaded.
```
erwanp, line 1065:
```suggestion
            Select the GPU device. If ``int``, specifies the device index, which is printed when Vulkan (the GPU backen d) is loaded.
```

added cleared message because users are not necessarily aware of what Vulkan is

erwanp (changes requested):

radis/lbl/calc.py
```diff
@@ -949,28 +956,22 @@ def _is_at_equilibrium():
...
+                f"mode= should be 'cpu' or 'gpu'; got {mode}"
             )
+            return
+        
+        eq_spectrum_mode = sf.eq_spectrum if mode == 'cpu' else sf.eq_spectrum_gpu
+        
+        s = eq_spectrum_mode(
+            Tgas=Tgas,
+            mole_fraction=mole_fraction,
+            path_length=path_length,
```
erwanp, line 971: pressure is missing (note : it as missing before already !?)

erwanp (changes requested):

radis/lbl/calc.py
dcmvdbekerom: Good point, same for non-eq it seems..
I updated it

minouHub:
From a user perspective, I read the docs, played with the functionality and had fun. No bug. Tried HITRAN, HITEMP, and ExoMol. 
This is however not working for the GEISA database probably because the database does not provide all the columns we expect in the other databases.

dcmvdbekerom (commented):
All changes applied

dcmvdbekerom:
> From a user perspective, I read the docs, played with the functionality and had fun. No bug. Tried HITRAN, HITEMP, and ExoMol. This is however not working for the GEISA database probably because the database does not provide all the columns we expect in the other databases.

Glad it's (mostly) working! The geisa issue (like the exomol issue before) is not technically a GPU issue so I propose we solve that in a separate PR.

For now, I think it's done so I'll merge it!

radis/lbl/calc.py
erwanp: @dcmvdbekerom

dcmvdbekerom: It's updated in the development branch: https://github.com/radis/radis/blob/7350e1702cd02b2cde0e993163335e3363bafa7a/radis/lbl/calc.py#L968

Issue #127, opened 2020-08-31
Add option to normalize at one point only


erwanp:
Extension of previous PR :

--- 

(details)
I would make clearer the if ... then ... else here

```
 if len(wrange) == 2:
             blabla
 elif isinstance(wrange, (int, float)):
             blabla
  else:
             UserWarning

```

_Originally posted by @minouHub in https://github.com/radis/radis/pull/122#discussion_r473656658_

4the4ryushin:
is the issue still open?

erwanp:
yes

minouHub:
This issue was closed with #627

PR #627, opened 2023-10-22
Add an option to normalize at one point

E-W-Jones:
### Description

This pull request is to address #127, adding an option to normalize at a single point. I've also added a few more lines to `test_spectrum` to test this.

I made two choices when doing this that might not be ideal:
1. No matter what the supplied `wrange` value is, we find the closest point in the spectrum and normalise to that. Other options could include testing if the given point is close enough to an existing one or using interpolation between existing points, but that makes it all much more complicated. Seems simplest to leave it up to the user to check they're happy.
2. Using the `"max"`, `"mean"` or `"area"` options for the `normalize_how` argument don't make sense to me because it's only one value. As `"max"` is the default option nothing happens using this. If you use `"mean"` or `"area"` then it'll warn you that it's ignoring them and uses `"max"` anyway. This again seems like the simplest option.

The test is simple, I've taken a wavelength value at a given index, normalized around this point, and tested in the normalized spectrum at this index is close to 1.

### DIFF
--- a/radis/spectrum/spectrum.py
+++ b/radis/spectrum/spectrum.py
@@ -5260,8 +5260,9 @@ class Spectrum(object):
             how to normalize. ``'max'`` is the default but may not be suited for very
             noisy experimental spectra. ``'area'`` will normalize the integral to 1.
             ``'mean'`` will normalize by the mean amplitude value
-        wrange: tuple
-            if not empty, normalize on this range
+        wrange: tuple of floats, float
+            normalize on this range if a tuple, else normalise at that point.
+            If an empty tuple, use the entire range.
         wunit: ``"nm"``, ``"cm-1"``, ``"nm_vac"``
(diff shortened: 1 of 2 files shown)


erwanp:
1. Ok with the choice you made in 1. 
2. RADIS follows a "better crash than guess" philosophy; if some choice don't make sense ("mean" or "area") you should raise an error (telling the user that it doesn't make sense) rather than avoid.  Max or Mean actually change the resulting units; so it could have bad consequences there is a hidden choice.

E-W-Jones:
> @erwanp @E-W-Jones what is the status of this PR?

I updated so it crashes instead of warning as suggested, should be ready to merge if everyone's happy

erwanp:
Rebased on latest develop , now that tests have been fixed #648

minouHub:
Merging because this branch passed the tests :)

PR #633, opened 2023-11-03
(too) simple solution to accelerate potential halt of travis

minouHub:
### Description
Here is a simple solution to accelerate the potential crash of Travis during the development phases of a PR. In the first stage, only the tests marked as "fast" are performed. If they pass, only the others "not fast", are tested.

Let me know what you think @erwanp @dcmvdbekerom

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,9 +51,21 @@ jobs:
         # Your test script goes here
         - echo ">>> Run tests"
         # use XVFB to have headless display port, and still run the Matplotlib tests.
-        - xvfb-run -a pytest --cov=./ --durations=10
+        - xvfb-run -a pytest -m "fast" --cov=./ --durations=10
         # --durations=N to print the slowest N tests
         # lookup 'addopts' in setup.cfg>[tools:pytest] for default tests
+
+      services:
+        - xvfb
+
+      before_script:
+        - export MPLBACKEND=Agg
+
+    - stage: long_tests
+      name: Long Tests
+      script:
+        - echo ">>> Run tests which are long "
+        - xvfb-run -a pytest -m "not fast" --cov=./ --durations=10
 
       services:
         - xvfb


erwanp:
It might break the coverage (we'll see and ajust) but except from that seems seems like a good idea; let's see how much overhead time it takes to transfer the Machine from one stage to the next
If it's significant, another option would be to keep them in the same Test stage; but on two different lines : 

```
        - xvfb-run -a pytest --cov=./ --durations=10 -m "fast"
        - xvfb-run -a pytest --cov=./ --durations=10 -m "not fast"
```

erwanp (approved):

.travis.yml
```diff
@@ -51,7 +51,8 @@ jobs:
         # Your test script goes here
         - echo ">>> Run tests"
         # use XVFB to have headless display port, and still run the Matplotlib tests.
-        - xvfb-run -a pytest --cov=./ --durations=10
+        - xvfb-run -a pytest -m "fast" --cov=./ --durations=10
```
erwanp, line 54: @minouHub I think that by using this you might have replaced the default configuration which is defined in [setup.cfg](https://github.com/radis/radis/blob/develop/setup.cfg#L13) , and which has this value : 

```
-m "not needs_cuda and not download_large_databases and not needs_db_CDSD_HITEMP and not needs_db_CDSD_HITEMP_PCN and not needs_db_CDSD_HITEMP_PC and not needs_db_HITEMP_CO2_DUNHAM and not needs_db_HITEMP_CO_DUNHAM"
```
The value is quite long, but it explicitely discards some tests, including the `not needs_db_CDSD_HITEMP_PC`  that require databases that are not in the test files. These tests are responsible for the current fails in https://github.com/radis/radis/pull/647

I'd suggest to keep : 

```
        - xvfb-run -a pytest -m "fast" --durations=10  # no "fast" tests use the local databases 
        - xvfb-run -a pytest --cov=./ --durations=10  # uses default parameters in setup.cfg
 ```

minouHub: oh, I see! By default, all the tests you quote are skipped, but when I mention `-m "fast"` it overrides the default... Thanks for looking at it

PR #641, opened 2024-01-28
Make codebase Pandas 2.0-ready

leroyvn:
### Description

Fixes #600.

This PR fixes the parts of the codebase that are currently preventing an upgrade to Pandas 2.0 and later. Two things worth noting:

* The machine on which I ran the code doesn't have an nVidia GPU, so I did not run the full test suite.
* I only fixed the code itself, I did not bump requirements. FWIW, the Python version support list is as follows:
  * [2.0.x: 3.8 to 3.11](https://pandas.pydata.org/pandas-docs/version/2.0/getting_started/install.html#python-version-support)
  * [2.1.x: 3.9 to 3.11](https://pandas.pydata.org/pandas-docs/version/2.1/getting_started/install.html#python-version-support)
  * [2.2.x: 3.9 to 3.12](https://pandas.pydata.org/docs/getting_started/install.html#python-version-support)
* Happy to add more to this PR in case you're aware of more parts in the code that would need a fix.

PR #647, opened 2024-03-29
Remove install dependency + support Python 3.9, 3.10

minouHub:
This PR is to attempt to remove dependencies in the install of RADIS. Each push will be tested for one of the requirements of `environment.yaml`

- [x] python=3.8  - required by vaex, see https://github.com/vaexio/vaex/issues/2397
- [x] cython
- [x] pandas < 2.0 - solved in #641
- [x] publib - Not possible with python>3.8
- [x] progressbar2 - noted as needed by vaex, so it should be automatically managed by vaex (not us)

Also:

- [x] add limit `python<3.9` in `setup.py` for pip install

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -5,13 +5,13 @@ channels:
 - astropy
 - plotly
 dependencies:
-- python=3.8  # Needed by vaex - https://github.com/vaexio/vaex/issues/2397
+- python=3.8  # Needed by vaex 4.17 - https://github.com/vaexio/vaex/issues/2397
 - astropy>=4.3.1  # Unit aware calculations
 - astroquery>=0.4.6  # to fetch HITRAN databases
 - beautifulsoup4>=4.10.0 # parse ExoMol website
 - cantera>=2.5.1   # for chemical equilibrium computations
 - configparser
-- cython          # EP : really needed? pipreqs doesn't say so
+#- cython          # Not needed, see #647
 - habanero>=1.2.0  # CrossRef API to retrieve data from doi
 - h5py>=3.2.1   # load HDF5
 - joblib  # for parallel loading of SpecDatabase
@@ -28,8 +28,7 @@ dependencies:
 - seaborn   # other matplotlib themes
 - termcolor  # terminal colors
 - specutils
-- vaex-core
-#- vaex-core>=4.17.0   # load HDF5 files  (version needed to fix https://github.com/radis/radis/issues/598)
+- vaex-core>=4.17.0   # load HDF5 files  (version needed to fix https://github.com/radis/radis/issues/598)
 - vaex-hdf5
 - vaex-viz
 - pip
--- a/environment.yml
+++ b/environment.yml
@@ -5,7 +5,7 @@ channels:
 - astropy
 - plotly
 dependencies:
-#- python=3.8 #NM not really needed
+- python=3.8  # Needed by vaex - https://github.com/vaexio/vaex/issues/2397
 - astropy>=4.3.1  # Unit aware calculations
 - astroquery>=0.4.6  # to fetch HITRAN databases
 - beautifulsoup4>=4.10.0 # parse ExoMol website
--- a/environment.yml
+++ b/environment.yml
@@ -19,7 +19,7 @@ dependencies:
 - matplotlib
 - numpy
 - numba  # just-in-time compiler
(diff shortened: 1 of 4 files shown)


environment.yml
```diff
@@ -19,7 +19,7 @@ dependencies:
 - matplotlib
 - numpy
 - numba  # just-in-time compiler
-- pandas<2.0.0   # not ready for 2.0 yet ; some breaking compatibility, see issue #600
+#- pandas<2.0.0   # solved in #641
```
erwanp, line 22:
```suggestion
- pandas
```

erwanp: I'd keep pandas itself, but release the version constraint

erwanp (changes requested):

erwanp:
Also upgrade publib dependency in requirements.txt to 
```
publib>=0.4.0
```

I just published a new version of the package (https://pypi.org/project/publib/ ) that fixes a non compatibility with Matplotlib 3.8.0  
due to https://github.com/matplotlib/matplotlib/issues/26831

erwanp:
Regarding the test failures, see my comment here:  https://github.com/radis/radis/pull/633/files#r1555134045

requirements.txt
```diff
@@ -1,7 +1,7 @@
 # Installed by "pip install -e .[dev]" as well as "conda env create --file environment.yml"
 lxml           # parser used for ExoMol website
 hjson          # Json with comments (for default_radis.json)
-publib>=0.4.0  # Plotting styles for Matplotlib. Version, see #647
+publib>=0.3.2  # Plotting styles for Matplotlib. Version, update to 0.4.0 needed for matplotlib==3.8. However, only 0.3.2 is compatible with python==3.8 (see #647)
```
erwanp, line 4: therefore test on python>=3.11 or something?

minouHub: sorry I didn't get it.

minouHub: vaex is not compatible with python>3.8 (see original post). matplotlib==3.8 requires python>3.8, thus, we should hold this update for the moment

minouHub:
@erwanp, waiting for your final review on this. I will try to sort the docs by tonight

erwanp:
Which is actively being worked : https://github.com/vaexio/vaex/pull/2331

minouHub:
@erwanp did you change this PR name because you want to wait for vaex? I would have prefer to merge now and put the support of Python 3.11 on a to-do list

erwanp:
@minouHub  ok I updated the PR title

HajimeKawahara:
@minouHub @erwanp 

We have an installation problem because of 
- add limit python<3.9 in setup.py for pip install

What is the reason for this restriction?

https://github.com/HajimeKawahara/exojax/pull/481

minouHub:
Hello,
This restriction is introduced because `vaex` cannot be installed with Python > 3.8. Before this PR, the limit was present in the `conda` requirements but not in `pip`. In other words, if you did that, the **regular** pip install failed, see below:
```
conda create --name test_radis python==3.9 # or anything above 3.9
conda activate test_radis
pip install radis
```
The problem was that the error message was hard to understand. With this requirement, it's clearer. In the past, you could somehow trick the install with the developer install `pip -e .` but this is not stable to my opinion.

The main strategy I applied was to wait for `vaex` to update their package. The alternative if that is a big problem for you is to revert this modification, but remains the installation problem.

HajimeKawahara:
@minouHub Thanks. I see. The problem is that jax requires python >3.8. so, for us, it is a big problem. Can you revert this modification for a while if you do not have a big problem?

minouHub:
oh wait .... it worked .... I'm trying a quick fix in #651. Will keep you updated

HajimeKawahara:
Ah, I see. Thanks for the quick check!

minouHub:
Hello,
Very sorry for the mess. Two mistakes on my side. 

1. `vaex` is actually compatible with python 3.9 and 3.10. I don't understand how I missed that.
2. if `vaex` limits the python version, we don't need to state it on our side

HajimeKawahara:
@minouHub thank you so much!

erwanp:
> 2. if `vaex` limits the python version, we don't need to state it on our side

If and only if vaex is installed through Conda and not Pip on our side; which it is=> so all good. Well done for fixing it

requirements.txt
minouHub: note: vaex is compatible up to python 3.10

PR #652, opened 2024-05-31
Adding support for the Kurucz atomic database

code29563:
Starting the PR for this project


code29563:
The code can currently be used to produce plots of equilibrium spectra. The file [plot_atomic_spectrum.py](https://github.com/code29563/radis/blob/5340e1cd7146a408e8eddd13be1dfc6617694561/examples/plot_atomic_spectrum.py) provides a working example.

Issue:

Leaving the `optimization` parameter at its default value (`simple`), a large jump appears at the upper end of the wavenumber range, starting from `wmax` minus the `truncation`, for which the default value is 50, so in the example above the jump occurs at 4950 cm-1 (see the attached plot 'Figure_5.pdf'). This jump doesn't appear when the `optimization` is set to `None`, rather the plot is instead clipped at 4950 cm-1 ('Figure_6.pdf').

There's also a smaller gradual drop at the beginning that isn't sharp and doesn't appear to vary significantly when changing the value of `truncation` nor the `optimization` setting, but is nonetheless erroneous as can be seen by lowering `wmin`, moving the drop leftwards and revealing the actual spectrum at 2000 cm-1 (in the above example) to be rather flat.

Any ideas how to fix this @minouHub @dcmvdbekerom? I expect the issue is at least partly with [`BroadenFactory._apply_lineshape_LDM`](https://github.com/code29563/radis/blob/5340e1cd7146a408e8eddd13be1dfc6617694561/radis/lbl/broadening.py#L2017).

[Figure_6.pdf](https://github.com/user-attachments/files/15508267/Figure_6.pdf)
[Figure_5.pdf](https://github.com/user-attachments/files/15508268/Figure_5.pdf)

radis/lbl/base.py
```diff
@@ -94,6 +94,8 @@
 from radis.phys.convert import cm2J, cm2J_vaex, nm2cm, nm_air2cm
 from radis.phys.units_astropy import convert_and_strip_units
 from radis.spectrum.utils import print_conditions
+#hcperk = 1.4387773538277202  # hc/kB (cm K)
+Tref_original=296
```
minouHub, line 98: See line `self.input.Tref = 296` where Tref is already defined

radis/db/classes.py
```diff
@@ -217,6 +219,38 @@
...
+        charge_in_roman = {1: "I", 2: "II", 3: "III", 4: "IV", 5: "V"}.get(charge, "")
+        return f"{element_name}_{charge_in_roman}"
+
+    # If species is an element
+    else:
+        try:
+            atomic_symbol = species.split("_")[0]
+            el = getattr(periodictable, atomic_symbol)
+            if el.symbol == species:  # check if symbols match
+                return f"{species}_I"
```
minouHub, line 247: I don't think that we need this test with the package periodictable.

minouHub (commented):
Main issue is the management of the 'species' keyword and the potential backward compatibility

.gitignore
```diff
@@ -17,6 +17,8 @@
 *.pdf
 *.tif
 *.h5
+*.hdf5
+*.all
```
minouHub, line 21: Not sure I understand what `*all` is ignoring.

radis/lbl/broadening.py
```diff
@@ -372,6 +379,77 @@ def pressure_broadening_HWHM(
...
+    gamma_case2 =  gamma6 / (4*np.pi*c_CGS)
+
+    gamma_vdw = np.where(vdWdamp >= 0., gamma_case1, gamma_case2)
+    # df['shft'] = (1.0/3.0)*2*gamma #Konjević et al. 2012 §4.1.3.2
+    #print(gamma_vdw)
+    
+    if 'e-' in diluent:
+        gamma_stark = (10**gamSta) * P*1e6*diluent['e-'] / (k_b_CGS*T) / (4*np.pi*c_CGS)
+    else:
+        gamma_stark = 0
```
minouHub, line 444: Here is what you could do:
```python
# determine temperature coefficient. Adiabatic theory summarized in Rivière 2002 - Eq. (24-29) - doi:10.1016/S0022-4073(01)00173-X
if "is_ionized":
   n_Stark = -0.5 #according to adiabatic theory, dipole and quadripole moment scale with <1/v> = T**-0.5
else: #for neutrals
   n_Stark = 1/6 #according to adiabatic theory, dipole moment is constant and quadripole moment scale with <v>**(1/3) = T**(1/6)
gamma_stark = (10**gamSta) * (T/10000)**n_Stark * P*1e6*diluent['e-'] / (k_b_CGS*T) / (4*np.pi*c_CGS)
```

radis/lbl/broadening.py
```diff
@@ -372,6 +379,77 @@ def pressure_broadening_HWHM(
...
+
+    Returns:
+      gammma_rad, gamma_stark, gamma_vdw - Radiation, Stark, and Van der Waals HWHM [cm^-1]
+
+    * Reference of van der Waals damping constant (pressure/collision gamma):
+    *   Unsöld1955: https://ui.adsabs.harvard.edu/abs/1955psmb.book.....U
+    *   Kurucz+1981: https://ui.adsabs.harvard.edu/abs/1981SAOSR.391.....K
+    *   Barklem+1998: https://ui.adsabs.harvard.edu/abs/1998MNRAS.300..863B
+    *   Barklem+2000: https://ui.adsabs.harvard.edu/abs/2000A&AS..142..467B
+    *   Gray+2005: https://ui.adsabs.harvard.edu/abs/2005oasp.book.....G
```
minouHub, line 408: to update once we get the final version

examples/plot_atomic_spectrum.py
minouHub, line 1: I could not run the example. The ionization energies from NIST seems to be missing in git
`FileNotFoundError: [Errno 2] No such file or directory: 'c:\\users\\nicolas minesi\\python\\radis\\db\\NIST_Atomic_Ionization_Energies.txt'`

radis/lbl/calc.py
```diff
@@ -685,11 +719,15 @@ def _is_at_equilibrium():
         drop_columns = "auto"
 
     # Run calculations
-    sf = SpectrumFactory(
+    if CustomSpectrumFactory:
+        sfclass = CustomSpectrumFactory
+    else:
+        sfclass = SpectrumFactory
```
minouHub, line 725: to me, there is no other need than specifying a custom Lorentzian broadening, and this need is perfectly fit by `lbfunc`

minouHub (changes requested):

radis/lbl/broadening.py
code29563: >    n_Stark = 1/6 #according to adiabatic theory, dipole moment is constant and quadripole moment scale with <v>**(1/3) = T**(1/6)
How are the dipole and quadrupole gammas combined? If they're added together, then wouldn't this line be equivalent of scaling both the dipolar and quadrupolar gamma by `T**(1/6)` when the former is supposed to be constant?

examples/plot_atomic_spectrum.py
code29563: Added the required file in dfc140d9eacc1b51d00486d41abe3009b78985ed

radis/lbl/broadening.py
minouHub: Yes you're right. There is no good answer. Let's keep the dependence in T**[1/6] to make the comparison with other codes easier

examples/1_Spectra_handling/plot_exomol_spectrum.py
```diff
@@ -13,7 +13,7 @@
 s = calc_spectrum(
     1080,
     1320,  # cm-1
-    molecule="SiO",
+    species="SiO",
```
erwanp, line 16: is the `molecule=` syntax still valid ? 

If not, it should be (even if not advertised for in the docs; but it should be read from the **kwargs and converted into "species", obviously raising an error if both "molecule=" and "species=" are given)

examples/plot_atomic_spectrum.py
```diff
@@ -0,0 +1,103 @@
...
+
+s = calc_spectrum(
+    wlim[0],
+    wlim[1]+truncation,  # cm-1
+    species="O_I",  # Enter species name
+    Tgas=4000,  # K
+    databank="kurucz",
+    pressure=1.01325,
+    # optimization=None,
+    truncation=truncation,
```
erwanp, line 93: same here, can we keep the default value to make the code simpler and more readable ? (focusing on the physics; not the numerical implementation)

examples/1_Spectra_handling/plot_exomol_spectrum.py
minouHub: I agree. We discussed it and I'd like the species argument to be still valid

examples/plot_atomic_spectrum.py
minouHub: Agreed with Erwan. There is still a problem with the truncation at the moment and I believe it was introduced to circumvent the problem until it is sorted. @code29563 can you add a #to-do to not forget to remove this hack later?

polu73:
> The code can currently be used to produce plots of equilibrium spectra. The file [plot_atomic_spectrum.py](https://github.com/code29563/radis/blob/5340e1cd7146a408e8eddd13be1dfc6617694561/examples/plot_atomic_spectrum.py) provides a working example.
> 
> Issue:
> 
> Leaving the `optimization` parameter at its default value (`simple`), a large jump appears at the upper end of the wavenumber range, starting from `wmax` minus the `truncation`, for which the default value is 50, so in the example above the jump occurs at 4950 cm-1 (see the attached plot 'Figure_5.pdf'). This jump doesn't appear when the `optimization` is set to `None`, rather the plot is instead clipped at 4950 cm-1 ('Figure_6.pdf').
> 
> There's also a smaller gradual drop at the beginning that isn't sharp and doesn't appear to vary significantly when changing the value of `truncation` nor the `optimization` setting, but is nonetheless erroneous as can be seen by lowering `wmin`, moving the drop leftwards and revealing the actual spectrum at 2000 cm-1 (in the above example) to be rather flat.
> 
> Any ideas how to fix this @minouHub @dcmvdbekerom? I expect the issue is at least partly with [`BroadenFactory._apply_lineshape_LDM`](https://github.com/code29563/radis/blob/5340e1cd7146a408e8eddd13be1dfc6617694561/radis/lbl/broadening.py#L2017).
> 
> [Figure_6.pdf](https://github.com/user-attachments/files/15508267/Figure_6.pdf) [Figure_5.pdf](https://github.com/user-attachments/files/15508268/Figure_5.pdf)

I was working on integrating the Kurucz implementation with the Common API and, interestingly, the drop and jump seem to have disappeared - not sure why. You can test using [plot_atomic_spectrum.py](https://github.com/radis/radis/blob/285e6832fd3c31c7abbe16fc9c7764d5eeb32a54/examples/plot_atomic_spectrum.py) for any species (except Fe_I, as the [server-side linelist file that gets downloaded for that](http://kurucz.harvard.edu/atoms/2600/gf2600.all) appears to have a mistake that results in an exception being raised when parsing).

code29563:
> The code can currently be used to produce plots of equilibrium spectra. The file [plot_atomic_spectrum.py](https://github.com/code29563/radis/blob/5340e1cd7146a408e8eddd13be1dfc6617694561/examples/plot_atomic_spectrum.py) provides a working example.
> 
> Issue:
> 
> Leaving the `optimization` parameter at its default value (`simple`), a large jump appears at the upper end of the wavenumber range, starting from `wmax` minus the `truncation`, for which the default value is 50, so in the example above the jump occurs at 4950 cm-1 (see the attached plot 'Figure_5.pdf'). This jump doesn't appear when the `optimization` is set to `None`, rather the plot is instead clipped at 4950 cm-1 ('Figure_6.pdf').
> 
> There's also a smaller gradual drop at the beginning that isn't sharp and doesn't appear to vary significantly when changing the value of `truncation` nor the `optimization` setting, but is nonetheless erroneous as can be seen by lowering `wmin`, moving the drop leftwards and revealing the actual spectrum at 2000 cm-1 (in the above example) to be rather flat.
> 
> Any ideas how to fix this @minouHub @dcmvdbekerom? I expect the issue is at least partly with [`BroadenFactory._apply_lineshape_LDM`](https://github.com/code29563/radis/blob/5340e1cd7146a408e8eddd13be1dfc6617694561/radis/lbl/broadening.py#L2017).
> 
> [Figure_6.pdf](https://github.com/user-attachments/files/15508267/Figure_6.pdf) [Figure_5.pdf](https://github.com/user-attachments/files/15508268/Figure_5.pdf)

I was working on integrating the Kurucz implementation with the Common API and, interestingly, the drop and jump seem to have disappeared - not sure why. You can test using [plot_atomic_spectrum.py](https://github.com/radis/radis/blob/285e6832fd3c31c7abbe16fc9c7764d5eeb32a54/examples/plot_atomic_spectrum.py) for any species (except Fe_I, as the [server-side linelist file that gets downloaded for that](http://kurucz.harvard.edu/atoms/2600/gf2600.all) appears to have a mistake that results in an exception being raised when parsing).

minouHub:
I confirm that there is no more problem when comparing the two optimization modes (`None` vs `simple`)

```python
from radis import calc_spectrum, plot_diff
from radis.lbl.factory import SpectrumFactory
from radis.lbl.broadening import gamma_vald3
import numpy as np

wlim = (12050, 13900)
# truncation = 50
mole_fraction = 0.01
s = calc_spectrum(
    wlim[0],
    wlim[1],#+truncation,  # cm-1
    species="Fe_II",  # Enter species name
    Tgas=10000,  # K
    databank="kurucz",
    pressure=1.01325,
    # optimization=None,
    # truncation=truncation,
    diluent={'H':1-mole_fraction-1e-3, 'e-': 1e-3},
    mole_fraction=mole_fraction,
    path_length=15,
    # lbfunc=func2,
    # CustomSpectrumFactory=CustomSpectrumFactory,
    warnings={"AccuracyError": "ignore", "AccuracyWarning": "ignore"},
)
s2 = calc_spectrum(
    wlim[0],
    wlim[1],#+truncation,  # cm-1
    species="Fe_II",  # Enter species name
    Tgas=10000,  # K
    databank="kurucz",
    pressure=1.01325,
    optimization=None,
    # truncation=truncation,
    diluent={'H':1-mole_fraction-1e-3, 'e-': 1e-3},
    mole_fraction=mole_fraction,
    path_length=15,
    # lbfunc=func2,
    # CustomSpectrumFactory=CustomSpectrumFactory,
    warnings={"AccuracyError": "ignore", "AccuracyWarning": "ignore"},
    name="Optimization = None"
)
plot_diff(s, s2)
```
Figure 2024-06-24 084045

radis/lbl/base.py
```diff
@@ -2493,6 +2506,9 @@ def Qgas(self, df1, Tgas):
                 return df1["iso"].map(Qgas_dict)
 
         else:  # "iso" not in df:
+            if "gamvdW" in df1.columns:
```
minouHub, line 2509:
```suggestion
            if self.get_conditions()['dbformat'] == 'kurucz'
```
to test more rigorously if the species is an atom

radis/lbl/base.py
```diff
@@ -2521,6 +2528,9 @@ def Qref_Qgas_ratio(self, df1, Tgas, Tref):
 
         if "molecule" in df1.attrs:
             molecule = df1.attrs["molecule"]  # used for ExoMol, which has no HITRAN-id
+        # Specific to atomic data
+        elif "gamvdW" in df1.columns:
```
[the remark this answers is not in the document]
minouHub:
```suggestion
        elif self.get_conditions()['dbformat'] == 'kurucz':
```

radis/lbl/base.py
```diff
@@ -2555,6 +2574,28 @@ def Qref_Qgas_ratio(self, df1, Tgas, Tref):
...
+            Tref: reference temeparture
+
+        Returns:
+            line strength at Ttyp
+        """
+        if not Tref:
+            Tref = self.input.Tref
+        return Sij0 \
+            * np.exp(-hc_k*elower * (1./T - 1./Tref)) \
+            * np.expm1(-hc_k*nu_lines/T) / np.expm1(-hc_k*nu_lines/Tref) / qr
```
minouHub, line 2598:
```suggestion
    
```
This function is copied from ExoJax and not used

minouHub, line 2598:
```suggestion
   
```
This function is copied from ExoJax and not used.

radis/lbl/base.py
```diff
@@ -2846,7 +2879,7 @@ def Qneq_Qvib_Qrotu_Qrotl(
                     # ... make sure PartitionFunction above is calculated with the same
                     # ... temperatures, rovibrational distributions and overpopulations
                     # ... as the populations of active levels (somewhere below)
-                    print(df.columns)
+                    #print(df.columns)
```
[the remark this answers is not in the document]
minouHub:
```suggestion
```

radis/lbl/factory.py
```diff
@@ -763,13 +813,14 @@ def eq_spectrum(
             raise ValueError(
                 "Tgas should be float or Astropy unit. Got {0}".format(Tgas)
             )
-        self.input.rot_distribution = "boltzmann"  # equilibrium
-        self.input.vib_distribution = "boltzmann"  # equilibrium
+        if not self.input.isatom:
+            self.input.rot_distribution = "boltzmann"  # equilibrium
+            self.input.vib_distribution = "boltzmann"  # equilibrium
```
minouHub, line 818:
```suggestion
            self.input.rot_distribution = None 
            self.input.vib_distribution = None 
```

radis/lbl/factory.py
```diff
@@ -763,13 +813,14 @@ def eq_spectrum(
...
 
-        # Get temperatures
+            # Get temperatures
+            self.input.Tvib = Tgas  # just for info
+            self.input.Trot = Tgas  # just for info
```
minouHub, line 822:
```suggestion
            self.input.Tvib = None
            self.input.Trot = None
```

radis/lbl/factory.py
```diff
@@ -763,13 +813,14 @@ def eq_spectrum(
             raise ValueError(
                 "Tgas should be float or Astropy unit. Got {0}".format(Tgas)
             )
-        self.input.rot_distribution = "boltzmann"  # equilibrium
-        self.input.vib_distribution = "boltzmann"  # equilibrium
+        if not self.input.isatom:
```
minouHub, line 816: This is failing when ran with molecules

code29563:
In addb90c0dc2fe3bdfb915defc6e548802ef0df1f, I've added support for using the partiton function files at http://kurucz.harvard.edu/atoms/xxyy/partfnxxyy.dat (for a given species code xxyy). In doing so I also encountered [a separate issue](https://github.com/radis/radis/issues/661).

radis/lbl/calc.py
```diff
@@ -384,8 +391,30 @@ def calc_spectrum(
...
+    if "molecule" in kwargs:
+        if species is not None:
+            if species != kwargs['molecule']:
+                raise Exception("Both `molecule` and `species` arguments have been given and aren't equal, but `molecule` is deprecated and `species` is just its replacement.")
+        species = molecule = kwargs['molecule']
+        warn(
+            DeprecationWarning(
+                "`molecule` is deprected - use `species` instead"
+            )
+        )
```
erwanp, line 403: No, our coding philosophy should be "Better Crash Than Hide" 
If users added a different molecule and species; then something is wrong. Let's raise an explicit error rather than try to fix it behind their back.

radis/lbl/calc.py
```diff
@@ -538,6 +546,9 @@ def _check_molecules_are_consistent(
             mole_fraction, diluent, molecule
         )
 
+        if "molecule" in kwargs_molecule:
+            del kwargs_molecule["molecule"]
+
```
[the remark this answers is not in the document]
erwanp: we usually do 
```
molecule = kwargs_molecule.pop("molecule")
```
which reads and deletes it from the dictionary  (would be done [L394](https://github.com/radis/radis/pull/652/files/285e6832fd3c31c7abbe16fc9c7764d5eeb32a54#diff-0ec4ab793955b21d70e149abc006fc780f68eca5f98897df88af429c1dcd3dbaR394) for instance)

radis/lbl/loader.py
```diff
@@ -1498,6 +1510,52 @@ def fetch_databank(
...
+            if isotope == "all":
+                isotope_list = None
+            else:
+                isotope_list = ",".join([str(k) for k in self._get_isotope_list()])
+            # local_paths, df = fetch_kurucz(
+            #     molecule,
+            #     isotope=isotope_list,
+            #     engine=memory_mapping_engine
+            # )
+            df, local_paths = fetch_kurucz_new(
```
erwanp: should be `fetch_kurucz`, we should not keep old code

radis/lbl/loader.py
```diff
@@ -2459,6 +2511,17 @@ def load_and_concat(files, output="pandas"):
...
+                    #     # if self.dataframe_type == "pandas":
+                    #     #     engine = "pytables"
+                    #     # elif self.dataframe_type == "vaex":
+                    #     #     engine = "vaex"
+                    #     # if isotope == "all":
+                    #     #     isotope_list = None
+                    #     # else:
+                    #     #     isotope_list = ",".join([str(k) for k in self._get_isotope_list()])
+                    #     # df = fetch_kurucz(self.input.species, isotope=isotope_list, engine=engine)[1]
+                    #     raise NotImplementedError("use fetch_databank('kurucz')")
```
erwanp, line 2532: @code29563 check this part, delete if not needed

radis/lbl/loader.py
```diff
@@ -2555,7 +2618,8 @@ def load_and_concat(files, output="pandas"):
         minwavdb = df.wav.min()
 
         # ... Explicitly write molecule if not given
-        if self.input.molecule in [None, ""]:
+        if self.input.molecule in [None, ""] and self.input.species not in [None, ""]:
```
erwanp, line 2629: there should not be `input.molecule` and `input.species`. Only use `species`

examples/plot_atomic_spectrum.py
```diff
@@ -0,0 +1,96 @@
...
+from radis.lbl.factory import SpectrumFactory
+from radis.lbl.broadening import gamma_vald3
+import numpy as np
+
+wlim = (12850, 13120)
+# truncation = 50
+mole_fraction = 0.01
+
+def func1(**kwargs):
+    """An example implementing the default broadening formula and values in SpectraPlot"""
```
erwanp, line 29: Do you mean : 
```suggestion
    """An example implementing the default broadening formula and values of SpectraPlot (https://spectraplot.com/)"""
```

examples/plot_atomic_spectrum.py
```diff
@@ -0,0 +1,96 @@
...
+    databank="kurucz",
+    pressure=1.01325,
+    # optimization=None,
+    # truncation=truncation,
+    diluent={'H':1-mole_fraction-1e-3, 'e-': 1e-3},
+    mole_fraction=mole_fraction,
+    path_length=15,
+    # lbfunc=func2,
+    # CustomSpectrumFactory=CustomSpectrumFactory,
+    warnings={"AccuracyError": "ignore", "AccuracyWarning": "ignore"},
```
erwanp: @code29563 @minouHub  I don't like this design choice of adding a CustomSpectrumFactory parameter in calc_spectrum

`calc_spectrum` should remain simple. Advance users can use the SpectrumFactory directly. This example seems to be advanced (we change broadening parameters) and therefore should use SpectrumFactory directly. 

Also the name & description of the example should explain it is a complex example meddling with internal physical equations.

radis/lbl/calc.py
```diff
@@ -293,6 +298,8 @@ def calc_spectrum(
                 sf.eq_spectrum(...)  #  new calculation without reloading the database
     engine : string
         Vaex or Pandas . Default Pandas, if engine is vaex memory performance is improved
+    CustomSpectrumFactory: class
```
erwanp, line 301: See my comment [there](https://github.com/radis/radis/pull/652/files/285e6832fd3c31c7abbe16fc9c7764d5eeb32a54#diff-08e99f95955a5697a50f8f766e787f3440bb5bda43ccf32f9d045005fb3284b7R92), but I very much dislike the idea of adding such a parameter in `calc_spectrum`. It kills the intent of calc_spectrum to be as simple as possible (and it isn't as simple as it should be already). Let's discuss it.

radis/io/kurucz.py
```diff
@@ -0,0 +1,162 @@
...
+
+"""
+
+import os
+from os.path import abspath, expanduser, join
+
+from radis.api.hdf5 import DataFileManager
+from radis.api.kuruczapi import AdBKurucz,get_atomic_number,get_ionization_state, KuruczDatabaseManager
+
+def fetch_kurucz_new(
```
erwanp: see my other comment : there should not be two functions "old" and "new". Just replace the old code with the new one.

erwanp (changes requested):

radis/lbl/calc.py
minouHub: Agreed with Erwan. If the user specifies both, just crash with an explicit error. 
Also, I'm not sure that we will depreciate the "molecule" input in the near future, so I would simply ask the user to pick one or the other.

radis/lbl/loader.py
minouHub: agreed

minouHub:
To start fresh on a few comment of @erwanp.

@code29563 added 2 new arguments to the function `calc_spectrum`: `lbfunc=` and `CustomSpectrumFactory=`. 

- `CustomSpectrumFactory=`: as discussed in meetings, I don't think this argument is useful. 
- `lbfunc=`: as discussed in meetings, I think this is super useful! However, I agree that it is advanced.
 
Concerning the simplicity of calc_spectrum. I remember that we had the same debate when adding the option of using the engine pandas or vaex. In the end, we preferred to not show this argument in `calc_spectrum` to keep it simple and let the advanced user modify it in `SpectrumFactory`.

**Long story short:** I agree with Erwan. The argument `lbfunc=` should be moved to `SpectrumFactory`.

radis/lbl/factory.py
```diff
@@ -171,7 +176,7 @@ class SpectrumFactory(BandFactory):
         Half-width over which to compute the lineshape, i.e. lines are truncated
         on each side after ``truncation`` (:math:`cm^{-1}`) from the line center.
         If ``None``, use no truncation (lineshapes spread on the full spectral range).
-        Default is ``300`` :math:`cm^{-1}`
+        Default is ``50`` :math:`cm^{-1}`
```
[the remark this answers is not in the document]
code29563: It had already been done in the code in 00720e0139614095d7b13b40173d939791c6aa1f, it was just the docstring that wasn't updated

radis/lbl/factory.py
code29563: removed as inapplicable to equilibrium spectra

radis/lbl/factory.py
code29563: removed as inapplicable to equilibrium spectra

radis/lbl/calc.py
code29563: > No, our coding philosophy should be "Better Crash Than Hide" If users added a different molecule and species; then something is wrong. Let's raise an explicit error rather than try to fix it behind their back.

That's precisely what we're doing here: `if species != kwargs['molecule']` then raise an exception, otherwise if they are equal, then it doesn't cause a problem per se, it's just that giving both arguments is redundant so the `DeprecationWarning` is given.

Or should we be raising an exception even when both are equal?

radis/lbl/loader.py
code29563: already fixed: https://github.com/radis/radis/pull/652/commits/addb90c0dc2fe3bdfb915defc6e548802ef0df1f#diff-8694bc131153e428bcabb11b9ff0b5ab0bc832e1b1e5f9431a5c95c936fba48eR21

radis/io/kurucz.py
code29563: already fixed: https://github.com/radis/radis/pull/652/commits/addb90c0dc2fe3bdfb915defc6e548802ef0df1f#diff-8694bc131153e428bcabb11b9ff0b5ab0bc832e1b1e5f9431a5c95c936fba48eR21

radis/db/classes.py
```diff
@@ -217,6 +219,38 @@
 """ str: list of [HITRAN-2020]_ molecules. """
 
 
+def is_atom(species):
+    if "_I" in species:
```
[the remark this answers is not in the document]
code29563: No longer called in cases where `species` would be `None`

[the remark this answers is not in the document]
code29563: No longer called in cases where `species` is `None`

.gitignore
code29563: was ignoring the 'gfxxyy.all' files for the Kurucz linelists when they were being downloaded to the repo folder - no longer applicable after integrating with Common API.

code29563:
> @code29563 added 2 new arguments to the function `calc_spectrum`: `lbfunc=` and `CustomSpectrumFactory=`.
> ...
> **Long story short:** I agree with Erwan. The argument `lbfunc=` should be moved to `SpectrumFactory`.

To clarify, `lbfunc` is a parameter of `SpectrumFactory`, not `calc_spectrum`, but it can be specified in `calc_spectrum` and then gets passed among `kwargs` on to `SpectrumFactory`

radis/lbl/calc.py
erwanp: Sorry I was too quick in reading your code. This is exactly what we should do. Keep it as is ! You can resolve this conversation

code29563:
> @code29563, I suggested a test in [162fcda](https://github.com/radis/radis/commit/162fcda72e521eb490cc80324e0208954a85053b) (and [a45bdf3](https://github.com/radis/radis/pull/652/commits/a45bdf38b18ce3af119f712df5db6e04ef26dc9e)). The comparison with SpectraPlot looks good to me! There might be a difference between the two implementations of the shift but that's not really a problem. Well done!
> 
> image

Got the example for radiance working too, but this close result only seems to be the case for O_I and not necessarily other species, e.g. Fe_I.

Should work fine now with both `fetch_databank` and `load_databank` without setting "ALLOW_OVERWRITE" to `True`. `parfunc` gets set automatically for the former, and [the columns relevant for the Kurucz dataframes are different to those for molecular dataframes](https://github.com/code29563/radis/blob/9abb6d212b5ae9d3785d006f32e1e9e56f4cc4ba/radis/lbl/loader.py#L1518).

minouHub:
Atomic oxygen at 777 nm is very well known. For other wave range or atoms, I do expect the two databases to be different sometimes.

minouHub:
#656 is merged. you can rebase on develop to pass the exomol tests

radis/db/classes.py
code29563: For checking whether the species in that form is an atom, in which case append '_I' to put it in spectroscopic notation, or not in which case return as is.

code29563:
Work on equilibrium spectra seems to be done. Confirm whether the current handling of `species` vs `molecule` is sufficient.

minouHub:
you could use this paper: https://adsabs.harvard.edu/full/1995ASPC...78..205K
[1] R. L. Kurucz, « The Kurucz (Smithsonian) Atomic and Molecular Database », Highlights of Astronomy, vol. 10, p. 579‑579, 1995, doi:10.1017/s1539299600012119

code29563:
I take it that while Vaex shouldn't be a hard requirement, and hence users should be able to run tests locally without Vaex installed, it is nevertheless still intended to remain supported, and hence the tests on Travis presumably should be running the Vaex tests too? From the logs, it looks like the error on Python 3.11 is due to Vaex not being installed.

minouHub:
> I take it that while Vaex shouldn't be a hard requirement, and hence users should be able to run tests locally without Vaex installed, it is nevertheless still intended to remain supported, and hence the tests on Travis presumably should be running the Vaex tests too? From the logs, it looks like the error on Python 3.11 is due to Vaex not being installed.

As of July 2024, vaex is not supported on python 3.11. Thus, for those who install radis on python 3.11:

- RADIS should be able to run without vaex
- The tests should pass on python 3.11, because RADIS is intended to be working on 3.11

I re-started the tests. The last ones failed because of a connection problem.

environment.yml
```diff
@@ -9,7 +9,7 @@ dependencies:
 - astropy>=4.3.1  # Unit aware calculations
 - astroquery>=0.4.6  # to fetch HITRAN databases
 - beautifulsoup4>=4.10.0 # parse ExoMol website
-- cantera>=2.5.1   # for chemical equilibrium computations
+- cantera>=2.5.1
```
minouHub, line 12:
```suggestion
- cantera>=2.5.1 # for chemical equilibrium computations
```

examples/1_Spectra_handling/Kurucz_partition_functons_potential_lowering.py
```diff
@@ -0,0 +1,123 @@
...
+# and where you can calculate spectra at a particular temperature with [Barklem & Collet 2016]_:
+#
+
+s = sf.eq_spectrum(Tgas=99)
+s.plot("radiance_noslit", wunit="cm-1")
+
+#%%
+# you encounter an error at the same temperature with the Kurucz tables:
+#
+sf.input.potential_lowering = -500
```
minouHub, line 56:
```suggestion
sf.input.potential_lowering = -500 #choose from [-500, -1000, -2000, -4000, -8000, -16000, -32000] cm-1
```

radis/levels/partfunc.py
```diff
@@ -1146,6 +1151,57 @@ def _at(self, T):
...
+    def __init__(self, species, pfpath = None):
+        super(PartFuncKurucz, self).__init__()
+        # Load data in constructor
+        self.species = species
+        self.pfTdat, self.pfdat = load_pf_Barklem2016()
+        # Locate the row for the specific atom and ionization state
+        try:
+            pf_atom = self.pfdat.loc[f"{species}"]
+        except KeyError:
+            warn("Note: the partition functions from Barklem & Collet (2016) don't include this species.")
```
minouHub, line 1167: raise an error on this. It is normally not possible to compute a spectrum without the partition function

minouHub (changes requested):
sorry, a short review for the moment

examples/1_Spectra_handling/Kurucz_partition_functons_potential_lowering.py
code29563: https://github.com/radis/radis/pull/652/files/f6527524d72bd319761f27a75ee1033de68f97f6#diff-5f1942f499bb46dca336a03f3664a0792b23a87a65cd15b82c6f980c7985c803R13

examples/1_Spectra_handling/custom_Lorentzian_broadening.py
```diff
@@ -0,0 +1,96 @@
...
+
+#%%
+# Now compare the result with that of the default handling of broadening in RADIS by removing the `lbfunc` parameter, recalculating the spectrum without it, and plotting the diff between the results:
+#
+sf.params.lbfunc = None
+s_default = sf.eq_spectrum(4000)
+
+from radis import plot_diff
+
+plot_diff(s1, s_default, label1='s1', label2='s_default')
```
minouHub, line 47: I would prefer to directly use `SpectrumFactory` here. Again, we both know it's possible to use calc_spectrum and kwargs for that, but let's keep it simple for calc_spectrum, and more advanced for SpectrumFactory.

- [ ] Use only SpectrumFactory
- [ ] Start with the default parameters `s1 = sf.eq_spectrum(Tgas = 4000, name = "RADIS default - Kurucz parameters")`
- [ ] Then compute with the custom lbfun, `s2 = sf.eq_spectrum(Tgas = 4000, name = "SpectraPlot default")`
- [ ] Compare as you did

radis/io/kurucz.py
```diff
@@ -0,0 +1,170 @@
...
+        error = False
+        if cache == "regen" or not local_files:
+            error = True
+        files_to_check = local_files + pf_path
+        if ldb.get_missing_files(files_to_check):
+            error = True
+        if not pf_path:
+            get_pf_files = False #assume partition function file is either unavailable or undesired for current species
+        if error:
+            raise Exception('Changes are required to the local database, and hence updating the registered entry, but "ALLOW_OVERWRITE" is False')
```
minouHub, line 80: Add suggestion such as 
```python
"Type `radis.config["ALLOW_OVERWRITE"]=True` if you want to change this parameter"
```

radis/io/kurucz.py
```diff
@@ -0,0 +1,170 @@
...
+        ldb.remove_local_files(local_files + pf_path)
+    ldb.check_deprecated_files(
+        ldb.get_existing_files(local_files),
+        auto_remove=True if cache != "force" else False,
+    )
+    
+    get_main_files = True
+
+    if len(local_files) > 1 or len(urlnames) > 1:
+        raise Exception('only 1 database file is expected')
```
minouHub, line 94: Add suggestion to the user, such as "delete this folder"

minouHub (changes requested):
Partial review - mainly on the examples

minouHub (changes requested):
A few other comments:
- [x] `test_Kurucz_vs_Spectraplot.py` and `test_Kurucz_vs_Spectraplot_O_I_4000K.py ` have some redundancy. Please merge them.
- [x] Start the name of these files with "plot_" so that the plots of these examples show up in the documentation: `examples/1_Spectra_handling/custom_Lorentzian_broadening.py` and `examples/1_Spectra_handling/Kurucz_partition_functons_potential_lowering.py`

minouHub:
If you rebase on the develop branch, the doc should be building now

examples/1_Spectra_handling/Kurucz_partition_functons_potential_lowering.py
```diff
@@ -0,0 +1,123 @@
...
+
+When such tables are unavailable or `potential_lowering` is `None` (the default), the partition functions are instead interpolated from [Barklem & Collet 2016]_ Table 8.
+
+The value can thereafter be changed on the fly by changing the `sf.input.potential_lowering` attribute of the :py:class:`~radis.lbl.factory.SpectrumFactory` instance ``sf``, the result of which is reflected the next time partition function interpolator's `._at` method is used, without any need to re-initialise it.
+
+Allowable values are usually: -500, -1000, -2000, -4000, -8000, -16000, -32000. This is based on what's been encountered in the partition function tables of the species checked so far, but documentation of the Kurucz linelists is unclear on whether this is the case for all species.
+
+The temperature ranges of the partition functions from [Barklem & Collet 2016]_ may differ from those of the dedicated Kurucz tables, so where you can calculate spectra at a particular temperature with the Kurucz tables:
```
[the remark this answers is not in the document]
code29563: Kurucz temperature values seem to typically start at 100 actually, hence the choice of 99 in the example

radis/io/kurucz.py
code29563: The issue shouldn't arise if the db entry was automatically registered by RADIS. This is intended more for a case where the user provides their own custom settings.

radis/api/dbmanager.py
```diff
@@ -421,6 +425,33 @@ def download_and_parse_one_file(urlname, local_file, Ndownload):
...
+        urlname,
+        local_file,
+        pbar_active=True,
+        pbar_t0=0,
+        pbar_Ntot_estimate_factor=None,
+        pbar_Nlines_already=0,
+        pbar_last=True,
+
+    ) -> list:
+        """This function should be overwritten by the DatabaseManager subclass
```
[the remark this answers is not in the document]
code29563: Just used the same template as `fetch_urlnames`: https://github.com/radis/radis/blob/41aca80eb73da329a052d75e61085c481d2f6795/radis/api/dbmanager.py#L253

radis/lbl/broadening.py
```diff
@@ -383,6 +388,126 @@ def pressure_broadening_HWHM(
...
+        transition wavenumber
+    elower: array or Vaex expression   (cm-1)        [length N]
+        excitation potential (lower level)
+    ionE: array or Vaex expression   (eV)        [length N]
+        ionization potential
+    gamRad: array or Vaex expression   (s-1)        [length N]
+        log of the half-width half maximum coefficient (HWHM) for radiative broadening
+    gamSta: array or Vaex expression   (s-1 * cm-3)         [length N]
+        log of the half-width half maximum coefficient (HWHM) per electron density for Stark broadening, at 10000 K
+    gamVdW: array or Vaex expression   (s-1 * cm-3)        [length N]
```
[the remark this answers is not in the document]
code29563: Sort of following the template in `doppler_broadening_HWHM`, `lorentzian_lineshape`, `lorentzian_FT`, `voigt_broadening_HWHM`, `voigt_lineshape`, and `voigt_FT`: units in round brackets, shape in square brackets.

radis/lbl/factory.py
```diff
@@ -142,23 +143,26 @@ class SpectrumFactory(BandFactory):
...
+            from the database files being loaded. See the list of supported molecules
+            in :py:data:`~radis.db.MOLECULES_LIST_EQUILIBRIUM`
+            and :py:data:`~radis.db.MOLECULES_LIST_NONEQUILIBRIUM`.
+        For atoms:
+            The positive or neutral atomic species. It may be given in spectroscopic notation or any form that can be converted by :py:func:`~radis.db.classes.to_conventional_name`
         Default ``None``.
     isotope: ``int``, ``list``, ``str`` of the form ``'1,2'``, or ``'all'``
         isotope id (sorted by relative density: (eg: 1: CO2-626, 2: CO2-636 for CO2).
-        See HITRAN documentation for isotope list for all species. If 'all',
+        For molecules, see [HITRAN-2020]_ documentation for isotope list for all species. For atoms, use the isotope number of the isotope (the total number of protons and neutrons in the nucleus) - use 0 to select rows where the isotope is unspecified, in which case the standard atomic weight from the ``periodictable`` module is used when mass is required. If ``'all'``,
```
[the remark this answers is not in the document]
code29563: I still haven't actually encountered a Kurucz linelist in the code that lists anything but the most common isotope, but this is to account for the possibility should it arise. It impacts the result of `get_molar_mass` which affects the Doppler broadening.

radis/lbl/factory.py
```diff
@@ -142,23 +143,26 @@ class SpectrumFactory(BandFactory):
...
+        For molecules:
+            molecule id (HITRAN format) or name. If ``None``, the molecule can be inferred
+            from the database files being loaded. See the list of supported molecules
+            in :py:data:`~radis.db.MOLECULES_LIST_EQUILIBRIUM`
+            and :py:data:`~radis.db.MOLECULES_LIST_NONEQUILIBRIUM`.
+        For atoms:
+            The positive or neutral atomic species. It may be given in spectroscopic notation or any form that can be converted by :py:func:`~radis.db.classes.to_conventional_name`
         Default ``None``.
     isotope: ``int``, ``list``, ``str`` of the form ``'1,2'``, or ``'all'``
         isotope id (sorted by relative density: (eg: 1: CO2-626, 2: CO2-636 for CO2).
```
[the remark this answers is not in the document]
code29563: Any idea why it's supposed to be given 'sorted by relative density', or is that a statement of how to derive the ID for a particular molecule?

radis/lbl/factory.py
```diff
@@ -142,23 +143,26 @@ class SpectrumFactory(BandFactory):
...
-        in :py:data:`~radis.db.MOLECULES_LIST_EQUILIBRIUM`
-        and :py:data:`~radis.db.MOLECULES_LIST_NONEQUILIBRIUM`.
+    species: ``int``, ``str``, or ``None``
+        For molecules:
+            molecule id (HITRAN format) or name. If ``None``, the molecule can be inferred
+            from the database files being loaded. See the list of supported molecules
+            in :py:data:`~radis.db.MOLECULES_LIST_EQUILIBRIUM`
+            and :py:data:`~radis.db.MOLECULES_LIST_NONEQUILIBRIUM`.
+        For atoms:
+            The positive or neutral atomic species. It may be given in spectroscopic notation or any form that can be converted by :py:func:`~radis.db.classes.to_conventional_name`
```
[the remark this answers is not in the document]
code29563: They're unsupported in the code in general considering how the input is parsed to spectroscopic notation which is then split into element name and (positive) charge.

radis/levels/partfunc.py
```diff
@@ -1146,6 +1151,57 @@ def _at(self, T):
...
+            Temp = self.pfT_values
+            Qvals = self.pf_values
+        
+        if T < Temp.min() or T > Temp.max():
+            raise ValueError(
+                f"The temperature {T} K is outside the tabulated range of the partition functions [{Temp.min()}, {Temp.max()}] K" + addmsg
+            )
+        return np.interp(T, Temp, Qvals)
+
+
```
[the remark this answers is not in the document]
code29563: Added outputting the list of available potential lowerings

examples/1_Spectra_handling/Kurucz_partition_functons_potential_lowering.py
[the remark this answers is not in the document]
code29563: I think typical lower bound of temperature range for Kurucz pfs is actually 100. Implication for allowable values of potential_lowering was if you want to set it to 'something' rather than 'None' (default as mentioned).

radis/lbl/calc.py
```diff
@@ -382,8 +393,18 @@ def calc_spectrum(
...
-
-    # ... wavelengths / wavenumbers
+    if "molecule" in kwargs:
+        if species is not None:
+            if species != kwargs['molecule']:
+                raise Exception("Both `molecule` and `species` arguments have been given and aren't equal, but `molecule` is deprecated and `species` is just its replacement.")
+        species = molecule = kwargs.pop("molecule") #remove molecule from kwargs
+        warn(
+            DeprecationWarning(
+                "`molecule` is deprected - use `species` instead"
```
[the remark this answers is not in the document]
code29563: Switched to treating `molecule` as an alias for `species` rather than deprecated

radis/lbl/loader.py
```diff
@@ -1036,14 +1047,15 @@ def fetch_databank(
...
+        parfuncfmt: ``'cdsd'``, ``'hapi'``, ``'exomol'``, ``'kurucz'`` or any of :data:`~radis.lbl.loader.KNOWN_PARFUNCFORMAT`
             format to read tabulated partition function file. If ``'hapi'``, then
             [HAPI]_ (HITRAN Python interface) is used to retrieve [TIPS-2020]_
             tabulated partition functions.
             If ``'exomol'`` then partition functions are downloaded from ExoMol.
+            If ``'kurucz'``, if dedicated partition functions are available with the linelists for the species, they are used, otherwise partition functions from Barklem & Collett 2016 are used. Also see the documentation for the `potential_lowering` parameter of :py:class:`~radis.lbl.factory.SpectrumFactory`.
             Default ``'hapi'``.
         parfunc: filename or None
-            path to a tabulated partition function file to use.
+            path to a tabulated partition function file to use. For kurucz linelists, this would be the path to a dedicated partition function file for the species (dependant on both temperature and potential lowering).
```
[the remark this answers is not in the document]
code29563: Wording was unclear here - I meant the tables are dependent on both temperature and potential lowering

radis/lbl/loader.py
```diff
@@ -1091,6 +1103,7 @@ def fetch_databank(
             and dropping them with ``drop_columns``.
             If ``diluent`` then all additional columns required for calculating spectrum
             in that diluent is loaded.
+            Only ``'all'`` is available for kurucz linelists.
             Default ``'equilibrium'``.
```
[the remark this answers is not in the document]
code29563: Rephrased with `parameter is ignored for Kurucz linelists`, as any other value just results in a warning and getting switched back to 'all', rather than raising an error

radis/api/dbmanager.py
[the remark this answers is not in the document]
minouHub: @erwanp, this was your original wording. Could you explain?

radis/api/kuruczapi.py
```diff
@@ -0,0 +1,490 @@
...
+    dtypes = (
+        np.float64,
+        np.float64,
+        "object",  # make species string rather than float for easier parsing
+        np.float64,
+        np.float64,
+        "object",
+        np.float64,
+        np.float64,
+        "object",
```
minouHub, line 408: Is it not better to use string rather than objects?

radis/lbl/broadening.py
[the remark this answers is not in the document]
minouHub: alright

radis/lbl/broadening.py
```diff
@@ -902,124 +1065,138 @@ def _calc_broadening_HWHM(self):
...
-                df,
-                pressure_atm,
-                mole_fraction,
-                Tgas,
-                Tref,
-                diluent,
-                diluent_broadening_coeff,
+            # Adds hwhm_voigt:
+            df["hwhm_voigt"] = (
+                olivero_1977(2 * df["hwhm_gauss"], 2 * df["hwhm_lorentz"]) / 2
```
minouHub, line 1197: seems like the function `_add_voigt_broadening_HWHM` is not used anymore in RADIS. Not saying it's good or bad, but we can discuss it.

minouHub (changes requested):
Partial review

radis/api/dbmanager.py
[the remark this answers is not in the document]
minouHub: Resolved because this is how RADIS was built initially

radis/lbl/broadening.py
minouHub: Sorted during meeting. @code29563 wrote the code in a more compact way

radis/lbl/factory.py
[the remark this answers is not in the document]
minouHub: It won't happen. The line position and the line intensity of atomic spectra do **not** depend on the isotope number. Only the doppler broadening depends on the mass of the nuclei. Given that databases provide linelist with position and intensity, it is not expected to have any information on the isotope number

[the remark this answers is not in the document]
minouHub: In conclusions, please simply provide what is the default for an atom. "For atoms, isotope number is..., e.g. for nitrogen isotope = 14"

radis/lbl/factory.py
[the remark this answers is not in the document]
minouHub: For 90% of applications, you only need the most abundant molecular isotopologue to reproduce an experiment. Thus, it is convinient to only download and compute the spectrum only for the first isotopologue. Thus, it makes sense to my why the "first isotopologue" corresponds by arbitrary definition in HITRAN to the most abundant one.

radis/lbl/factory.py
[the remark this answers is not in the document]
minouHub: Alright. Please apply my suggestion then 🙂

radis/levels/partfunc.py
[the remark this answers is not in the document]
minouHub: OK saw it. Please be more concise. It's good to be polite ("you may want to") but as a user I would rather have a short error message with a clear idea of what went wrong with context. Please consider my suggestion indicating: the source used, the admissible temperature and the custom indication

radis/api/kuruczapi.py
code29563: I changed the dtype for the three string columns from 'string' to 'object' due to an error appearing in [Python 3.11 tests](https://app.travis-ci.com/github/radis/radis/jobs/624031722) `AttributeError: 'StringDtype' object has no attribute 'itemsize'`

code29563: Seems to be related to [this](https://github.com/pandas-dev/pandas/issues/31199)

radis/lbl/factory.py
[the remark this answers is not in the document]
code29563: To be more preciese, I haven't encountered a Kurucz linelist that specifies any isotope in the respective column, rather the column value is just 0, and indeed the default for atoms when that's the case is to just use 'the standard atomic weight from the `periodictable` module'

erwanp:
@code29563 @minouHub this PR introduced a "[_transmittance_noslitm1](https://github.com/radis/radis/pull/652/files#diff-dc80cb797f3d45572823e31d687c72556636885bc28081dc601dcf19c2efd59fR886)" quantity in the Spectrum objects (equal to transmittance-1) ; it seems like an unecessary addition, taking space. 

Do you know why it's there? 

Else let's clean it

code29563:
> @code29563 @minouHub this PR introduced a "[_transmittance_noslitm1](https://github.com/radis/radis/pull/652/files#diff-dc80cb797f3d45572823e31d687c72556636885bc28081dc601dcf19c2efd59fR886)" quantity in the Spectrum objects (equal to transmittance-1) ; it seems like an unecessary addition, taking space.
> 
> Do you know why it's there?
> 
> Else let's clean it

`expm1` is used to handle the same issues of precision when `absorbance` is small alluded to [here](https://github.com/radis/radis/blob/e14a83053cfb20dab4090921f0de13240ff156f0/radis/lbl/factory.py#L922) when calculating [`radiance_noslit`](https://github.com/radis/radis/pull/652/files#diff-dc80cb797f3d45572823e31d687c72556636885bc28081dc601dcf19c2efd59fR1558) and [`emissivity_noslit`](https://github.com/radis/radis/pull/652/files#diff-dc80cb797f3d45572823e31d687c72556636885bc28081dc601dcf19c2efd59fR1738)

PR #532, opened 2022-11-07
0.14

anandxkumar:
## Description
The major release `0.14` after `0.13`, brings essential fixes to the interface, and the final implementations of the Google Summer of Code 2022 new features with gallery examples.

Main changes : 
— non air mixtures automatically taken into account when using Hitran #495 
- fitting modules examples #520 
- Radis now uses a common API with Exojax to fetch databases 

## What's Changed
Features:
* Implement all new fitting modules and accompanied gallery examples into RADIS by @TranHuuNhatHuy in https://github.com/radis/radis/pull/520
* add explicit warning/error message if line with 0-cm-1 broadening by @erwanp in https://github.com/radis/radis/pull/524
* Allow astropy units to be used in crop() by @TranHuuNhatHuy in https://github.com/radis/radis/pull/526
* Removing completed TODO in radis/lbl/calc.py by @mdrodelo in https://github.com/radis/radis/pull/527
* Add/common api by @erwanp in https://github.com/radis/radis/pull/480
* Non air diluent 2 by @Supriya1702 in https://github.com/radis/radis/pull/495
* number of trans files in H3O_p/Exomol error manually fixed by @HajimeKawahara in https://github.com/radis/radis/pull/531

Docs:
* Additional illustrative gallery examples for new fitting module by @TranHuuNhatHuy in https://github.com/radis/radis/pull/522
Screenshot from 2022-11-08 03-18-34
* Add some small docs/examples by @TranHuuNhatHuy in https://github.com/radis/radis/pull/518

## New Contributors
* @mdrodelo made their first contribution in https://github.com/radis/radis/pull/527
* @HajimeKawahara made their first contribution in https://github.com/radis/radis/pull/531

**Full Changelog**: https://github.com/radis/radis/compare/0.13.1...0.14

PR #656, opened 2024-06-17
remove vaex hard dependency

### LINKED ISSUE #653: Support for Python 3.11
### 🐛 Describe the bug

Dear all,
I hope this description will be clear, and that I am not raising a pointless issue.
I am available for any questions,
Sincerely

### Problem description
I'm currently unable to install `radis` in my conda environment (miniconda distribution) when using Python 3.11 . 

### Issue reproduction
I create a Conda environment in Python 3.11
```bash
$ conda create -n radis_env python==3.11
$ conda activate radis_env 
```
The installation of radis is failing both under pip
```bash
$ pip install radis

      src/strings.cpp:43:49:   required from here
      vendor/boost/boost/concept/usage.hpp:16:48: warning: ‘this’ pointer is null [-Wnonnull]
         16 |     ~usage_requirements() { ((Model*)0)->~Model(); }
            |                             ~~~~~~~~~~~~~~~~~~~^~
      vendor/boost/boost/concept/usage.hpp:30:7: note: in a call to non-static member function ‘boost::SinglePassRangeConcept<T>::~SinglePassRangeConcept() [with T = std::__cxx11::basic_string<char>]’
         30 |       ~model()
            |       ^
      vendor/boost/boost/range/concepts.hpp:284:9: note: in expansion of macro ‘BOOST_CONCEPT_USAGE’
        284 |         BOOST_CONCEPT_USAGE(SinglePassRangeConcept)
            |         ^~~~~~~~~~~~~~~~~~~
      error: command '/usr/bin/gcc' failed with exit code 1
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for vaex-core
Failed to build vaex-core
ERROR: Could not build wheels for vaex-core, which is required to install pyproject.toml-based projects
```
and conda
```bash
$ conda install radis -c conda-forge

Could not solve for environment specs
The following packages are incompatible
├─ pin-1 is installable and it requires
│  └─ python 3.11.* , which can be installed;
└─ radis is not installable because there are no viable options
   ├─ radis 0.14 would require
   │  └─ python >=3.10,<3.11.0a0 , which conflicts with any installable versions previously reported;
   ├─ radis 0.14 would require
   │  └─ python >=3.8,<3.9.0a0 , which conflicts with any installable versions previously reported;
   └─ radis 0.14 would require
      └─ python >=3.9,<3.10.0a0 , which conflicts with any installable versions previously reported.
```

If I repeate the procedure with Python 3.10

```bash
$ conda create -n radis_env python==3.10
$ conda activate radis_env 
$ pip install radis
```
The package is correctly installed.

Additionally, I'm also able to install the package via the file `environment.yml`, following the  [procedure described in the develeoper's guide](https://radis.readthedocs.io/en/latest/dev/developer.html)

### Conclusion
The question is then if this issue is related to my personal set-up (in which case I would be sorry for the bother!) or `radis` just does not support python 3.11 . Is this the first time  an issue related to Python 3.11 is raised?
If it turns out that Python 3.11 is not supported today, I would love to know if it's planned to upgrade to newer Python versions.

### 💡 Possible solutions

From the conda install error message it looks that radis does not support python above 3.10

```bash 
Could not solve for environment specs
The following packages are incompatible
├─ pin-1 is installable and it requires
│  └─ python 3.11.* , which can be installed;
└─ radis is not installable because there are no viable options
   ├─ radis 0.14 would require
   │  └─ python >=3.10,<3.11.0a0 , which conflicts with any installable versions previously reported;
   ├─ radis 0.14 would require
   │  └─ python >=3.8,<3.9.0a0 , which conflicts with any installable versions previously reported;
   └─ radis 0.14 would require
      └─ python >=3.9,<3.10.0a0 , which conflicts with any installable versions previously reported.
 ```
 
 I guess the problem comes from the [ `vaex-core` ](https://github.com/vaexio/vaex/tree/core-v4.17.1) package, which supports Python up to 3.10.

### 🎲 Radis version

0.14

### 💻 Operating system

Linux

### PR DESCRIPTION
**Pull request by erwanp**
### Description

Fixes #653

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -28,9 +28,9 @@ dependencies:
 - seaborn   # other matplotlib themes
 - termcolor  # terminal colors
 - specutils
-- vaex-core>=4.17.0   # load HDF5 files  (version needed to fix https://github.com/radis/radis/issues/598)
-- vaex-hdf5
-- vaex-viz
+#- vaex-core>=4.17.0   # load HDF5 files  (version needed to fix https://github.com/radis/radis/issues/598)
+#- vaex-hdf5
+#- vaex-viz
 - pip
 - pip:
   - -r requirements.txt              # add all packages directly in requirements.txt
--- a/radis/api/dbmanager.py
+++ b/radis/api/dbmanager.py
@@ -8,9 +8,15 @@ from io import BytesIO
 from os.path import abspath, dirname, exists, expanduser, join, split, splitext
 from zipfile import ZipFile
 
-from radis.misc.config import addDatabankEntries, getDatabankEntries, getDatabankList
-from radis.misc.printer import printr
-from radis.misc.warning import DatabaseAlreadyExists, DeprecatedFileWarning
+from ..misc.config import addDatabankEntries, getDatabankEntries, getDatabankList
+from ..misc.printer import printr
+from ..misc.utils import NotInstalled, not_installed_vaex_args
+from ..misc.warning import DatabaseAlreadyExists, DeprecatedFileWarning
+
+try:
+    import vaex
+except ImportError:
+    vaex = NotInstalled(*not_installed_vaex_args)
 
 try:
     from .cache_files import check_not_deprecated
@@ -33,6 +39,17 @@ from numpy import DataSource
 LAST_VALID_DATE = (
     "01 Jan 2010"  # set to a later date to force re-download of all databases
 )
+
+
+def get_auto_MEMORY_MAPPING_ENGINE():
+    """see https://github.com/radis/radis/issues/653"""
+    import sys
+
+    if sys.version_info[2] <= 11:
+        return "vaex"
+    else:
+        return "pytables"
+
 
 # Add a zip opener to the datasource _file_openers
 def open_zip(zipname, mode="r", encoding=None, newline=None):
@@ -103,7 +120,7 @@ class DatabaseManager(object):
 
             engine = config["MEMORY_MAPPING_ENGINE"]  # 'pytables', 'vaex', 'feather'
             if engine == "auto":
-                engine = "vaex"
+                engine = get_auto_MEMORY_MAPPING_ENGINE()
 
         self.name = name
         self.molecule = molecule
@@ -490,8 +507,6 @@ class DatabaseManager(object):
                 nrows = store.get_storer("df").nrows
 
         elif engine == "vaex":
-            import vaex
-
             # by default vaex does not load everything
             df = vaex.open(local_file)
             nrows = len(df)
@@ -512,8 +527,6 @@ class DatabaseManager(object):
         from radis.misc.basics import is_number
 
         if is_number(self.alpha_ref):
-            import vaex
-
             if isinstance(df, vaex.dataframe.DataFrameLocal):
                 # see https://github.com/vaexio/vaex/pull/1570
                 df[key] = vaex.vconstant(float(value), length=df.length_unfiltered())
@@ -527,9 +540,9 @@ class DatabaseManager(object):
 
         mdb.rename_columns(df, {"nu_lines":"wav"})
         """
-        import vaex
-
-        if isinstance(df, vaex.dataframe.DataFrameLocal):
+        if not isinstance(vaex, NotInstalled) and isinstance(
+            df, vaex.dataframe.DataFrameLocal
+        ):
             for k, v in rename_dict.items():
                 df.rename(k, v)
         elif isinstance(df, pd.DataFrame):
--- a/radis/api/exomolapi.py
+++ b/radis/api/exomolapi.py
@@ -13,7 +13,7 @@ import warnings
 
 import numpy as np
 
-from radis.api.dbmanager import DatabaseManager
+from radis.api.dbmanager import DatabaseManager, get_auto_MEMORY_MAPPING_ENGINE
 from radis.db.classes import EXOMOL_MOLECULES, EXOMOL_ONLY_ISOTOPES_NAMES
 
 EXOMOL_URL = "http://www.exomol.com/db/"
@@ -1009,7 +1009,7 @@ class MdbExomol(DatabaseManager):
 
             engine = config["MEMORY_MAPPING_ENGINE"]
             if engine == "auto":
-                engine = "vaex"
+                engine = get_auto_MEMORY_MAPPING_ENGINE()
 
         self.path = pathlib.Path(path)
         if local_databases is not None:
--- a/radis/api/hdf5.py
+++ b/radis/api/hdf5.py
@@ -13,6 +13,13 @@ import h5py
 import pandas as pd
 from tables.exceptions import NoSuchNodeError
 
+from ..misc.utils import NotInstalled, not_installed_vaex_args
+
+try:
+    import vaex
+except ImportError:
+    vaex = NotInstalled(*not_installed_vaex_args)
+
 
 def vaexsafe_colname(name):
     """replace '/' (forbidden in HDF5 vaex column names with '_'
@@ -25,7 +32,6 @@ def vaexsafe_colname(name):
 def update_pytables_to_vaex(fname, remove_initial=False, verbose=True, key="df"):
     """Convert a HDF5 file generated from PyTables to a
     Vaex-friendly HDF5 format, preserving metadata"""
-    import vaex
 
     if fname.endswith(".h5"):
         fname_vaex = fname.replace(".h5", ".hdf5")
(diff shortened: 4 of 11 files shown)

### CODE REVIEW

**Comment by erwanp** [comment]
Some tests are failing in `test\io\test_exomol\test_calc_exomol_vs_hitemp()` 

image

There is clearly a problem here; and it is not related to broadening : the areas are not the same

- either there was a problem introduced in our calculations recently?
- if they two spectra use exactly the same calculation pathway in RAdis, then it's a problem in one of the two databases

@minouHub do you want to have a look ?

**Comment by dcmvdbekerom** [comment]
Related to https://github.com/radis/radis/issues/463?

**Comment by erwanp** [comment]
@dcmvdbekerom you're right it's very related; and also we now need different environments (with/without) vaex depending on python version. Conda is not designed for optional dependencies  (https://github.com/conda/conda/issues/12098) 

@minouHub all tests are now passing except from the test above, which is physics-related https://github.com/radis/radis/pull/656#issuecomment-2175398355

**Comment by ddelange** [comment]
fwiw, [polars](https://pypi.org/project/polars/) (zero copy rust bindings) is the fastest python dataframe library (believing the benchmarks) for many operations, and has a [lazy api](https://docs.pola.rs/user-guide/lazy/)

**Comment by erwanp** [comment]
Thanks @ddelange for the suggestion. We're using Vaex for IO operations (read/write HDF5 files), and more recently for some internal calculations. It seems Polar does not support HDF5 yet (https://github.com/pola-rs/polars/issues/3520) ; hence the move won't be straightforward. 
We're not married to HDF5 and Radis is architectured to handle other formats ; do you have a file format in mind that works well with Polars ?

**Comment by minouHub** [comment]
Something was messed up in the mapping of pytables. 

- [ ] Add a test to check that both `memory_mapping_engine="vaex"` and `memory_mapping_engine="pytables"` produce the same outputs

File: examples/0_Database_handling/plot_explore_database_vaex.py [hunk 1]
```diff
@@ -16,9 +16,17 @@
 
 """
 
+try:
```
erwanp, line 19: @minouHub this was my alternative (to keep the example the same)

```
#%% Discard this example if Vaex is not installed
try:
    import vaex
except ImportError:
    quit()

# ----
```

minouHub:
@erwanp I could not reproduce the failing test of Travis in `test_eq_spectrum_emulated_gpu`.

erwanp:
(last test seemed good and failed for connection reasons, I restarted the build from Travis)

minouHub:
- [ ] Still a problem with `plot_full_range.py`. The HITRAN H2O database is not parsed correctly and the CO2 broadening columns are Nan 
- [ ] I cannot reproduce the error with `test_eq_spectrum_emulated_gpu()`

@erwanp is you have a hint, you would help me a lot

examples/0_Database_handling/plot_explore_database_vaex.py
erwanp: @minouHub it doesn't work ?

minouHub: no, the example in the doc was just stopping at that point. I guess it's because there was no plot generated for the documentation image. Anyway, it's working now for a pandas dataframe

minouHub:
Somehow, the last commit sorted the gpu problem... A bit scared about that but let's move on. Next step is to merge #657 and rebase here

minouHub:
(pushed on the wrong branch)

minouHub:
@erwanp, seems like all objectives are attained. Let me know if anything else is needed but I think that the compatibility problem is solved. I let you merge if you think it's good. We should also consider merging 0.15 to the master!

.travis.yml
```diff
@@ -1,130 +1,166 @@
+# Runs Test scripts; Linting; Deployment on Travis-CI
 language: python
 python:
-  # We don't actually use the Travis Python, but this keeps it organized.
   #- "2.7"    # not tested anymore starting from 0.9.19
-  - "3.8"
+  - "3.8" #end of life, Novembre 2024, https://devguide.python.org/versions/
+  - "3.10"
+  - "3.11"
```
erwanp, line 7: @minouHub removing 3.8 to save some calculation time

```suggestion
  - "3.10"  # has vaex support
  - "3.11"  # has no vaex support (June 24)
```

.travis.yml
```diff
@@ -1,130 +1,166 @@
...
+      install: skip
+        #- python3 -m pip install cibuildwheel==2.19.1 #If you intend to build a pure-Python wheel, you don't need cibuildwheel
+
+      before_script: skip #do NOT run the before_script of "tests" stage
+
+      script:
+        # build the wheels, put them into './dist'
+        # - python3 -m cibuildwheel --output-dir dist #If you intend to build a pure-Python wheel, you don't need cibuildwheel
+        - python3 -m pip wheel -w dist .
+
```
erwanp, line 158: @minouHub I don't think we need this (nor the "skip_cleanup: true") step since we're not deploying cibuildwheels.  Pure Python wheels are already built and deployed by the "deploy" stage ([distributions: "sdist bdist_wheel"](https://github.com/radis/radis/pull/656/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485R176))

And unless I'm wrong, Pure Python wheels are the same for every OS (Linux/Windows)

minouHub: @erwanp, I just applied what is recommended by Travis https://docs.travis-ci.com/user/job-lifecycle/#deploying-your-code
To be honest, I'm completely new to this and simply applying what I read

> And unless I'm wrong, Pure Python wheels are the same for every OS (Linux/Windows)

You're right, but I don't think that this is comment is related to `skip_cleanup: true`

.travis.yml
minouHub: ok, will be applied the next commit

minouHub:
For the moment in this PR, `vaex` must be installed by the user depending if his python version is <3.11 or >=3.11. This should be handled automatically with environment markers, see https://pip.pypa.io/en/stable/reference/requirement-specifiers/#examples

minouHub:
Thanks to [5814fa5](https://github.com/radis/radis/pull/656/commits/5814fa5fe93044c4f1070425a1bf6281f72b7b1f):

- the correct python version is tested
- the tests are super fast now! only 7-9 min. I honestly don't understand why but happy days

.travis.yml
erwanp: For deployment; I think it is handled by : 

https://docs.travis-ci.com/user/deployment/pypi/#uploading-different-distributions

```
deploy:
  provider: pypi
  username: ...
  password: ...
  distributions: "sdist bdist_wheel" # Your distributions here
# If you specify bdist_wheel in the distributions, the wheel package will automatically be installed.
```

So I assume wheel packages are automatically created by the "deploy" stage too. 
- Unfortunately the Logs of the last push to Master branch [are not available anymore](https://app.travis-ci.com/github/radis/radis/jobs/552887026) so we don't know if wheels were built there.
- But, [Radis 0.14 deployed on PyPi](https://pypi.org/project/radis/#files) has a built wheel.  

So I'm confident an extra build_stage is not needed

minouHub: I just checked the travis.yaml file pushed on Nov 8, 2022 ([here](https://github.com/radis/radis/pull/532/files#)) and the 0.14 version released on Pypi on Nov 8, 2022.

I agree with you. The Python wheels are **probably** generated automatically. I'll switch back to the original version of the deploy phase and let you merge if you're happy

minouHub:
@erwanp there are some problems a user asks for several molecules calculated at once in `calc_spectrum`. I suggest leaving that for radis=0.16, see #663

minouHub:
@erwanp merge if you're happy 🥳

erwanp:
For information @minouHub @dcmvdbekerom Vaex support for Python 3.11; 3.12 was merged https://github.com/vaexio/vaex/pull/2331

We should keep Vaex optional as is done now; but can be the default for Python up to 3.11 and 3.12 now 

Given how [get_memory_engine](https://github.com/radis/radis/blob/c1f886da15f7dee2938b5b1dc473f393b0ae1b8d/radis/api/dbmanager.py#L51) is implemented, we do not have anything to do : if vaex is installed it works, else it defaults to pandas. 

Maybe we should update the Conda recipees of Radis; however; to have Vaex as a dependency if Python<=3.12 and remove the dependency if >3.12. @arunavabasucom is that possible to do in Conda recipees?

Issue #596, opened 2023-07-31
📚 todo: Provide compiled Cython/C/C++ binaries


dcmvdbekerom:
### 💭 Description

Over the past years RADIS has been heavily optimized. Some of these optimization rely on cython/c++ code which is compiled during installation of RADIS. However, we shouldn't expect all users to have a compiler installed, and we also don't want them to miss out on optimizations. ~~The situation is even worse for GPU code, as it currently requires `cupy`, which requires the `CUDA Toolkit` (3GB), which requires `Microsoft Visual Studio` (10+GB).~~

To prevent users have to compile code, we should do it for them. ~~The pre-built libraries should be "staged", because we also don't want developers to have to rebuild all code if they're only working on one thing. For example, if a dev works on a cython optimization, we shouldn't require them to also be able to compile GPU code (requiring CUDA Toolkit and MSVC).~~

Cython/C++ code could be compiled into pre-built wheels. The problem with Wheels is that we need a separate version for OS (windows/linux/macOS), architecture (32/64bit), and Python version. However, there is a package called `cibuildwheels` that let's the CI build all these different flavors. I propose we build these whenever a version is released to PyPy.

Currently all cython functions have a python backup in case compilation fails. It would be better if we provide the backup in the form of Wheels, because then everyone will be running the same code. The disadvantage is that the barrier for development is slightly higher, because it would require more code to be re-compiled during development.

Curious to hear thoughts & suggestions!

*Edit: CUDA can be handled entirely without C/C++ compilation, so I made a separate thread here: https://github.com/radis/radis/issues/603)*

erwanp:
I'm curious about how it is done and implemented in other packages. Do you know of any gpu-accelerated Python package that doesn't require the C++ compiler & CUDA Toolkit ? Their devs could help us make the right choices in terms of deployment

dcmvdbekerom:
It seems at least for the C++ part, [`cibuildwheels` is how most packages do it](https://github.com/pypa/cibuildwheel#working-examples)
For the GPU part, most packages I know of require the CUDA toolkit (e.g., Vaex is built on top of cupy, which requires the CUDA toolkit)

erwanp:
Btw the `deploy` step of Radis (activated on pushes to the master branch) already builds a wheel https://github.com/radis/radis/blob/develop/.travis.yml#L84

erwanp:
Said today : we'll release next version 0.15 with `cibuildwheels` and see if it works ! I'm taking this.

[...]

PR #657, opened 2024-06-17
update Test environment

erwanp:
@minouHub this should make tests faster : 
- only one Test stage; i.e. test environment is built only one
- however Fast Tests are ran first, which I think was the intent behind splitting in two test stages


minouHub:
Hello. A quick reply: is the build failing as soon as the fast tests are failing. I might have missed something a few months ago but the build would run all the commands even if the fast tests failed

minouHub:
See comment here: https://github.com/travis-ci/travis-ci/issues/1066#issuecomment-570898290

erwanp:
Indeed you're right, tests continued even after failure of fast tests : https://app.travis-ci.com/github/radis/radis/jobs/623023907#L2177

erwanp:
It worked correctly and failed after 3min https://app.travis-ci.com/github/radis/radis/jobs/623146582

minouHub:
Ok, I'm convinced. It's a good idea to include fast_tests in the `before_script` section.

minouHub:
@erwanp the Travis file is now updated to test Python 3.8, 3.9, 3.10. Please comment, and thick the following if you agree.

- [ ] We can add `python=3.11` that is supposed to be problematic with `vaex` in #656 . 
- [ ] We can still add some testing on Windows. Don't know if that relevant.

Concerning `cbuildwheel`, there was an error because Radis is "a pure Python project". Using `cbuildwheel` was mentioned with @dcmvdbekerom in https://github.com/radis/radis/issues/596#issuecomment-1665532986. Since that comment `cython` is not used anymore. Thus, I just commented the line and replaced it with a pure Python wheel. I'm completely new to all that jazz so do not hesitate to change it. 
_In short:_ Do we still need `cbuildwheel`?

.travis.yml
```diff
@@ -1,130 +1,121 @@
+# Runs Test scripts; Linting; Deployment on Travis-CI
 language: python
 python:
   # We don't actually use the Travis Python, but this keeps it organized.
   #- "2.7"    # not tested anymore starting from 0.9.19
   - "3.8"
+  - "3.9"
+  - "3.10"
 os: linux
+dist: focal #needed for python 3.10 (xenial is default for Travis-CI, but does not support python 3.10)
```
erwanp, line 10: that was hard to find ! @minouHub

.travis.yml
```diff
@@ -1,130 +1,121 @@
+# Runs Test scripts; Linting; Deployment on Travis-CI
 language: python
 python:
   # We don't actually use the Travis Python, but this keeps it organized.
   #- "2.7"    # not tested anymore starting from 0.9.19
   - "3.8"
+  - "3.9"
+  - "3.10"
```
erwanp, line 8: The more Python versions the faster we burn out Credits. I'd keep 1 Vaex-version only; and 1 non-Vaex version (after #656)  @minouHub  

So for the time being , let's test on Python 3.10  and drop 3.8; 3.9 tests ?

In the future we might want to test multiple versions only for merges to Master branch , but I don't know how to add that easily in the Travis matrix.
At some point we'll also switch to GitHub Actions; so let's do this multi-python testing at that time (GSOC project?)

minouHub: On the other hand ... if we had the tests ready for python 3.11 and 3.12, we would have caught the problem with vaex earlier. I don't know if the Travis credit are really a problem, we can always ask for a renewal right? On top of that, we now have the "fast" tests stopping the build and saving a lot of credits. I'd prefer to keep several python versions tested. Let's say three at least, distributed the way we prefer.

minouHub: On that: 

- [ ] let's drop the tests on Windows for the moment
- [ ] I'll look for a way to implement more tests but only on the master branch

minouHub:
linting failed .... too bad, I'm merging anyway

PR #670, opened 2024-07-24
[bug fix] Solve failing tests on python 3.8 

minouHub:
Solve failing tests on python 3.8


minouHub:
The tests passed on Python 3.8 and 3.12, see [e8e964b](https://github.com/radis/radis/pull/670/commits/e8e964b1167883a0cc2660e7580e98542ae9eac4)
Keeping these tests only for merging on master. The tests on Python 3.10 and 3.11 should be sufficient.

PR #671, opened 2024-07-25
Remove duplicate Einstein coefficient Aul in df0

minouHub:
### Description

This pull request is to address the redundancy between the df.A and df.Aul column. After this PR, the only column for the forst Einstein coefficient will be df.A. If it is missing from the database, df.A will still be recalculated automatically.

Fixes #655


minouHub:
Interesting. As shown in [68fd595](https://github.com/radis/radis/pull/671/commits/68fd595649c4a2bce75dd6912075267e16719262), the HITRAN database seems to provide a better optically thick spectrum than HITEMP. 
Anyway, there is no more mention of df.Aul from this point.

PR #674, opened 2024-08-03
A bug related to map_add

HajimeKawahara:
### Description

Some updates to the `exomolapy.py` file were not made to accommodate the additional arguments due to changes in the `map_add` specification. This causes a serious bug when downloading specific molecules.

```python
        map_add("E", "eupper", "i_upper")
```
should be 
```python
        trans = map_add(trans, "E", "eupper", "i_upper")
```

Specifically, when the `trans` file does not include `nu_lines`, `exomolapi` is designed to calculate `nu_lines` from `eupper` and `elower`, but this is not functioning correctly. For example, H2S is such a molecule, so initialization fails. For more details, please see https://github.com/HajimeKawahara/exojax/issues/504 .

Fixes https://github.com/HajimeKawahara/exojax/issues/504

### DIFF
--- a/radis/api/exomolapi.py
+++ b/radis/api/exomolapi.py
@@ -460,7 +460,7 @@ def pickup_gE(states, trans, dic_def, skip_optional_data=True, engine="vaex"):
             return column.hasnans
 
     if not "nu_lines" in trans or has_nan(trans["nu_lines"]):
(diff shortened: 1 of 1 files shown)


HajimeKawahara:
If possible, I would appreciate it if this could be included as a patch in release 0.15.1.

PR #676, opened 2024-08-07
switch to using Einstein Coefficient for non-equilirbium linestrength calculation

code29563:
Should fix the case where some non-equilibrium atomic spectra like N_II weren't appearing. `-hc_k * df.El / Tref` is so low that [the reference linestrength is evaluated to 0](https://github.com/radis/radis/blob/430e5045e0719db891aa81c0e955b1e220518b0f/radis/lbl/base.py#L2258), so this change renders the `int` column redundant in those cases and switches to just using the already-present `A` column (through `calc_S0`),

Also removes `calc_reference_linestrength` which was originally unused, then used to calculate the reference linestrength of atomic lines, and is now redundant as they no longer use it and the issue of `Aul` vs `A` has been resolved in 430e5045e0719db891aa81c0e955b1e220518b0f.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -17,7 +17,7 @@ dependencies:
 - joblib  # for parallel loading of SpecDatabase
 - lmfit  # for new fitting modules
 - matplotlib
-- numpy
+- numpy<2.0
 - numba  # just-in-time compiler
 - pandas
 - plotly>=2.5.1  # for line survey HTML output
--- a/radis/lbl/base.py
+++ b/radis/lbl/base.py
@@ -2135,7 +2135,7 @@ class BaseFactory(DatabankLoader):
         References
         ----------
         Weighted transition-moment squared :math:`R_s^2` from linestrength :math:`S_0`
-        at temperature :math:`T_ref`, derived from Eq.(A5) in [Rothman-1998]_
+        at temperature :math:`T_{ref}`, derived from Eq.(A5) in [Rothman-1998]_
 
         .. math:
             R_s^2=10^{+36}\\frac{3h c}{8{\\pi}^3} \\frac{1}{n_u} \\frac{1}{\\frac{I_a g_l}{Q_{ref}} \\operatorname{exp}\\left(\\frac{-E_l}{T_{ref}}\\right)} \\frac{1}{1-\\operatorname{exp}\\left(\\frac{-n_u}{T_{ref}}\\right)} S_0
@@ -2310,7 +2310,8 @@ class BaseFactory(DatabankLoader):
         return
 
     def calc_S0(self, df):
-        """Calculate the unscaled intensity from the tabulated Einstein coefficient.
+        """Calculate S0 from A [s-1], the tabulated Einstein coefficient. S0 is
+        the part of the linestrength that does not depend on the temperature.
 
         Parameters
         ----------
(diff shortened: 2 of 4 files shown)


code29563:
Also removed `_get_S0` which was unused.

minouHub:
@erwanp I'd like your opinion on this.

This test is inspired from `test_all_calc_methods_CO2pcN` which is failing. I'm testing different databases.
```python
from radis.misc.config import getDatabankEntries
from radis.test.utils import (
    define_Evib_as_min_of_polyad,
    discard_lines_with_na_levels,
    setup_test_line_databases,
)

def calc_and_print(TAG):
    s_nq = sf.non_eq_spectrum(Tvib=Tgas, Trot=Tgas)
    s_eq = sf.eq_spectrum(Tgas=Tgas)

    plot_diff(s_eq, s_nq, 'absorbance')
    A_eq = s_eq.get_integral("abscoeff")
    A_nq = s_nq.get_integral("abscoeff")
    print("***")
    print(TAG)
    print(A_eq, A_nq, f"{(1-A_eq/A_nq)*100:.2f}%")

import numpy as np
Tgas = 500
sf = SpectrumFactory(
    wavenum_min=2284, #2284 or 2250
    wavenum_max=2285, 
    truncation=2.5,  
    mole_fraction=1,
    path_length=0.025,
    cutoff=1e-25,
    molecule="CO2",
    isotope='1',
    export_lines=True,
    verbose=False
)

## HITRAN
sf.fetch_databank("hitran", load_columns='noneq')
calc_and_print("hitran")

## HITEMP
sf.fetch_databank("hitemp", load_columns='noneq')
sf.df0.drop(sf.df0.index[sf.df0['v1u']==-1], inplace=True)
calc_and_print("hitemp")

## "HITEMP-CO2-HAMIL-TEST"
setup_test_line_databases()
# Generate a Levels database with p,c,N energy partitioning
# ... copy "HITEMP-CO2-HAMIL-TEST" info
database_kwargs = getDatabankEntries("HITEMP-CO2-HAMIL-TEST")
# ... adapt it to 'cdsd-pcN' mode
del database_kwargs["info"]
database_kwargs["levelsfmt"] = "cdsd-pcN"
# ... load the new database
sf.load_databank(**database_kwargs, load_energies=True, load_columns="noneq")
# Now, define Evib:
Q_calc.df = define_Evib_as_min_of_polyad(Q_calc.df, keys=["p", "c", "N"])

# With this Evib definition, clean the Lines database from where Evib is not defined
# (because Levels does not exist in the reduced, test Level Database)
discard_lines_with_na_levels(sf)
calc_and_print("HITEMP-CO2-HAMIL-TEST")
```

I got (with some warnings removed):
***
```
hitran
1.9919313756425836 1.991066979062065 -0.04%
***
hitemp
1.9836390824576413 1.9889425030924346 0.27%
***
HITEMP-CO2-HAMIL-TEST
1.7460209183591857 1.7221937816764832 -1.38%
```
Approximately the same % difference are obtain when extending the waverange down to 2285 cm-1. 
In conclusions, it seems to me that the Einstein coef. and linestrength of hitran match very well, a bit less for hitemp. I propose we relax a bit this test, or propose another test.

minouHub:
By the way ... the `test_all_calc_methods_CO2pcN` fails in the develop branch at any temperature above 500 K. So I'm guessing `rtol=0.001` was just set by convenience. Waiting for feedback of @erwanp

minouHub (commented):
I checked the calculation. Everything is good to my opinion.

Waiting a few days before merging:

- [ ] @erwanp , is it ok to increase rtol to `rtol=2e-2` in `test_all_calc_methods_CO2pcN`
- [ ] @dcmvdbekerom, I find the notations `S0` and the name of the function `calc_S0` a bit confusing. It could be mistaken with the actual linestrength at 296 K. Would you see a problem naming them `S_cst` and `calc_S_cst` instead?

minouHub:
I'd like to merge 0.15.1 this week. 

> The change in accuracy for test_eq_vs_noneq_isotope is minor, and can be resolved by setting rtol to 7.2e-3 instead of 7e-3,

@code29563 you can increase the tolerance for this one. 

>  @dcmvdbekerom, I find the notations S0 and the name of the function calc_S0 a bit confusing. It could be mistaken with the actual linestrength at 296 K. Would you see a problem naming them S_cst and calc_S_cst instead?

If no news from @dcmvdbekerom concerning this, we will leave it for future work.

Issue #682, opened 2024-08-12
abscoeff2xsection causes crash when using MergeSlabs


Airfooox:
### 🐛 Describe the bug

First of all the error traceback for this issue:
```
Traceback (most recent call last):
  File "C:\Users\username\AppData\Local\anaconda3\envs\Radis-Simulator\Lib\threading.py", line 1073, in _bootstrap_inner
    self.run()
  File "C:\Users\username\AppData\Local\anaconda3\envs\Radis-Simulator\Lib\threading.py", line 1010, in run
    self._target(*self._args, **self._kwargs)
  File "D:\_dev\Radis Simulator\src\main.py", line 777, in startSeriesSimulationTask
    gasSpectraDict = Root.combineCalculatedSpectraIntoGasSpectrum(
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\_dev\Radis Simulator\src\main.py", line 1038, in combineCalculatedSpectraIntoGasSpectrum
    hotGasMerge = MergeSlabs(
                  ^^^^^^^^^^^
  File "D:\_dev\Radis Simulator\radis-0.15\radis\los\slabs.py", line 830, in MergeSlabs
    s.update(
  File "D:\_dev\Radis Simulator\radis-0.15\radis\spectrum\spectrum.py", line 1626, in update
    return update(
           ^^^^^^^
  File "D:\_dev\Radis Simulator\radis-0.15\radis\spectrum\rescale.py", line 614, in update
    _recalculate(
  File "D:\_dev\Radis Simulator\radis-0.15\radis\spectrum\rescale.py", line 2163, in _recalculate
    rescaled, rescaled_units = rescale_xsection(
                               ^^^^^^^^^^^^^^^^^
  File "D:\_dev\Radis Simulator\radis-0.15\radis\spectrum\rescale.py", line 1816, in rescale_xsection
    xsection = abscoeff2xsection(
               ^^^^^^^^^^^^^^^^^^
  File "D:\_dev\Radis Simulator\radis-0.15\radis\spectrum\equations.py", line 62, in abscoeff2xsection
    return abscoeff_cm1 * (k_b * Tgas_K / mole_fraction / pressure_Pa) * 1e6  # cm2
                           ~~~~~~~~~~~~~^~~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for /: 'float' and 'str'
```

[...]

Issue #681, opened 2024-08-12
Missing gpu files in pypi 


Airfooox:
### 🐛 Describe the bug

Files for the GPU acceleration functionality are missing in the pypi release archive (https://pypi.org/project/radis/#files)
compared to the archive obtained in GitHub https://github.com/radis/radis/releases/tag/0.15

Namely the missing files I found were in the radis/gpu directory: 
gpu.py, params.py, structs.py and the README. 
The cuda directory in that same gpu dir is missing a few files aswell: 
build_kernals.bat, driver.py, emulate.py

### 💡 Possible solutions

Current workaround is of course to download radis 0.15 from GitHub and implement it locally.

Im not familiar with the release procedure or how it it done generally, so I cant suggest any changes. But I think it would be good to check all files from the pypi release archive against the correct release from GitHub, maybe some other files were not included correctly.

### 🎲 Radis version

0.15

### 💻 Operating system

Windows

minouHub:
Hello
Thank you so much for reporting this major issue!
Indeed, running `python -m build` outputs the gpu files, **but in the wheel it gives**:
```
[... a lot of stuff ...]
C:\Users\Nicolas Minesi\AppData\Local\Temp\build-env-br19d3_h\lib\site-packages\setuptools\command\build_py.py:215: _Warning: Package 'radis.gpu' is absent from the `packages` configuration.
!!

        ********************************************************************************
        ############################
        # Package would be ignored #
        ############################
        Python recognizes 'radis.gpu' as an importable package[^1],
        but it is absent from setuptools' `packages` configuration.

        This leads to an ambiguous overall configuration. If you want to distribute this
        package, please make sure that 'radis.gpu' is explicitly added
        to the `packages` configuration field.

        Alternatively, you can also rely on setuptools' discovery methods
        (for example by using `find_namespace_packages(...)`/`find_namespace:`
        instead of `find_packages(...)`/`find:`).

        You can read more about "package discovery" on setuptools documentation page:

        - https://setuptools.pypa.io/en/latest/userguide/package_discovery.html

        If you don't want 'radis.gpu' to be distributed and are
        already explicitly excluding 'radis.gpu' via
        `find_namespace_packages(...)/find_namespace` or `find_packages(...)/find`,
        you can try to use `exclude_package_data`, or `include-package-data=False` in
        combination with a more fine grained `package-data` configuration.

        You can read more about "package data files" on setuptools documentation page:

[...]

PR #683, opened 2024-08-12
Missing GPU files in deploy and xsection rescaling

minouHub:
### Description - Bug 1

This pull request is to address the missing files in the deployment of `radis 0.15`. The file setup.py is now simplified and do not exclude the gpu files, see #681.

Everything seems fine in the **tar.gz section**. However, here is an example of the output when running `python -m build`  in the **wheel section**:
```
[... a lot of stuff ...]
C:\Users\Nicolas Minesi\AppData\Local\Temp\build-env-br19d3_h\lib\site-packages\setuptools\command\build_py.py:215: _Warning: Package 'radis.gpu' is absent from the `packages` configuration.
!!

        ********************************************************************************
        ############################
        # Package would be ignored #
        ############################
        Python recognizes 'radis.gpu' as an importable package[^1],
        but it is absent from setuptools' `packages` configuration.

        This leads to an ambiguous overall configuration. If you want to distribute this
        package, please make sure that 'radis.gpu' is explicitly added
        to the `packages` configuration field.

        Alternatively, you can also rely on setuptools' discovery methods
        (for example by using `find_namespace_packages(...)`/`find_namespace:`
        instead of `find_packages(...)`/`find:`).

        You can read more about "package discovery" on setuptools documentation page:

        - https://setuptools.pypa.io/en/latest/userguide/package_discovery.html

        If you don't want 'radis.gpu' to be distributed and are
        already explicitly excluding 'radis.gpu' via
        `find_namespace_packages(...)/find_namespace` or `find_packages(...)/find`,
        you can try to use `exclude_package_data`, or `include-package-data=False` in
        combination with a more fine grained `package-data` configuration.

        You can read more about "package data files" on setuptools documentation page:

        - https://setuptools.pypa.io/en/latest/userguide/datafiles.html


        [^1]: For Python, any directory (with suitable naming) can be imported,
              even if it does not contain any `.py` files.
              On the other hand, currently there is no concept of package data
              directory, all directories are treated like packages.
        ********************************************************************************

!!
  check.warn(importable)
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -165,5 +165,5 @@ jobs:
       - VERSION_RADIS=$(cat radis/__version__.txt)
       - pip install radis==$VERSION_RADIS
       - pip install pytest
-      - xvfb-run -a pytest -m "not fast and not needs_cuda and not download_large_databases and not needs_db_CDSD_HITEMP and not needs_db_CDSD_HITEMP_PCN and not needs_db_CDSD_HITEMP_PC and not needs_db_HITEMP_CO2_DUNHAM and not needs_db_HITEMP_CO_DUNHAM" --durations=10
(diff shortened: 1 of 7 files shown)


minouHub:
### Description - Bug 2

Fix #682

HajimeKawahara:
The change in `exomolapi.py` (removing `check_bdat`)  by this PR look to make the bug https://github.com/HajimeKawahara/exojax/issues/519

Issue #597, opened 2023-08-01
📚 API : make sure the API remains Exojax compatible


erwanp:
### 💭 Description

Since #480 Radis embeds a common API shared by Radis-Line-by-line  (spectrum calculation code) and [Exojax](https://secondearths.sakura.ne.jp/exojax/)

We should add tests in the Radis CI to make sure the API always remains fully compatible with Exojax functions (in terms of signature, i.e. input parmaters ). 

I see it as : 

- in the develop branch, add classes representatives of how Exojax uses the API ; and make sure it works 
- before each new merge to the Master branch, add in our Travis-CI test suite a full installation & test of Exojax  

@HajimeKawahara @YuiKasagi any idea / request on this ?

HajimeKawahara:
This is a bit difficult problem but important for long time development.  Writing some test code for api.py in ExoJAX, and importing these tests to radis? In general, writing the unit tests for database I/O is a bit challenging because it sometimes needs to be downloaded from the database websites.

erwanp:
Yes. Do you already have some unitests in Exojax/api? I can borrow them and add them in Radis.api to ensure backward compatibility

HajimeKawahara:
back to this issue. See [exojax #462](https://github.com/HajimeKawahara/exojax/pull/462)
Maybe you are right. We need to make the unittest of ExoJAX/api in exojax?

erwanp:
We do 2 things : 

- for some test cases, we are downloading small databases from Exojax (SiO for instance), which requires ExoMol to be online; but limits the data transfer
- for other test cases, we have mock-data available in https://github.com/radis/radis/tree/develop/radis/test/files  and that can be fetched from anywhere with [radis.test.utils.getTestFile](https://radis.readthedocs.io/en/latest/source/radis.test.utils.html#radis.test.utils.getTestFile)

```python
from radis.test.utils import getTestFile
from radis import load_spec
load_spec(getTestFile('CO_Tgas1500K_mole_fraction0.01.spec'))
```

These are very small files, shipped with Radis, usually used to ensure I/O compatibility

HajimeKawahara:
Thanks. It looks there is no ExoMol fragment file available in [radis.test.utils.getTestFile](https://radis.readthedocs.io/en/latest/source/radis.test.utils.html#radis.test.utils.getTestFile). Then, I will use a fragment ExoMol file in exojax.

minouHub:
[...]

PR #691, opened 2024-08-31
Exomolapi broadening bug

### LINKED ISSUE #690: bug related to exomolapi (broadening)
### 🐛 Describe the bug

The change in `exomolapi.py` (removing `check_bdat`)  by #683 look to make the bug https://github.com/HajimeKawahara/exojax/issues/519

in the case of `codelv = ["a0", "a1"]` (for POKAZATEL case) it does not define the broadening parameters
otherwise it uses the default parameter even if codelv = ["a0"] or ["a1"].  
 

### 💡 Possible solutions

@minouHub @erwanp 

I plan to fix this bug, but I have a suggestion.
Ideally, we should have a comprehensive set of automated tests for the API, as suggested in #597 and https://github.com/HajimeKawahara/exojax/issues/408 , but writing automated tests that accommodate the various patterns corresponding to the database is very costly (sorry for the slow progression and I believe that the particular pattern we encountered this time couldn't have been prevented by automated testing because I couldn't image such error). As a provisional measure, how about making it mandatory for someone from the `ExoJAX` team (mainly myself) to review any changes to the `exomolapi.py` and `hitranapi.py` modules? This way, we can prevent bugs like the one we encountered this time and the one in #674 from happening in the future. Of course, this proposal is not intended to hinder code refactoring (thanks for refactoring!), and I would be more than happy to do the reviews! 

### 🎲 Radis version

0.15.1

### 💻 Operating system

Linux

HajimeKawahara:
### Description

This pull request is to address #690 and https://github.com/HajimeKawahara/exojax/issues/519
i.e. the bug for the multiple case of the broadening code level, `np.unique(bdat["code"]) = ["a0", "a1"]`.
In this case, the code level should be `codelv = "a1"`.

Also, this PR started to have the unittest for the common API (#597 and https://github.com/HajimeKawahara/exojax/issues/408 ,) .

Fixes #690

### DIFF
--- a/radis/api/exomolapi.py
+++ b/radis/api/exomolapi.py
@@ -550,27 +550,29 @@ def read_broad(broadf):
     return bdat
 
 
-# def check_bdat(bdat):
-#     """checking codes in .broad
-#     Args:
-#        bdat: exomol .broad data given by exomolapi.read_broad
-#     Returns:
-#        code level: None, a0, a1, other codes unavailable currently,
-#     """
+def check_bdat(bdat):
+    """checking codes in .broad
+    Args:
+        bdat: exomol .broad data given by exomolapi.read_broad
+    
+    Returns:
+        code level: None, a0, a1, other codes unavailable currently,
+        if a0 and a1 are available, a1 is returned.
+    """
 
-#     def checkcode(code):
-#         cmask = bdat["code"] == code
-#         if len(bdat["code"][cmask]) > 0:
-#             return True
-#         else:
-#             return False
+    def checkcode(code):
+        cmask = bdat["code"] == code
+        if len(bdat["code"][cmask]) > 0:
+            return True
+        else:
+            return False
 
-#     codelv = None
-#     for code in ["a0", "a1"]:
-#         if checkcode(code):
-#             codelv = code
+    codelv = None
+    for code in ["a0", "a1"]:
+        if checkcode(code):
+            codelv = code
 
-#     return codelv
+    return codelv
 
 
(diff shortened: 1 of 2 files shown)


minouHub:
Hello @HajimeKawahara ,

First of all, I was the one who commented this function. Sorry if I caused issues in ExoJax! 

> i.e. the bug for the multiple case of the broadening code level, np.unique(bdat["code"]) = ["a0", "a1"].

Physics: is it not an error in the ExoMol side if there are two broadening codes? If this is an error, we should contact them, they usually correct their mistakes quickly. Otherwise, I'm fine with your implementation.

> Also, this PR started to have the unittest for the common API (https://github.com/radis/radis/issues/597 and https://github.com/HajimeKawahara/exojax/issues/408 ,) .

Great!

minouHub:
Alrigth ... so I read a little bit more about these coefficients and it's not a bug :)
You still have some linting to do but I'm happy with the changes

HajimeKawahara:
Hello @minouHub Thank you. This confusion might be due to an implementation problem. When both `a0` and `a1` are included, in exomolapi.py, it is represented as `codelv=a1`. 

I believe I've fixed as the linting issues as possible, but Travis CI is still not turning green. Could this be a problem with other parts of the code?

minouHub:
Hello,
It's true that there are few problems I need to sort with Travis, but in [f8e6095](https://github.com/radis/radis/pull/691/commits/f8e60951c015fcdb2cd5dae2a9cfc06194f8086f), the problem was still the linting :)

PR #693, opened 2024-09-07
Bug fix for downloading the HDO exomol database

HajimeKawahara:
### Description

This pull request is to address the bug fix for downloading the HDO exomol. 
The bug that Exomol's HDO database cannot be read was raised as an issue in ExoJAX. This is due to two reasons.

- 1. Name parsing. The simple name should have been `H2O`, but due to a parsing issue, it was incorrectly recognized as `HHO`.
- 2. An error caused by the non-uniform wavenumber range in the Transfile. The original code assumed the wavenumber range arrays to be uniform, such as 0-500, 500-1000, 1000-1500. However, HDO/VTT does not follow this assumption. As a temporary measure, I manually provided the array as an exception within the code. A better implementation should be considered in the future (e.g., automatically retrieving the data from the ExoMol website).

Also, I added a new unit test for the common API (https://github.com/radis/radis/issues/597 and https://github.com/HajimeKawahara/exojax/issues/408 ,)

### name change

Since the name "e2s" does not allow us to infer its function (this was a name I assigned in the past, my apologies), I have renamed it to something more descriptive: `exact_molname_exomol_to_simple_molname`

Fixes https://github.com/HajimeKawahara/exojax/issues/528

### DIFF
--- a/radis/api/exomolapi.py
+++ b/radis/api/exomolapi.py
@@ -98,7 +98,7 @@ def read_def(deff):
         tag for wavelength ranges.
 
     Note:
-       For some molecules, ExoMol provides multiple trans files. numinf and numtag are the ranges and identifiers for the multiple trans files.
+        For some molecules, ExoMol provides multiple trans files. numinf and numtag are the ranges and identifiers for the multiple trans files.
 
     """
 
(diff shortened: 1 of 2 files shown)


erwanp:
Good fix.

I fix the linting tests with 
```
pre-commit run --all-files
``` 

Merging

erwanp (approved):

PR #705, opened 2024-11-28
Fix bug in NO2 molparams 

### LINKED ISSUE #703: NO2 isotope bug
### 🐛 Describe the bug

**There is an error in the handling of the isotopes of NO2.** 
**Error 1: In the molparam.txt file:**
The data format should be:
```
id iso isoname abundance     Q_296K      gj    molar_mass
#   H2O (1)
1  1  161  .997317E+00    1.7464E+02    1     18.010565
1  2  181  1.99983E-03    1.7511E+02    1     20.014811
1  3  171  3.71884E-04    1.0479E+03    6     19.014780
1  4  162  3.10693E-04    8.5901E+02    6     19.016740
```

But for NO2 is written as:
```
#   NO2 (10)
10 1  646  .991616E+00    1.3578E+04    3     45.992904
10 1  656  3.64564E-03    9.2929E+03    2     46.989938
```

Note the second row of the second column for NO2, the isotope should be 2 rather than 1. 

**Error 2: In the molparam.py file:**

Only one of the isotopes of NO2 is present, error is on line 62:
```
    (10, 1): "(14N)(16O)2",
```
the correct version should be: 
```
    (10, 1): "(14N)(16O)2",
    (10, 2): "(15N)(16O)2",
```
Finally, after implementing these changes this result in a further error on line 357 of molparam.py:

```
        else:
            out = self.df.loc[(M, isotope), key]
            assert len(out) == 1
            return out.iloc[0]
```
The assert statement does not work anymore as self.df.loc always returns a float64. In keeping with the original intention, on my local hard drive I changed it to:
```
else:
            out = self.df.loc[(M, isotope), key]
            assert isinstance(out, (float, int, np.float64))
            return out
```

### 💡 Possible solutions

As described above

### 🎲 Radis version

0.15.2

### 💻 Operating system

MacOS

alexarntzen:
### Description

This pull request is to address issue #703. It fixes the wrong indexing in `molparam.txt` and the missing NO2 isotope in `molparam.py`.  The assert statement is also updated since `.loc` does not return a `pandas.Series` anymore. 
You can verify the changes here:[https://hitran.org/docs/iso-meta/](https://hitran.org/docs/iso-meta/) and here: [https://hitran.org/docs/iso-meta/](https://hitran.org/docs/iso-meta/).

Also: 
`Cantera` 3.0.1 from conda-forge is incorrectly configured as of the 29th of November. It will not install correctly when resolved with `h5py`. Therefore I added `,!=3.0.1` to `cantera` in `environment.yml`.

Fixes #703


alexarntzen:
The reason for CI is unrelated to this PR:
> ImportError: Cantera is needed to calculate equilibrium mole fractions. Install with  `pip install cantera` or (better) `conda install -c cantera cantera`

alexarntzen:
Yes a PR with no changes fails the CI: pull request #706

erwanp:
Thanks for this fix, looks good to me, and tests pass : merging

Issue #158, opened 2020-12-21
glitch when plotting wavelengths at 132 nm ( air_index_dispersion )


erwanp:
### 🐛 Describe the bug

There is a glitch at 132 nm when plotting a Planck function, likely due to the air_index_dispersion

### 🖥 Steps To Reproduce

```
from radis import sPlanck
sPlanck(wavelength_min=100, wavelength_max=200, T=6000).plot()
```

![image](https://user-images.githubusercontent.com/16088743/102727709-8fc88200-4327-11eb-81db-d10937d28879.png)

### 💡 Possible solutions

Possibly due to the conversion of nm in vacuum, to air : uses Ciddor refractive index, see [air2vacuum](https://radis.readthedocs.io/en/latest/source/radis.phys.air.html#radis.phys.air.vacuum2air)  

See also: 

```

from radis.phys.air import air_index_dispersion
import numpy as np
import matplotlib.pyplot as plt
w = np.arange(100, 200)
plt.figure()
plt.plot(w, air_index_dispersion(w*1e-3))
```

References on Ciddor : 

https://refractiveindex.info/?shelf=other&book=air&page=Ciddor

erwanp:
Fix : I think that Planck function formula in wavelength should anyway be using the "vacuum" wavelength value (currently I'm quite sure it uses the default value, i.e., in air)

PR #718, opened 2025-01-01
implementing plotly backend in spectrum.py

darshvn:
### Description

This pull request is to address [Issue #179](https://github.com/radis/radis/issues/179)

This PR introduces two new methods to the `Spectrum` class: `_plot_matplotlib()` and `_plot_plotly()`. The existing `plot()` method is now moved to _plot_matplotlib().

The `_plot_matplotlib()` method uses Matplotlib for static plotting, while the `_plot_plotly()` method uses Plotly for interactive plotting. 

**Same spectrum as a matplotlib plot (left) and plotly plot (right) in Jupyter notebook:**

ipnb test

The choice between the two is determined by the `is_kernel()` function, which checks if the script is running in an IPython kernel environment (e.g., Jupyter Notebook).

In addition to resolving this issue, I made a small unrelated improvement by adding the `plt.tight_layout()` function in `_plot_matplotlib.` Previously, some plot elements were being cut off, and this change ensures that all elements fit properly within the figure area.

**Before:**
before

**After:**
before

Fixes #<179>

@erwanp @minouHub

### DIFF
--- a/examples/0_Database_handling/plot_explore_database_vaex.py
+++ b/examples/0_Database_handling/plot_explore_database_vaex.py
@@ -19,10 +19,10 @@ In this example, we plot the dependance of the broadening coefficients to the ro
 """
 import matplotlib.pyplot as plt
 
-from radis.io.hitemp import fetch_hitemp
+from radis.io.hitran import fetch_hitran
 
 #%% Adapt this example if vaex is installed or not
-# June 2024: vaex is not compatible with python>=3.11, see https://github.com/radis/radis/pull/656
+# June 2024: vaex is not compatible with python>=3.11, see https://github.com/radis/radis/pull/656 and the WIP solution here https://github.com/radis/radis/pull/698
 try:
     import vaex
 
@@ -31,7 +31,7 @@ try:
 except ImportError:
     output = "pytables"  # standard pandas format
 
-df = fetch_hitemp("CO2", output=output, load_wavenum_min=2150, load_wavenum_max=2450)
+df = fetch_hitran("CO2", output=output, load_wavenum_min=100, load_wavenum_max=10000)
 print(f"{len(df)} lines in HITEMP CO2; 2150 - 2450 cm-1")
 
 #%%
--- a/examples/0_Database_handling/plot_hitemp_OH_database.py
+++ b/examples/0_Database_handling/plot_hitemp_OH_database.py
@@ -17,9 +17,9 @@ RADIS allows you to use a Vaex DataFrame instead (out-of-RAM).
 See the :ref:`Explore Database with Vaex example <example_explore_database_vaex>`
 """
 
-from radis.io.hitemp import fetch_hitemp
+from radis.io.hitran import fetch_hitran
 
-df = fetch_hitemp("OH")
+df = fetch_hitran("OH")
 print(df.columns)
 
 
@@ -37,5 +37,5 @@ print(df.columns)
 # function. The already downloaded database will be used:
 #
 
-df = fetch_hitemp("OH", load_wavenum_min=31500, load_wavenum_max=33000, isotope="1")
+df = fetch_hitran("OH", load_wavenum_min=31500, load_wavenum_max=33000, isotope="1")
 df.plot("wav", "int")
--- a/examples/0_Database_handling/plot_linestrengths.py
+++ b/examples/0_Database_handling/plot_linestrengths.py
@@ -5,7 +5,7 @@ Scale Linestrengths of carbon-monoxide
 ================================================
 
 This example scales the linestrengths of CO to Tgas=300 from Tref = 296 and then plots
-the linestrengths against the wavenumbers. We are using :py:func:`~radis.io.hitemp.fetch_hitemp`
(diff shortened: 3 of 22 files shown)


radis/spectrum/spectrum.py
```diff
@@ -88,8 +88,17 @@
...
+import plotly.express as px
+import pandas as pd
+
 # %% Spectrum class to hold results )
 
+def is_kernel():
+    if 'IPython' not in sys.modules:
+        return False
+    from IPython import get_ipython
+    return getattr(get_ipython(), 'kernel', None) is not None
```
minouHub, line 101:
```suggestion
def is_running_in_notebook():
        """Check if the code is running in a Jupyter Notebook."""
        try:
            shell = get_ipython().__class__.__name__
            if shell == 'ZMQInteractiveShell':
                return True  # Jupyter notebook or qtconsole
            elif shell == 'TerminalInteractiveShell':
                return False  # Terminal running IPython
            else:
                return False  # Spyder
        except NameError:
            return False  # Probably standard Python interpreter
```

radis/spectrum/spectrum.py
```diff
@@ -2334,11 +2343,70 @@ def clean_error_msg(string):
...
+        force=False,
+        plot_by_parts=False,
+        show=True,
+        show_ruler=False,
+        **kwargs,
+    ):
+        if is_kernel():
+            self._plot_plotly(var, wunit, **kwargs)
+        else:
+            self._plot_matplotlib(var, wunit, **kwargs)
```
minouHub, line 2408:
```suggestion
        if is_running_in_notebook():
            self._plot_plotly(var, wunit, **kwargs)
        else:
            self._plot_matplotlib(var, wunit, **kwargs)
```

minouHub (changes requested):
I tested your implementation on Spyder. It activated Plotly. Here is a suggestion (not tested on Jupyter)

darshvn (commented):
changes merged

radis/phys/blackbody.py
```diff
@@ -198,12 +198,7 @@ def sPlanck(
         waveunit = "cm-1"
     else:
         assert wavelength_min < wavelength_max
-        if medium == "air":
-            waveunit = "nm"
-        elif medium == "vacuum":
-            waveunit = "nm_vac"
-        else:
-            raise ValueError(medium)
+        waveunit = "nm_vac"
```
erwanp, line 201: Why do we remove medium?

radis/spectrum/spectrum.py
```diff
@@ -2334,11 +2350,78 @@ def clean_error_msg(string):
...
+        yscale="linear",
+        show_medium="vacuum_only",
+        normalize=False,
+        force=False,
+        plot_by_parts=False,
+        show=True,
+        show_ruler=False,
+        **kwargs,
+    ):
+        if is_running_in_notebook():
```
erwanp, line 2420: Read docstrings, these are the only one visible on the online docs

erwanp (changes requested):
Hello thanks for the changes.

I think we should not force people to use neither Plotly nor Matplotlib depending on their environment. Some people are Notebook users but might want a Matplotlib output because that's the standard to export in an academia paper.

I'd keep it optional but, with a parameter ``plotting_library`` in ``plot() ``. Parameters could be ``matplotlib, plotly, auto, default``. If ``auto`` use your existing implementation : plotly if the environment is a notebook else matplotlib. If ``defaut``, read from a new key in ``radis.config`` (read more about how we handle user configurations with Radis.json). And the by-default value in the ``defaut_radis.json`` dictionary could be ``auto``

radis/phys/blackbody.py
darshvn: I removed the `medium` parameter from the `sPlanck` function because it was previously used to condition for air/vacuum, but it no longer serves any utility. However, we can retain it for informational purposes.

erwanp: When was it decided we're using vacuum by default ? I missed that part.

darshvn: In your [comment](https://github.com/radis/radis/issues/158#issuecomment-814120736), you mentioned we should use vacuum wavelength, so I made the changes accordingly. Let me know if you'd like me to make any changes.

erwanp: You're right, I indeed said that (in 2021), thanks for refreshing my memory. 

What I meant : the calculations (i.e. the call to the `planck()` function where calculations are performed) require a wavelength in vacuum ; however the user interface should be versatile : users must be able to ask for a Planck Spectrum either in a vacuum or air medium , and it is up to the sPlanck() class to convert the wavelength accordingly before running the calculation. 

So, I think, keep the medium key.

darshvn: thanks for the clarification. changes reverted back and @minouHub defined wavelength air below 132nm.

radis/spectrum/spectrum.py
darshvn: added docstrings, examples and implemented all the necessary functions, please review it and let me know if it requires any changes?

minouHub: @erwanp I guess he is talking about this new PR https://github.com/radis/radis/pull/723

PR #731, opened 2025-01-31
docs:  Added HDF5 save/load gallery example for 'to_hdf5' serialization function and improve to_hdf5 documentation

### LINKED ISSUE #708: Improve docs and Gallery example for the 'to_hdf5' serialization function
Note : we should improve docs and Gallery example for the [to_hdf5](https://radis.readthedocs.io/en/latest/source/radis.spectrum.spectrum.html#radis.spectrum.spectrum.Spectrum.to_hdf5) serialization function

_Originally posted by @erwanp in https://github.com/radis/radis/issues/378#issuecomment-2517021384_

darshvn:
### Description

This pull request is to address Issue #708 

### Changes:
1. Added new example `examples/1_Spectra_handling/plot_hdf5_handling.py` that demonstrates:
   - Saving a spectrum to HDF5 using `to_hdf5()`
   - Loading complete spectrum using `from_hdf5()`
   - Loading partial spectrum by wavenumber range (1020-1040 cm-1) 
   - Loading specific quantities (radiance only)
   - Preserving and verifying metadata (conditions, units)
   - Comparing original vs loaded spectra on same plot

2. Added docstring for `to_hdf5()` method including:
   - Complete parameter descriptions

I'm not sure why am i getting this error even after the hdf5 data is retrieved
`Error reading metadata from spectrum_data.h5`

Fixes #708 
@erwanp

### DIFF
--- a/examples/1_Spectra_handling/plot_hdf5_handling.py
+++ b/examples/1_Spectra_handling/plot_hdf5_handling.py
@@ -16,8 +16,6 @@ It shows:
 
 from radis.test.utils import getTestFile
 from radis.tools.database import load_spec
-from radis.spectrum.compare import plot_diff
-import matplotlib.pyplot as plt
 
 # Load an example spectrum
 my_file = getTestFile("synth-NH3-1-500-2000cm-P10-mf0.01-p1.spec")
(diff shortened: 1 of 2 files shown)


minouHub (changes requested):
I'm fine with the idea. Is there a reason why you use a saved spectrum of NH3 and do not generate one? This could be a way to check if the error `Error reading metadata from spectrum_data.h5` comes from the saved NH3 spectrum or is a more general problem we should dig into.

darshvn:
I wasn't sure which example to use, so @erwanp suggested creating a new example and mentioned that it could be as simple as [this one](https://radis.readthedocs.io/en/latest/auto_examples/2_Experimental_spectra/plot_experimental_spectrum.html#sphx-glr-auto-examples-2-experimental-spectra-plot-experimental-spectrum-py), which uses a test file. I selected the NH3 test file from the available options, but I was encountering same error for all test spectra after saving them to HDF5 and retrieving them. Could you suggest an example that might work better?

minouHub:
Well maybe you could merge these two examples because they reproduce the same stuff. Concerning the error, I checked with other spectra and it still occurs. This error arises because `*** KeyError: 'No object named populations in the file'` (line 693 in `hdf5.py`).  Could you please check why this happens?

darshvn:
Populations data is optional in the spectrum structure.
- When saving to HDF5 (`spec2hdf`), populations are saved only if they exist.
- When loading from HDF5 (`hdf2spec`), the code attempts to read populations but handles their absence.

The error suggests populations weren't saved, possibly because:
- The spectrum was computed without population data.
- `export_populations `wasn't set.
- Populations were empty during saving.

 you can ignore this error most operations will still work, but population data won't be available.

minouHub (approved):

PR #720, opened 2025-01-02
Hitemp manual download

minouHub:
### Description of problems

This pull request is to partially address #717 

Last update of HITRAN website implied:

1. A login is required to download HITEMP file
2. The format of the CO2 HITEMP file is now a single file (6 GB)
3. The isotopologues up to iso=12 are included in CO2 HITEMP

### Solutions

> 1. A login is required to download HITEMP file

I circumvented the problem by asking the user to manually download the .zip or .bz2 files. Not ideal but at least it's working on my computer. Feedback needed!

> 2. The format of the CO2 HITEMP file is now a single file (6 GB)

Some portions of RADIS were changed according to that. The conversion to HDF5 is very long!

> 3. The isotopologues up to iso=12 are now included in CO2 HITEMP

Some changes to the expected dtype of the files were made. We can now computed a lot of isotopologues!!!

PR #739, opened 2025-02-09
Add test for hitemp download

### LINKED ISSUE #737: Test HITEMP download
### 💭 Description

Since January 2025, it is necessary to log into HITRAN to retrieve the HITEMP database. This was sorted #726. However, this makes the tests of the hitemp fetch impossible to run because we need credentials. In other words, if one day someone breaks the download of HITEMP, there is no way to see it immediately.

One thing I could think of is to prepare a manual test of HITEMP download that a user could run. Not the best but at least it's easy to do. Open to more ideas...

darshvn:
### Description

This pull request is to addres #737 

HITEMP Download Test checks:

- login functionality
- Downloads the smallest HITEMP molecule (OH, ~900kb)
- Verifies basic parsing functionality
- Tests wavelength filtering

Fixes #737 
@minouHub


minouHub (changes requested):
There is no need to write new tests (sorry). Everything is ready and was previously commented in #720. Please review the examples and the tests I modified in #720.
One of the things I did was to switch the examples from hitemp to hitran. This is good to accelerate the download but does not always work. Here is a non-exhaustive list of things to do (feel free to tick them):

- [ ] rename `examples/0_Database_handling/plot_hitemp_OH_database.py` to 'plot_hitran'
- [ ] revert to CO hitemp for `examples/0_Database_handling/plot_linestrengths.py`
- [ ] revert to CO hitemp for `examples/1_Spectra_handling/plot_compare_CO_exomol_hitemp.py`
- [ ] revert to CO hitemp for `examples/3_Fitting/plot2_fit_Tgas-molfrac.py`
- [ ] remove try except in `examples/3_Fitting/plot3_fit_Trot-Tvib-molfrac.py`
- [ ] uncomment test in `radis/test/io/test_exomol.py`
- [ ] uncomment test in `radis/test/io/test_hitemp.py`

radis/api/hitempapi.py
```diff
@@ -125,8 +125,8 @@ def get_last(b):
 def setup_credentials():
     """Set up HITRAN credentials and store them in .env file"""
     # Get credentials from user
-    username = input("Enter HITRAN username: ")
-    password = getpass("Enter HITRAN password: ")
+    username = os.environ.get("HITRAN_USERNAME") or input("Enter HITRAN username: ")
+    password = os.environ.get("HITRAN_PASSWORD") or getpass("Enter HITRAN password: ")
```
minouHub, line 129: I think this needs to be done in two steps. 

- First, try to get the saved variables. Please comment that this is especially designed for tests in Travis.
- If `os.environ.get("HITRAN_USERNAME")` gives you `None` or whatever when you're not on Travis, switch to the other solution

.travis.yml
```diff
@@ -8,6 +8,11 @@ python:
 os: linux
 dist: focal #needed for python 3.10 (xenial is default for Travis-CI, but does not support python 3.10)
 
+env:
+  global:
+    - HITRAN_USERNAME=traviscitest@proton.me
+    - HITRAN_PASSWORD=::#78-mBbayZ:#!
```
minouHub, line 14: Please follow https://docs.travis-ci.com/user/environment-variables#define-encrypted-variables to put encrypted variables and not show the HITRAN password to anyone.

minouHub (changes requested):

radis/api/hitempapi.py
darshvn: Changes have been made.

darshvn: I encrypted the variables, but I think it was done on my fork, `darshvn/radis`, since I don’t have access to `radis/radis` to add encrypted variables.

.travis.yml
darshvn: I encrypted the variables, but I think it was done on my fork, darshvn/radis, since I don’t have access to radis/radis to add encrypted variables.

minouHub:
The doc is still failing. Could you maybe store a very small portion of HITEMP in the docs files? This way we won't need to fetch the HITEMP database in readthedocs

minouHub:
Following conversation on Slack, I'll try (as an admin) to do the credentials part on Travis.

For Read The Docs: could you take care of that? I guess it would be simpler to use Hitran when we can. For cases where hitemp is needed, you could save a subdatabase of CO HITEMP. For instance, the portion between 2000 and 2020 cm-1. See example with `HITEMP-CO2-TEST` at `radis\\test\\files\\cdsd_hitemp_09_fragment.txt`

minouHub:
@darshvn I think that you do not retrieve the Travis variable correctly. See https://stackoverflow.com/questions/21047182/how-to-set-environment-variables-in-travis-ci-and-access-them-from-python-script

radis/api/hitempapi.py
minouHub, line 129:
```suggestion
    username = os.environ["HITRAN_USERNAME"] or input("Enter HITRAN username: ")
    password = os.environ["HITRAN_PASSWORD"] or getpass("Enter HITRAN password: ")
```

minouHub: based on https://stackoverflow.com/questions/21047182/how-to-set-environment-variables-in-travis-ci-and-access-them-from-python-script

minouHub:
Just turned on "share encrypted variables with forks" so that @darshvn can test hitemp fetch. 
image
To the admins. I tag you so that you are aware: @erwanp @dcmvdbekerom

minouHub:
Some examples are still inconsistent. This is partially due to my previous contributions. Please update the following:

- [ ] `Download the HITEMP database`. Change the title to `Download the HITRAN database`
- [ ] `Compare NO cross-sections from HITRAN and ExoMol`. Change the title to `Compare NO cross-sections from GEISA and ExoMol`

Issue #666, opened 2024-07-22
`save_memory` being set to `True` in `calc_spectrum`


code29563:
The `save_memory` parameter is [set to `True`](https://github.com/radis/radis/blob/ab6b9ef2dc1ecdc8f099b2eabc3d2e6c37ee0eff/radis/lbl/calc.py#L658) from `calc_spectrum`. The result is that [the example given in the docstring for the `return_factory` parameter](https://github.com/radis/radis/blob/ab6b9ef2dc1ecdc8f099b2eabc3d2e6c37ee0eff/radis/lbl/calc.py#L291) appears to fail unless you also pass the kwarg `save_memory=False` or first [reload the database](https://github.com/radis/radis/blob/develop/radis/lbl/base.py#L3757). `save_memory` is [`False` by default if using `SpectrumFactory` directly](https://github.com/radis/radis/blob/develop/radis/lbl/factory.py#220), and `calc_spectrum` changing that appears to be undocumented.

Should the condition [here](https://github.com/radis/radis/blob/ab6b9ef2dc1ecdc8f099b2eabc3d2e6c37ee0eff/radis/lbl/calc.py#L658) be changed to `if not "save_memory" in kwargs and not return_factory:`, or the setting of `save_memory` to `True` be scrapped entirely?

erwanp:
Indeed it is. Id keep save_memory to True by default because most users need this, but update the example in the docstrings to add the kwargs argument ``save_memory=True``

PR #741, opened 2025-02-20
Updates doc-string to `save_memory=False` 

Prtm2110:
### Description

Fixes: #666

I have set `save_memory=False` in the docstring (assuming [this comment](https://github.com/radis/radis/issues/666#issuecomment-2508923847) is a typo). please correct me if I am wrong, and I will update it.

### DIFF
--- a/radis/lbl/calc.py
+++ b/radis/lbl/calc.py
@@ -306,9 +306,7 @@ def calc_spectrum(
     return_factory: bool
         if ``True``, return the :py:class:`~radis.lbl.factory.SpectrumFactory` that
         computes the spectrum. Useful to access computational parameters, the line database,
-        or to start batch-computations from a first spectrum calculation.
-        By default, ``save_memory=True`` is set when using ``calc_spectrum``.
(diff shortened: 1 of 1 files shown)


radis/lbl/calc.py
```diff
@@ -306,9 +306,11 @@ def calc_spectrum(
     return_factory: bool
         if ``True``, return the :py:class:`~radis.lbl.factory.SpectrumFactory` that
         computes the spectrum. Useful to access computational parameters, the line database,
-        or to start batch-computations from a first spectrum calculation. Ex::
+        or to start batch-computations from a first spectrum calculation.
+        By default, ``save_memory=True`` is set when using ``calc_spectrum``.
+        If you need to retain the full database without reloading, explicitly set ``save_memory=False``. Ex::
```
erwanp, line 311: I think these extra doc lines can be removed (keep the original); but indeed you are right to have added "save_memory=False" in the example itself

erwanp (changes requested):

radis/lbl/calc.py
Prtm2110: I get it, Done!

Issue #736, opened 2025-02-07
Password typing compatible with Spyder


minouHub:
### 🔖 Feature description

Since #726, the user can give his HITRAN login and password to download the hitemp database. Sadly, this feature is not compatible with Spyder because the package `getpass` that hides the password when you type it is not compatible. We could write our own function to type the password, see this suggestion: https://stackoverflow.com/questions/47213310/getpass-is-not-working-for-spyder-python

```python
def prompt_password(user):
    """
    Parameters
    ----------
    user : user name

    Returns
    -------
    text : user input password
    """
    from PyQt5.QtWidgets import  QInputDialog, QLineEdit, QApplication
    from PyQt5.QtCore import QCoreApplication

    # Let's avoid to crash Qt-based dev environment (Spyder...)
    app = QCoreApplication.instance()
    if app is None:
        app = QApplication([])

    text, ok = QInputDialog.getText(
        None,
        "Credential",
        "user {}:".format(user),
        QLineEdit.Password)
    if ok and text:
        return text
    raise ValueError("Must specify a valid password")
```

PR #743, opened 2025-02-20
Ensure HITRAN Login is Compatible with Spyder with `PyQt` Support

Prtm2110:
### Description

Fixes: #736 : Password typing compatible with Spyder
As we are moving `requirements.txt` to `.toml` file for dep, we can put `PyQt5` in optional dep if it is unnecessary for core.

### DIFF
--- a/radis/api/hitempapi.py
+++ b/radis/api/hitempapi.py
@@ -155,7 +155,7 @@ def _prompt_password(user):
             if ok and text:
                 return text
             raise ValueError(
-                "The dialog window was probably closed and the password could not be retrieved left empty. A valid password is required."
+                "The dialog window was probably closed or left empty. Please enter a valid password."
             )
         except ModuleNotFoundError:
             raise ImportError(
--- a/radis/api/hitempapi.py
+++ b/radis/api/hitempapi.py
@@ -121,8 +121,15 @@ def get_last(b):
     return b[non_zero]
 
 
+def running_in_spyder():
+    """Check if the console is running within Spyder."""
+    return "SPYDER_ARGS" in os.environ
+
+
 def _prompt_password(user):
     """
+    Prompts the user for a password securely and handels input if spyder is used.
+    
     Parameters
     ----------
     user : str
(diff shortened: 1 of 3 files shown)


radis/api/hitempapi.py
```diff
@@ -122,11 +121,43 @@ def get_last(b):
...
+        app = QCoreApplication.instance()
+        if app is None:
+            app = QApplication([])
+
+        text, ok = QInputDialog.getText(
+            None, "Credential", f"User {user}:", QLineEdit.Password
+        )
+        if ok and text:
+            return text
+        raise ValueError("Must specify a valid password")
```
minouHub, line 149: Could you provide a more explicit reason for this ValueError?

requirements.txt
```diff
@@ -3,6 +3,7 @@ lxml           # parser used for ExoMol website
 hjson          # Json with comments (for default_radis.json)
 publib 		   # Plotting styles for Matplotlib. Version, update to 0.4.0 needed for matplotlib==3.8. However, only 0.3.2 is compatible with python==3.8 (see #647)
 hitran-api     # HAPI, used to access TIPS partition functions
+PyQt5
```
minouHub, line 6: I agree with your comment in the PR, it could be optional. Maybe you could try the import and if the package is missing, raise an error "Please install PyQt5 with pip install PyQt5"

minouHub (changes requested):
I tried your PR on Spyder and I add to remove this line: https://github.com/radis/radis/blob/a78274093b43c63e69bbc39bbc75359fad769357/radis/io/hitemp.py#L186
Did you try your PR on your side?

Prtm2110:
> I tried your PR on Spyder and I had to remove this line:

Ah Apologies for that, I assumed that will work on spyder as well (it worked on my IDE, and wasn't aware of `running_in_spyder()`). I now tried using spyder but my conda env isn't working there, so could you please check this on your side. I have done some refactoring there please check it out, also do I need to add tests for this changes as most of the code in the `test_hitemp` is commented?

Also on CI `test_calc_hitran_spectrum` is failing but it is passing locally for me, why this could be so?

radis/api/hitempapi.py
```diff
@@ -154,12 +154,17 @@ def _prompt_password(user):
             )
             if ok and text:
                 return text
-            raise ValueError("Password entry was canceled by the user or left empty. A valid password is required.")
+            raise ValueError(
+                "Password entry was canceled by the user or left empty. A valid password is required."
```
minouHub, line 158:
```suggestion
                "Password entry was removed by the user or left empty. A valid password is required."
```

minouHub (approved):
Fine by me. Please commit my suggestion if it's correct. Waiting for tests to pass and we can merge.

radis/api/hitempapi.py
Prtm2110: Actually "cancelled" is right here, if a user cancels the dialog box then it will raise this error. And the failing test is unrelated.

minouHub: Ooh I see what you meant now sorry. Then I would change to "The dialog window was probably closed and the password could not be retrieved"

Prtm2110:
I believe the doc build fail is unrelated.

minouHub (approved):

erwanp:
Note @Prtm2110 I have made changes to requirement.txt (removed it) in #734 ; it might impact this PR once #734 is merged

Prtm2110:
Hi, I had removed PyQt from `requirements.txt` because I think it would be a unnecessary dep for most of the users. But since now we have a toml file we can put it in optional dependencies in #734 itself?

minouHub:
Let's keep PyQt optional for the moment. The error message is clear "install PyQt" for the only case we need it (i.e. when a Spyder user wants to download hitemp).

PR #744, opened 2025-02-21
codelv = "m0" for vaex in api.exomolapi.MdbExomol.set_broadening_coef

### LINKED ISSUE #742: The current radis development version has a loading error for ExoMol CO/12C-16O/Li2015
### 🐛 Describe the bug

This error occurs in the current radis development version, where ExoMol CO Li2015 cannot be loaded. It seems related to #716 (in particular `codelv=="m0"` algorithm. However, the species modified in #716 is "air,"  (``` self.set_broadening_coef(df[mask], add_columns=False, species="air") ``` in `exojax/spec/api.py L274`) and it has been confirmed that H2O/POKAZATEL can still be loaded without issues.

Do you have any idea on this bug? or Do you know any reference on `codelv=="m0"`? @minouHub 

```python
from exojax.spec.api import MdbExomol
nu_grid, wav, resolution = wavenumber_grid( 22920.0, 23000.0, 3500, unit="AA", xsmode="premodit")
mdb = MdbExomol(".database/CO/12C-16O/Li2015", nurange=nu_grid)
```

```sh
  File "/home/kawahara/testco/testco.py", line 9, in <module>
    mdb = MdbExomol(".database/CO/12C-16O/Li2015", nurange=nu_grid)
  File "/home/kawahara/exojax/src/exojax/spec/api.py", line 165, in __init__
    self.activate(df)
  File "/home/kawahara/exojax/src/exojax/spec/api.py", line 274, in activate
    self.set_broadening_coef(df[mask], add_columns=False, species="air")
  File "/home/kawahara/anaconda3/envs/myenv39/lib/python3.9/site-packages/radis-0.15.2-py3.9.egg/radis/api/exomolapi.py", line 1464, in set_broadening_coef
    df["m"] = np.where(
  File "/home/kawahara/anaconda3/envs/myenv39/lib/python3.9/site-packages/vaex/dataframe.py", line 5305, in __setitem__
    self.add_column(name, value)
  File "/home/kawahara/anaconda3/envs/myenv39/lib/python3.9/site-packages/vaex/dataframe.py", line 3480, in add_column
    if _len(ar) != self.length_original():
  File "/home/kawahara/anaconda3/envs/myenv39/lib/python3.9/site-packages/vaex/dataframe.py", line 73, in _len
    return o.__len__()
TypeError: len() of unsized object
```

update: It looks `np.where` does not work for Vaex version?
```python
df["m"] = np.where(
                    df["PQR"] == -1,
                    -df["jlower"],
                ...
```

### 💡 Possible solutions
At least, `np.where` can be replaced to the Vaex version as 
```python
 df["m"] = df.func.where(
                    df["PQR"] == -1,
                    -df["jlower"],
                    df.func.where(
                        df["PQR"] == 0,
                        df["jlower"],
                        df.func.where(df["PQR"] == 1, df["jlower"] + 1, np.nan),
                    ),
                )
```
I understand we need to write the code for vaex. I will try. 

-> See #744

### 🎲 Radis version

0.15.2 (develop)

### 💻 Operating system

Linux

HajimeKawahara:
### Description

This pull request is to address #742

The broadening algorithm `m0` for air in ExoMol was not compatible with `vaex`, so I rewrote the code for `vaex`. 

Additionally, I standardized the return type of `self.alpha_ref` and `self.n_Texp` to return `np.array`.

 I also added `self.engine` to `MdbExomol`.

Fixes #742


HajimeKawahara:
Thank you! 

> Quick question: is it not heavily loading the memory if we convert all the broadening coefficients to a numpy in the end?

Eventually, I would like to implement a method to construct the line density using vaex's lazy evaluation. However, for ExoJAX at the moment, the bottleneck is device memory rather than RAM, so using `np.array` is not critical.

minouHub:
Ok, adding an issue so that we don't forget and we might have a GSOC candidate who want to work on it.

minouHub (approved):

minouHub:
@HajimeKawahara since this PR I cannot run locally the test `plot_compare_CO_geisa_hitran.py`. I believe this is related to the test failing in RDT in #747 
I have a bug because the engine is set to "vaex" but df is actually a pandas DataFrame. I could solve the bug locally with the following but it's not very elegant. Any idea?
```python
if type(df) != pd.DataFrame: #normally "vaex"
      df["m"] = df.func.where(
          df["PQR"] == -1,
          -df["jlower"],
          df.func.where(
              df["PQR"] == 0,
              df["jlower"],
              df.func.where(df["PQR"] == 1, df["jlower"] + 1, 0),
          ),
      )
      # np.nan causes the error in vaex when using map,
      # 0 is regarded as the exeption value because m must not be zero
else:
    if self.engine == "vaex": 
        warnings.warn("The DataFrame is a Pandas DataFrame but the engine should be 'vaex'.")
    df["m"] = np.where(
        df["PQR"] == -1,
        -df["jlower"],
        np.where(
            df["PQR"] == 0,
            df["jlower"],
            np.where(df["PQR"] == 1, df["jlower"] + 1, np.nan),
        ),
    )
```

PR #747, opened 2025-02-23
added an option (broadf_download) in MdbExomol (not) to attempt download all of the broadening files from ExoMol server

HajimeKawahara:
### Description

This pull request is to address to choose the attempt of downloading broadening files from the ExoMol server

In the current specification, ExoMol's broadening file (`.broad`) is attempted to be downloaded for every possible molecule. If `broadf=False`, downloading is skipped, but the default values are used instead. This works fine when the ExoMol server is running, but if the server is down for maintenance or other reasons, it is often the case that the required `.broad` files have already been downloaded. Therefore, it would be useful to have an option to prevent unnecessary download attempts. I have implemented this feature as the `broadf_download` option.

### DIFF
--- a/radis/api/exomolapi.py
+++ b/radis/api/exomolapi.py
@@ -1026,7 +1026,9 @@ class MdbExomol(DatabaseManager):
     bkgdatm: str
         background atmosphere for broadening. e.g. H2, He,
     broadf: bool
-        if False, the default broadening parameters in .def file is used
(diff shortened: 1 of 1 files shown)


minouHub (approved):
You raised a good point. It's sad to raise an error only because we try to download non-existing files and the server is down. I'm ok with it. I let you do a rebase on the develop to pass the tests

minouHub:
You also have some linting issues

PR #755, opened 2025-03-06
Allowing caching in `calc_spectrum` 

### LINKED ISSUE #135: Same script to compute and replot spectra

Ever happened that you computed a spectrum, plotted it, then need to re-plot it few months later and therefore need to recompute it ? What if the spectrum took several minutes to compute ?

- SpectrumFactory already comes with a [feature](https://radis.readthedocs.io/en/master/source/radis.tools.database.html#module-radis.tools.database) that connects it to a SpecDatabase (a folder of Spectrum objects) which initially checks that the Spectrum exists, returns it if it does, or recomputes and add to the folder if it does not 

We could have a similar feature for calc_spectrum that automatically saves to a file, and retrieves from the file if it's there already. 

1. we give an output filename to calc_spectrum. 
2. if filename exists : 

  - load the spec
  - check that metadata correpsond to computational conditions
  - if not > error or recompute 
  - if yes > return the Spectrum directly 

3. else:
  - compute, and save 
 

```
s = calc_spectrum(Tgas=... ,   output_file='blabla.spec', **kwargs)
s.plot()
```

Questions : 
- what [store](https://radis.readthedocs.io/en/latest/source/radis.spectrum.spectrum.html#radis.spectrum.spectrum.Spectrum.store) parameters to use ? Allow other parameters to be given in **kwargs ?  (note : saving all lines is probably a bad idea --> if this feature is used in a loop with spectra with millions of lines this would quickly use all of the disk space !)

---

Prtm2110:
### Description
Fixes #135
Following this for now - https://github.com/radis/radis/issues/135#issuecomment-2701254731

### DIFF
--- a/radis/api/tools.py
+++ b/radis/api/tools.py
@@ -5,6 +5,7 @@
 """
 # TODO refactor : rename this file as hitran_utils.py
 
+import pathlib
 from warnings import warn
 
 import numpy as np
@@ -17,6 +18,8 @@ try:
 except ImportError:
     vaex = NotInstalled(*not_installed_vaex_args)
 
+
+import radis
 
 from ..misc.warning import PerformanceWarning
 
@@ -317,3 +320,8 @@ def replace_PQR_with_m101(df):
             ),
             category=PerformanceWarning,
         )
+
+
+def root_dir():
+    """return the root directory of the Radis package"""
+    return str(pathlib.Path(radis.__path__[0]).parent)
--- a/radis/lbl/calc.py
+++ b/radis/lbl/calc.py
@@ -18,7 +18,7 @@ Routine Listing
 
 
 from copy import deepcopy
-from os.path import exists
+from os.path import exists, join
 
 import numpy as np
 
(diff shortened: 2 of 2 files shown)


radis/api/tools.py
```diff
@@ -317,3 +320,8 @@ def replace_PQR_with_m101(df):
             ),
             category=PerformanceWarning,
         )
+
+
+def root_dir():
```
erwanp, line 325: there is [getProjectRoot()](https://github.com/radis/radis/blob/6136932b371cc10ad32cdafe2a0a0ea0ff7c5e54/radis/misc/utils.py#L21) already

radis/lbl/calc.py
```diff
@@ -401,6 +403,16 @@ def calc_spectrum(
     :py:class:`~radis.lbl.factory.SpectrumFactory`
     """
 
+    # path = join(radis.root_dir(), 'cache', f"{s_name}.spec")
+    path = join(root_dir(), "cache", "CO-hitran-700K-#5328.spec")
```
erwanp, line 407: /!\ this is hardcoded

erwanp (changes requested):

radis/lbl/calc.py
minouHub: What you proposed as a naming convention is good to me ("molecule-database-T-###.spec"). There is a sufficient number of information (e.g. no need for pressure) and a random number to avoid overwriting.

minouHub (changes requested):
I would rather have this option in SpectrumFactory or make another function. `calc_spectrum` is supposed to be a simple function and there are already too many arguments in my opinion

PR #757, opened 2025-03-07
Sort failing tests in doc

minouHub:
## Description

This pull request is mainly here to address the failing tests in the docs. Commented some parts of the failing tests to try to understand.

### List of actions:

- [x] removed tests failing the doc `plot_compare_CO_exomol_hitemp.py` and `plot_compare_CO_geisa_hitran.py` . It is now replaced by `plot_compare_NO_hitran_geisa.py`. I think the test was trying to download the new hitemp CO database...
- [x] switched Travis to sequential tests of Python 3.10 and 3.11 instead of parallel. Recall that we need to test the case when vaex can be installed (python<=3.10) and when it cannot be installed (python>3.10)


dcmvdbekerom:
Hi Nicolas,

I noticed that the fast tests are randomized, but the slow tests are in alphabetical order.
Can you also randomize the slow tests? this will reduce pressure on the HITRAN server!
I guess this PR is a good place to do that.

erwanp:
or sequentially run the two python versions? this will anyway reduce the credits spent if tests are bound to fail

.travis.yml
```diff
@@ -138,6 +138,9 @@ jobs:
     - stage: Py38
       name: "Python 3.8"
       python: 3.8
+    - stage: Py310
+      name: "Python 3.10"
+      python: 3.10
```
erwanp, line 143: is this temporary or final?

minouHub: If we run the tests sequentially: yes this will be final

minouHub: Check it out https://app.travis-ci.com/github/radis/radis/builds/274479051

minouHub:
> I noticed that the fast tests are randomized, but the slow tests are in alphabetical order.
Can you also randomize the slow tests? this will reduce pressure on the HITRAN server!
I guess this PR is a good place to do that.

@dcmvdbekerom it actually does not work despite my previous attempts. I think I'll keep the tests sequential for now.

.travis.yml
erwanp: What triggered my comment is I was seing a 3rd Python version added. Actually, the issue is there is this Python 3.8 still there. If 3.10 is already Vaex-compatible; we could remove 3.8 ?

minouHub: As long as it passes on the master for a new version, I'd like to keep it. If one day it raises problems, let's remove it. I could add a comment next to it

erwanp: ok

examples/1_Spectra_handling/plot_compare_CO_exomol_hitemp.py
erwanp, line 1: file is removed ? didn't you just rename it ? if so you need to add the new file ; or directly use ``git mv OLD NEW`` to rename

examples/1_Spectra_handling/plot_compare_CO_geisa_hitran.py
erwanp, line 1: same

examples/1_Spectra_handling/plot_compare_CO_exomol_hitemp.py
minouHub: There is no point having two files for this comparison. I also wanted to test if the problem is actually coming from these files (answer is "yes").

examples/1_Spectra_handling/plot_compare_CO_geisa_hitran.py
minouHub: same answer

minouHub:
ok. This does not make sense. 

- The doc is failing on `plot_compare_NO_hitran.py`. This issues comes from the computation of the geisa spectrum during the download (trying to get a file named `None`). https://readthedocs.org/projects/radis/builds/27437991/

If I write the same test in Travis

- [792caa6](https://github.com/radis/radis/pull/757/commits/792caa674fb269284511aae3bd432d4cee9e7eaf) shows that now the hitran spectrum has a problem with vaex but all the other tests are passing - see https://app.travis-ci.com/github/radis/radis/jobs/631195089
- I revert the order to see if the pytables are passign the test... and now everything is ok [11c014a](https://github.com/radis/radis/pull/757/commits/11c014a94373c698f50be084e02832837a39c7b6)  - see https://app.travis-ci.com/github/radis/radis/builds/274479479

PR #762, opened 2025-03-10
Improve Plotly and Matplotlib Plotting: Utilize kwargs for Customization

### LINKED ISSUE #760: `Spectrum.plot`'s kwargs are not being pass down to `plotly` when it is used as backend
### 💭 Description

When using [plotly](https://github.com/Prtm2110/radis/blob/6136932b371cc10ad32cdafe2a0a0ea0ff7c5e54/radis/spectrum/spectrum.py#L2318) (or running in notebook) the `kwargs` are not being utilised, we should be able to do something like `s.plot('radiance', width = 800, height = 800)`.

mohyware:
### Description
This pull request is to enhance the plot function by ensuring that kwargs such as width, height, color, and lw (line width) are properly utilized when using the Plotly library. Additionally, support for width and height has been added when plotting with Matplotlib.

Fixes #760 

New Usage Examples:
```bash
                molecule='CO',
                isotope='1,3,10,1000',
                pressure=1.01325,   # bar
                Tgas=700,           # K
                mole_fraction=0.1,
                path_length=1,      # cm
                )
```
### Default
Screenshot from 2025-03-10 16-14-34

```bash
s.plot('radiance', plotting_library='plotly')  
```
### With Customization
Screenshot from 2025-03-10 16-14-21

```bash
s.plot('radiance', plotting_library='plotly',lw=4, color='grey', width=1200, height=600)    
```
# Using matplotlib
### Default
Screenshot from 2025-03-10 16-05-43
```bash
s.plot('radiance', plotting_library='matplotlib')  
```
### With Customization
Screenshot from 2025-03-10 16-05-53
```bash
s.plot('radiance', plotting_library='matplotlib',lw=4, color='grey', width=1200, height=600)    
```

### DIFF
--- a/examples/6_Misc/plot_styles.py
+++ b/examples/6_Misc/plot_styles.py
@@ -48,3 +48,20 @@ for plotlib, context, style in [
     s.plot()
     plt.title(", ".join(f"{v}" for v in radis.config["plot"].values()))
     plt.tight_layout()
+
+#%%
+# You can also choose to not show the figure immediately, and instead return
+# the Matplotlib or Plotly object. This allows for **advanced customization**
+# (e.g., modifying the matplotlib figure).
+
+fig, ax = s.plot("radiance", plotting_library="matplotlib", show=False)
+
+ax.set_xlabel("Updated X-axis Label")
+ax.set_ylabel("Updated Y-axis Label")
+ax.grid(True, linestyle=":", linewidth=1)
+ax.set_facecolor("beige")
+
+fig.gcf().set_size_inches(8.6, 5)  # Resize
+fig.gcf().set_dpi(150)  # Change DPI
+# Then Show
+fig.show()
--- a/radis/spectrum/spectrum.py
+++ b/radis/spectrum/spectrum.py
@@ -2238,7 +2238,7 @@ class Spectrum(object):
                 still experimental ! Try it, feedback welcome !
         **kwargs: **dict
             kwargs forwarded as argument to plot (e.g: lineshape
-            attributes: `lw=3, color='r', width=800, height=600`)
+            attributes: `lw=3, color='r'`)
 
         Returns
         -------
@@ -2275,14 +2275,9 @@ class Spectrum(object):
 
         set_style()
 
-        # set figure size (width, height)
-        dpi = 100  # dots per inch
-        width = kwargs.pop("width", 800)
-        height = kwargs.pop("height", 600)
-
         if nfig == "same":
             nfig = plt.gcf().number
-        fig = plt.figure(nfig, figsize=(width / dpi, height / dpi))
+        fig = plt.figure(nfig)
 
         # If figure exist, ensures xlabel and ylabel are the same (prevents some
         # users errors if plotting difference units!)... Note that since
@@ -2385,8 +2380,8 @@ class Spectrum(object):
 
         if show:
             plt.show()
-
-        return line
+        else:
+            return plt
 
     def _plot_plotly(
         self,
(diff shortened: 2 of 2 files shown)


erwanp:
nice. can you add some screenshots of the results in the core PR text?

Prtm2110:
Thank you for this, what I had in my mind was that we separate `kwargs` and pass them to  `update_layout` and `update_traces` based on what properties they accept. 
For eg - `paper_bgcolor="lightgray"` and `plot_bgcolor="white"` are meant for `update_layout` but `line_color` can only be passed to `update_traces`

mohyware:
done @erwanp

mohyware:
Sure We have two approaches
1. **Using the most common ones and passing them to the appropriate function** (similar to the current approach):
```
trace_kwargs = { k: kwargs.pop(k) for k in ["line_width", "line_color"] if k in kwargs } 

layout_kwargs = { k: kwargs.pop(k) for k in ["paper_bgcolor", "plot_bgcolor"] if k in kwargs }

```
2. **Prefixing parameters to indicate their target function** (we can get all functionalities from plotly):

```
s.plot('radiance', plotting_library='plotly',t_line_width=4, l_paper_bgcolor=600)

pre t_ for update_traces kwargs # t_line_width
pre l_ for update_layout kwargs # l_paper_bgcolor
```
Which one should we go with? @Prtm2110

Prtm2110:
I prefer the first one because in 2nd we are basically changing arguments for `plotly` (which will differ from what plotly have in their documentation) and users won't expect this, but @erwanp shall make the call. 
> ["line_width", "line_color"] 

I think there will be more of these so I think you should find all of them and put up a list, this will be hardcoded but we don't have a choice I guess. Thanks!

darshvn:
Hello @mohyware @erwanp 

Issues with the PR Solution:  

- Hardcoded Parameter Mappings:  
  - Manually maps matplotlib to plotly parameters, making it non-maintainable and error-prone.  
  - Unnecessarily complex handling & manual parameter management increases the chance of mistakes.  
  - Mixes high-level (plotly express) and low-level (graph objects) APIs.  

- False Compatibility:  
  - Provides limited matplotlib compatibility, creating false expectations.  
  - Better to be explicit about using plotly's native parameters.  

- Not Future-Proof:  
  - Likely to break with future plotly updates.  
  - Requires continuous maintenance for both matplotlib and plotly changes.  
  - Locks into specific parameter mappings.
  
Better approach:

- Let's plotly's own validation handle parameters
- Works with all plotly parameters automatically
- No maintenance needed for parameter mappings
- Clear expectations about parameter usage

image

mohyware:
Both solutions look great to me we can wait for @erwanp to share his thoughts.

erwanp:
thanks for your contributions and critical thinking. A few insights on how to make the decision here :

Radis is not a plotting library, nor a common interface to Plotly+Matplotlib. As mentioned up, having this objective would create huge maintenance in the future, that's not our goal. 

We're just providing user with a way to quickly plot something decent looking. 

These plot() functions should return the Matplotlib or Plotly objects so that users can fine-tune the final output as needed.

(it might be worth adding examples on how to do that on the gallery, based on a Radis.test_spectrum() calculation.)

Here, let's choose the implementation that is the most clear and low maintenance.

mohyware:
@erwanp 

Sure! What about this solution with ``plot_mode`` variable? Should I proceed with it?

1- This is the same as the current behavior (quick plot)
it is also the default if ``plot_mode`` is not passed.
```
s.plot('radiance', plotting_library='plotly',plot_mode='simple')  
```
2- This is more customizable it returns an object that the user can modify before plotting.
```
obj = s.plot('radiance', plotting_library='plotly',plot_mode='advanced')  

obj.update_traces(line_color='green', line_width=4)

obj.update_layout(
width=1200,
height=600,
)  

obj.show()
```

minouHub:
- Long wait
- Working, with an example

>> Merging!

Issue #581, opened 2023-07-19
Slit size depends on the wave range that it is applied on


saradelahaie:
### 🐛 Describe the bug

When I apply a slit by giving its FWHM in nm, I get a slit size that changes depending on the used wave range.
![image](https://github.com/radis/radis/assets/138567099/5a68a00a-d62e-46ab-816f-c041ea2220c7)

```
from radis import calc_spectrum
for w_min in [900, 1600]:
	s = calc_spectrum(w_min * u.nm, 2500 * u.nm,
					  molecule=molecule,
					  isotope='1',
					  pressure=1.01325,  # bar
					  Tgas=T,
					  path_length=1 * u.cm,  # cm
					  mole_fraction=1,
					  databank='hitran'
					  )
	s.apply_slit(slit_size, unit="nm")
	s.take('radiance').plot(nfig=1)
plt.show()
```

### 💡 Possible solutions

It seems that the problem is solved when translating the slit size to cm-1 (but then, it's not constant over the wave range in cm-1).

### 🎲 Radis version

0.14

### 💻 Operating system

Windows

erwanp:
Possible fixes : 

Method 1 :
- Interpolate the spectrum from the calculated waverange (here: cm-1) to the slit waverange (here : wavelengths)
- Apply sit 
- Interpolate back to the first waverange 

We'll induce small interpolation errors. 

Method 2: 

- divide the spectrum in ranges where the slit expressed in cm-1 is approximately constant
- apply slit on each range, merge


Method 1 sounds easier

minouHub:
@saradelahaie I hope this issue is not hindering your progress. Let me know if this is the case. Otherwise I'll come back later on this issue (and secretly hope someone will apply Erwan's fix).

minouHub:
Labeled "Good first issue" since Erwan provided a method to solve the problem

tactipus:
i wanna take a stab at this

tactipus:
I'm looking at this now. Does this bug require knowledge of any advanced mathematics? I confess I only got up to Calc 2...

minouHub:
If you can understand the steps described by Erwan, it means you're good enough at maths :)

tactipus:
would you kindly explain the "slit_size" variable? cuz the docstring states that there are different ways to handle the "slit_size" parameter depending on the data type.

minouHub:
In the example, the slit_size is a float. For instance, you can set it to `slit_size=1`

tactipus:
[...]

Issue #1, opened 2018-02-23
Applying slit function - change of wavelength range


minouHub:
When using s.applyslit(), the convolution change the range of the calculation. I know you already solved the problem before, but obviously not on this version.

Minou

erwanp:
It's actually on purpose: The wings of the spectrum are not valid after the convolution (because we don't know what lines are out of the range) so they are discarded and only the valid part is kept.

It could be added as an option in apply_slit to keep the same range, but it shouldn't be the default behavior

minouHub:
Alright, so I should calculate my spectrum on a wider wave-range to take that into account.

PR #763, opened 2025-03-11
Feat: Add interpolate spectrum

salmanhiro:
### Description

Fixes #<Issue Number>

### DIFF
--- a/radis/lbl/calc.py
+++ b/radis/lbl/calc.py
@@ -33,7 +33,22 @@ from radis import config
 from radis.misc.basics import all_in
 from radis.misc.utils import Default
 from radis.spectrum.spectrum import Spectrum
+from scipy.interpolate import interp1d
 
+def interpolate_spectrum(spectrum, new_wavelengths, kind='cubic'):
+    """
+    Interpolates a spectrum to a new wavelength range.
+
+    Parameters:
+    - spectrum: radis Spectrum object
+    - new_wavelengths: numpy array of desired wavelengths
+    - kind: interpolation method ('linear', 'cubic', etc.)
+
+    Returns:
+    - interpolated intensity values at new wavelengths
+    """
+    f_interp = interp1d(spectrum.wavelength, spectrum.radiance, kind=kind, fill_value="extrapolate")
+    return f_interp(new_wavelengths)
 
 # %%
 def calc_spectrum(


erwanp:
hello @salmanhiro, thanks for taking on this issue. Before you move further, we must answer a critical question : is a constant-nm slit function physically correct, and in what setup?

in Dispersive spectrometer of Czerny Turner type, for instance, slit dispersion will depend on the wavelength, because it essentially is a spatial dispersion over nearby CCD pixel, but the pixel-to-wavelength conversion depends on the wavelength itself : this relation is given by the spectrometer wavelength calibration procedure. Essentially, a slit function measured at 632 nm and expressed in nm will be different at 300 nm or 1500 nm. The correction can be calculated with the Spectrometer reciprocal dispersion function. 

in that context (for these Czerny Turner instruments), there is already a built in function in Radis to correct for the dispersion. It will basically cut the spectrum over small subsets, calculate the updated slit function for each subset, apply the slit and merge the spectra. See the ``slit_dispersion`` parameter in apply_slit : https://radis.readthedocs.io/en/latest/source/radis.spectrum.spectrum.html#radis.spectrum.spectrum.Spectrum.apply_slit

For these reasons, the expected behavior is, anyway, not to apply a nm-slit over such a broad range as in your example above. This is from an experimental point of view. When applying a proper slit function with the dispersion function, the problem evidenced here and in #581 disappears because each subsets are narrow enough. 

I also understand that : 

1. there might be other sound use cases where applying a constant nm slit makes physical sense. I don't know of them but would be interested to learn. 

2. from a purely numerical point of view, and for a user without experimental spectroscopy knowledge, the fact to have unconsistent behavior is anyway a problem,which we must fix. Shall we fix it by Implementing a constant nm slit (for instance with the implementation you suggest) or by Raising an error and telling the user this is not valid? so they don't make mistakes. (unless there are valid cases, see #1). 

Because one of Radis core principle is to make advanced models safely available to people without advanced spectroscopy knowledge, Id rather raise an error and suggest anorger project. Better crash than hide a mistake.


tagging @minouHub, @saradelahaie

salmanhiro:
> hello @salmanhiro, thanks for taking on this issue. Before you move further, we must answer a critical question : is a constant-nm slit function physically correct, and in what setup?
> 
> in Dispersive spectrometer of Czerny Turner type, for instance, slit dispersion will depend on the wavelength, because it essentially is a spatial dispersion over nearby CCD pixel, but the pixel-to-wavelength conversion depends on the wavelength itself : this relation is given by the spectrometer wavelength calibration procedure. Essentially, a slit function measured at 632 nm and expressed in nm will be different at 300 nm or 1500 nm. The correction can be calculated with the Spectrometer reciprocal dispersion function.
> 
> in that context (for these Czerny Turner instruments), there is already a built in function in Radis to correct for the dispersion. It will basically cut the spectrum over small subsets, calculate the updated slit function for each subset, apply the slit and merge the spectra. See the `slit_dispersion` parameter in apply_slit : https://radis.readthedocs.io/en/latest/source/radis.spectrum.spectrum.html#radis.spectrum.spectrum.Spectrum.apply_slit
> 
> For these reasons, the expected behavior is, anyway, not to apply a nm-slit over such a broad range as in your example above. This is from an experimental point of view. When applying a proper slit function with the dispersion function, the problem evidenced here and in #581 disappears because each subsets are narrow enough so that the slit is approximately the same on the whole subset range. The max error is calculated and controlled by the `slit_dispersion_threshold` parameter. This is already built in
> 
> I also understand that :
> 
> 1. there might be other sound use cases where applying a constant nm slit makes physical sense. I don't know of them but would be interested to learn.
> 2. from a purely numerical point of view, and for a user without experimental spectroscopy knowledge, the fact to have unconsistent behavior is anyway a problem,which we must fix. Shall we fix it by Implementing a constant nm slit (for instance with the implementation you suggest) or by Raising an error and telling the user this is not valid? so they don't make mistakes. (unless there are valid cases, see [Applying slit function - change of wavelength range #1](https://github.com/radis/radis/issues/1)).
> 
> Because one of Radis core principle is to make advanced models safely available to people without advanced spectroscopy knowledge, Id rather raise an error and suggest anorger project. Better crash than hide a mistake.
> 
> tagging @minouHub, @saradelahaie

Hi @erwanp, thanks for your feedback and context on this issue. Since there is already a way to handle such case, I think we can just left the correct way as to use this function. 

I wonder if is it fine if we just throw an error and tell them to give the function to `slit_dispersion` argument?

salmanhiro:
Hi @erwanp do you have idea about meaningful physical wavelength range limit to define how broad is the range to not applying constant nm slit and then raising not valid error?

minouHub:
I think we could always find a argument to say that a 10% change is a good limit, or 1%. Could we just accept that typing a constant in the argument of the function means "apply a constant broadening to the spectrum"? I vote for the first option suggested by @erwanp :

>  Shall we fix it by Implementing a constant nm slit (for instance with the implementation you suggest)

minouHub (commented):
Did a naive review of your code. When I try to run your bug example in your branch, I still get the previous behavior. Is this normal?

salmanhiro:
> Did a naive review of your code. When I try to run your bug example in your branch, I still get the previous behavior. Is this normal?

I commented this out when you checked it, that's why there is a part of commented code. Now I bring it back and working on the tests.

minouHub:
Hello,
One test is failing but if you check this specific tests it's normal because you changed a bit the way a convolution is performed. 

- [ ] Could you update `test_calc_spectrum_overpopulations`?

PR #764, opened 2025-03-11
Update databank entries in radis.json

### LINKED ISSUE #661: Update databank entries in radis.json [apply policy of atomic databases]
I've noticed that the current functionality of RADIS (for at least some of the databanks) is to modify the local databank files without updating the relevant entry in radis.json, e.g. redownloading files when they're missing or the `cache` argument is set to "regen", and then not updating the "download_date". Rather, an attempt to 're-register' and already-registered databank to update its values would just result in a `DatabaseAlreadyExists` error. It's not clear if support for updating databank entries was what was meant by the ability to 'Delete/update existing databases from the command line' referred to in [a previous issue](https://github.com/radis/radis/issues/167#issue-777636105).

It's also understandable if it's considered undesirable to let the library modify an existing entry in radis.json, thereby potentially overwriting user-entered data automatically, and hence the current functionality where RADIS seems to be restricted to appending new entries. But it's debatable then whether it should still be modifying local files automatically and thereby rendering existing databank entries false and incorrect.

In addb90c0dc2fe3bdfb915defc6e548802ef0df1f I've implemented a new config parameter "ALLOW_OVERWRITE" that allows the user to choose between RADIS modifying local files and hence updating entries of registered databanks, and not updating entries of registered databanks and hence not modifying local files (instead raising an Exception wherever it needs to do so). Details are in [`__init__.py`](https://github.com/radis/radis/blob/addb90c0dc2fe3bdfb915defc6e548802ef0df1f/radis/__init__.py#L69). It arose as a side task in the course of adding support for different partition functions for [Kurucz linelists](https://github.com/radis/radis/pull/652), so it's currently functional for `fetch_kurucz` which might provide a model for the other databank-specific functions called in `fetch_databank`, in case the ability to update databank entries is worth adding to RADIS.

mohyware:
### Description
- Implemented the ALLOW_OVERWRITE functionality following the Kurucz approach
  - Allowing users to control whether RADIS modifies local files and updates registered databank entries or raises an exception instead.
  - Fixed an issue where files were redownloaded when missing or when `cache="regen"`, but the `"download_date"` was not updated accordingly.
- Applied these changes to the GEISA database.

Should I proceed with this approach for other databases as well?

Fixes #661


radis/io/geisa.py
```diff
@@ -139,43 +144,78 @@ def fetch_geisa(
...
 
-    # Get missing files
-    download_files = ldb.get_missing_files(local_files)
+    get_main_files = True
 
-    # Download files
-    if len(download_files) > 0:
-        if urlnames is None:
-            urlnames = ldb.fetch_urlnames()
+    if len(local_files) > 1 or len(urlnames) > 1:
```
minouHub, line 175:
```suggestion
    if len(local_files) != 1:
```
or is there something I'm missing?

radis/io/geisa.py
```diff
@@ -139,43 +144,78 @@ def fetch_geisa(
...
+        raise Exception("only 1 database file is expected")
```
minouHub, line 176: Be more explicit:
"Found the following files [...] but only 1 is expected"

radis/api/geisaapi.py
```diff
@@ -373,6 +377,48 @@ def fetch_urlnames(self):
...
+            for url in urlnames
+        ]
+
+        try:
+            os.mkdir(local_databases)
+        except OSError:
+            pass
+        else:
+            if verbose:
+                print("Created folder :", local_databases)
```
minouHub, line 403:
```suggestion
                print("Created folder:", local_databases)
```

radis/io/geisa.py
```diff
@@ -139,43 +144,78 @@ def fetch_geisa(
...
+            try:
+                ldb.download_and_parse([url], [file], 1)
+            except OSError:
+                if i == len(main_urls) - 1:  # all possible urls exhausted
+                    print(f"Error downloading {url}.")
+                    print(f"No source found for {ldb.molecule}")
+                    raise
+                else:
+                    print(f"Error downloading {url}")
+                    continue
```
minouHub, line 198: You should return the OSerror to help the debugging

radis/io/geisa.py
```diff
@@ -139,43 +144,78 @@ def fetch_geisa(
...
+            else:
+                print(f"Successfully downloaded {url}")
```
minouHub, line 200:
```suggestion
                if verbose: print(f"Successfully downloaded {url}")
```

minouHub (commented):
Seems like you're on the right path. Please continue with other databases!

If you realize that other portions of the code are redondant between the databases, do not hesitate to write a common function. 

Btw, the NIST database (atomic) will be deployed soon on the develop branch, see #689

radis/io/geisa.py
mohyware: I followed Kurucz, but I think I'll change this on both now

radis/api/geisaapi.py
```diff
@@ -431,24 +477,49 @@ def parse_to_local_file(
...
+                        "wavenumber_max": self.wmax,
+                    }
+                )
+        # else:
+        #     raise Exception("GEISA database can't be registered until the correct url to use is known")
+        dict_entries.update(
+            {
+                "parfuncfmt": "GEISA",
+            }
+        )
```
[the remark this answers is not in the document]
mohyware: Are you sure? Removing it caused this error. Also, my bad it should be hapi not GEISA.

Screenshot from 2025-03-13 01-25-31
[here](https://github.com/radis/radis/blob/9a8ccce2540bd28b9573c5968b90320d58eb6ff6/radis/misc/config.py#L976)

radis/io/geisa.py
mohyware: With this approach, we start with an empty array for local files, then search for them in the entries. If the db is not registered, we retrieve them using get_possible_files in the download section. So using if len(local_files) != 1 will raise an error if there are no entries, whereas it should only raise an error if there is more than one file or URI in radis.json.

radis/io/geisa.py
mohyware: I followed Kurucz, but I think I'll change this on both

mohyware:
Allow Overwrite Config must be set to true for this to pass tests.

radis/io/geisa.py
minouHub: Sorry my bad

radis/api/geisaapi.py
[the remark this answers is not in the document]
erwanp: isn't that key the key we use to define which partition function we want to use with this database ? (and indeed, we can use different partition functions with different databases)

radis/api/hitempapi.py
```diff
@@ -713,34 +714,51 @@ def parse_to_local_file(
...
+            local_files, urlnames = self.get_filenames()
+            info = f"HITEMP {self.molecule} lines ({self.wmin:.1f}-{self.wmax:.1f} cm-1) with TIPS-2017 (through HAPI) for partition functions"
 
-        # Add energy level calculation
-        if self.molecule in MOLECULES_LIST_NONEQUILIBRIUM:
-            dict_entries[
-                "info"
-            ] += " and RADIS spectroscopic constants for rovibrational energies (nonequilibrium)"
-            dict_entries["levelsfmt"] = "radis"
+            if self.molecule in ["CO2", "H2O"]:
```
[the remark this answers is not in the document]
erwanp: /!\ this is true for HITEMP-2025 (latest "hitemp") but we must keep a working code for HITEMP-2010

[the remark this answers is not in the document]
minouHub: @erwanp this should be a separate issue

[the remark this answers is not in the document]
minouHub: Alright, but @mohyware you can forget my suggestion

radis/api/geisaapi.py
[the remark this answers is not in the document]
minouHub: So ... Is there a point keeping this key? I can use any partition function with any line database. For instance, I can use the `hapi` or `exomol` partition functions for the CO lines from the `GEISA` database right?

minouHub:
Hello,
You need to rebase to pass the tests. I'm focusing my reviews on PR passing the tests

mohyware:
@minouHub Sorry for that. I resolved the conflict, but tests keep failing cause of ALLOW_OVERWRITE.


image

minouHub:
If this is expected behavior, you can add ALLOW_OVERWRITE = True in the tests

mohyware:
image
it's NIST server this time.

radis/api/geisaapi.py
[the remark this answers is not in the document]
minouHub: @erwanp reminder. Do we need to keep this argument? I don't think so.

minouHub (approved):
Fine by me but waiting for @erwanp answer on "parfuncfmt".

Possible solution. We leave this argument as it is and solve that question later in a separate issue.

@mohyware please resolve conflict to allow the merging when Erwan will answer.

mohyware:
@minouHub  done

radis/api/geisaapi.py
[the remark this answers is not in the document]
minouHub: No news from @erwanp. Not urgent, I set an issue to not forget #806

autumnthothincorperated (commented):
Return quote

PR #698, opened 2024-09-18
Vaex compatible up to python<=3.12

minouHub:
### Description

This pull request allows the install of vaex up to `python<=3.12`. 

- [ ] Adapt comments written when vaex was **not** compatible with python 3.11 and 3.12
- [ ] Allow install of vaex up to `python<=3.12`
- [ ] Travis tests can be run on only 1 python version. Stop random pytest.
- [ ] Add link to vaex release for documentation.

Documentation:

- Pull request: https://github.com/vaexio/vaex/pull/2331
- Release: (waiting)

Issue #53, opened 2020-01-07
Long term Todo list 


erwanp:
List of small issues / future features / low priority. 

New contributors: you may want to pick one of these and implement them. 

# Performances

- [ ] 'auto' parameter for 'chunksize', based on number of lines * gridpoints & memory size of computer

# Features 

- [ ] ... 

# User interface

- [ ] ...

dcmvdbekerom:
Can we add user defined vibrational populations to the long term todo list?

erwanp:
Oh. We have that, or almost. There is an overpopulation key in calc_spectrum and Spectrum Factory. Divide your arbitrary vibrational population by the Boltzmann-distribution and you can give it directly as an input !

Springder:
I can't wait to load .spec files in MATLAB！

erwanp:
Hello @Springder  ! I do not have a MATLAB installation on my machine so I won't be able to work on that. Do you want to have a look at it ? If MATLAB already has a way to load .json files this shouldn't be hard; as .spec are only hidden .json

Springder:
It's very easy to load .json files in matlab, so .spec can be load as well . We just need a tool, [JSONLab](https://github.com/fangq/jsonlab): a native JSON/UBJSON/MassagePack encoder/decoder for MATLAB/Octave.  But it can't support .spec files compressed yet.

tanishadahale:
Hi, I'd like to work on "switch from os.path to pathlib" refactor.

minouHub:
Hello,
I updated a few items in your list @erwanp (notably removed a few that are not a priority)

PR #777, opened 2025-03-23
Add auto option for chunksize based on nlines, ngridpoints, and available memory

### LINKED ISSUE #776: TODO: Implement auto option for chunksize based on nlines, ngridpoints, and available memory
Currently, the chunksize parameter in RADIS requires manual tuning, which can be difficult to optimize for large datasets. 

When set to None, all lines are processed at once this is typically faster but can significantly slow down the system due to high memory usage.
 
To enhance performance and memory efficiency, we need an 'auto' mode that dynamically determines the optimal chunksize based on:

- **Number of spectral lines (`nlines`)**    
- **Grid points (`ngridpoints`)**    
- **Available system memory**

mohyware:
### Description
This PR introduces an **`'auto'` option for `chunksize`**, allowing dynamic adjustment based on:
- **Number of spectral lines (`nlines`)**    
- **Grid points (`ngridpoints`)**    
- **Available system memory**

This follows the version implemented in [radis-benchmarks](https://github.com/radis/radis-benchmark/blob/9daa0c8f57b1384e11bc315d9926971883065a66/benchmarks/benchmarks.py#L168)
Fixes #776

### DIFF
--- a/radis/lbl/factory.py
+++ b/radis/lbl/factory.py
@@ -237,11 +237,14 @@ class SpectrumFactory(BandFactory):
         if ``True``, saves details of all calculated lines in Spectrum. This is
         necessary to later use :py:meth:`~radis.spectrum.spectrum.Spectrum.line_survey`,
         but can take some space. Default ``False``.
-    chunksize: int, or ``None``
+    chunksize: int, 'auto', or ``None``
         Splits the lines database in several chunks during calculation, else
         the multiplication of lines over all spectral range takes too much memory
-        and slows the system down. Chunksize let you change the default chunk
-        size. If ``None``, all lines are processed directly. Usually faster but
+        and slows the system down. Chunksize let you change the default chunk size.
+        If ``'auto'``, the optimal Chunksize is determined dynamically based on the
+        number of lines, grid points, and available system memory. This helps balance
+        performance and memory usage by ensuring a reasonable number of lines per chunk.
+        If ``None``, all lines are processed directly. Usually faster but
         can create memory problems. Default ``None``
     optimization : ``"simple"``, ``"min-RMS"``, ``None``
         If either ``"simple"`` or ``"min-RMS"`` LDM optimization for lineshape calculation is used:
@@ -695,6 +698,22 @@ class SpectrumFactory(BandFactory):
         self.params.lbfunc = lbfunc
 
         # used to split lines into blocks not too big for memory
+        if chunksize == "auto":
+            from sys import getsizeof
+
+            from psutil import virtual_memory
+
+            from radis.lbl.factory import _generate_broadening_range
+
+            n = 1
+            if isinstance(truncation, Default):
+                truncation = truncation.value
+            x = virtual_memory().available / getsizeof(
+                _generate_broadening_range(wstep, truncation)
+            )
+            chunksize = round(x * 10 ** (-np.floor(np.log10(x))), n - 1) * 10 ** (
+                np.floor(np.log10(x))
+            )
         self.misc.chunksize = chunksize
         # Other parameters:
         self.save_memory = save_memory


minouHub:
Hello,
I think the `chunksize` option is not available when using the LDM algorithm. The latter being the default algorithm I think this does not help most of RADIS usage. Also, when running radis on python version <= 3.10, we use vaex to virtually remove any limit to the memory usage. (There are on-going plans to make vaex compatible up to python 3.12, see https://github.com/radis/radis/pull/698). 

Please provide:
- [ ] an example demonstrating the reduction of memory usage on `python 3.11` or later
- [ ] an example demonstrating the reduction of memory usage on `python 3.10` or before (with vaex)

mohyware:
I saw it [here](https://github.com/radis/radis/issues/53) as a todo in performance section and thought it could be beneficial. But honestly even if its safer, the process feels slower with it, I see a progress bar, which doesn’t appear when just using the default `chunksize` 100,000. I’m not sure if the algorithm needs to be modified or if we should wait for full Vaex support. If this isn't critical, I can close it.

In python 3.11 or later:

https://github.com/user-attachments/assets/c57b93df-bff6-4f07-83c8-8e77f929e25c

minouHub:
Thanks for the demo. By the way you can print the computation time with verbose=3. 
So it seems that even on Python 3.11 this does not improve the performance. Is there a reason to even keep the chunksize option?

mohyware:
With auto chunksize (chunksize value became 30.000)
image
When chunksize is set to 1e7:
image

Without chunksize option:
image

I think it is for better memory handling as mention [here](https://github.com/radis/radis/blob/ada5c335c2bc53922232240a88dbc6a39e84f93d/radis/lbl/broadening.py#L2666)

minouHub:
Alright. Thanks for the computation. Can you demonstrate that the memory usage drops?

mohyware:
Sure, I'll try a larger one like CO₂. It will take some time since my internet isn't great.

minouHub:
You can try with CH4 which is already a large database but not as large as CO2

mohyware:
I don't see a reason to keep it as you mentioned.

I tried CH4 with:

- No chunksize option → Max resident set size : 578412 kB

- chunksize = 40000 → Max resident set size : 577912 kB

Both are (≈ 578 MB)

mohyware:
Execute the following code twice once with the option ``chunksize`` and once without  using this command:
```bash
/usr/bin/time -v python file.py
```
```python

sf = SpectrumFactory(
    2150,
    2450, 
    molecule="CH4",
    isotope="1",
    wstep=0.002,
    chunksize=1e7, # chunck option
)

sf.fetch_databank(
    source="hitemp"
) 

T = 1500.0  # K
p = 1.0  # bar
x = 0.8
l = 0.2  # cm
w_slit = 0.5  # cm-1


s_cpu = sf.eq_spectrum(
    name="CPU",
    Tgas=T,
    pressure=p,
    mole_fraction=x,
    path_length=l,
)
```

minouHub:
Did this on my side. Python 3.10 and using vaex.
```python
import psutil
import time

for chunksize in [5e6, 1e7, 1e8, 1e9]:
    start_time = time.time()
    sf = SpectrumFactory(
        2150,
        2250, 
        molecule="CH4",
        isotope="1",
        wstep=0.002,
        verbose = 0,
        chunksize=chunksize, # chunksize option
    )
    
    sf.fetch_databank(
        source="hitemp", #database="2010",
    ) 
    
    T = 1500.0  # K
    p = 1.0  # bar
    x = 0.8
    l = 0.2  # cm
    
    
    s_cpu = sf.eq_spectrum(
        name="CPU",
        Tgas=T,
        pressure=p,
        mole_fraction=x,
        path_length=l,
    )
    process = psutil.Process()
    print(f"-------\nChunk: {chunksize:.0e}.")
    print(f"Memory usage: {process.memory_info().rss / (1024 * 1024)} MB")
    print(f"Time to run: {time.time() - start_time:.2f} seconds")

```
```
-------
Chunk: 5e+06.
Memory usage: 504.75390625 MB
Time to run: 10.78 seconds
-------
Chunk: 1e+07.
Memory usage: 506.9453125 MB
Time to run: 5.75 seconds
-------
Chunk: 1e+08.
Memory usage: 537.90234375 MB
Time to run: 1.25 seconds
-------
Chunk: 1e+09.
Memory usage: 568.8046875 MB
Time to run: 0.80 seconds
```
Seems to me that the gain in memory not relevant. 
@erwanp @dcmvdbekerom do you think there is a need to do a new example? Or any new idea to use chunksize?

minouHub:
Closing ... as it seems that the implementation (although correct) does not bring significant improvement to the performance

PR #779, opened 2025-03-25
Minor fix 0.16 release

minouHub:
Improve doc and slightly change the example broadening in examples


HajimeKawahara:
@minouHub  What does the release schedule for v0.16 look like? I'd like to check if everything works properly with ExoJAX before the release.

PR #754, opened 2025-03-04
Add Vaex Support for Defining alpha_ref and n_Texp in ExoMol Broadening Coefficients

mohyware:
### Description
This pull request is to address #746

Add Vaex Support for Defining alpha_ref and n_Texp in ExoMol Broadening Coefficients
That's based on my understanding let me know if I missed anything

Fixes #746

Issue #658, opened 2024-06-20
Add fast-memory-mapping alternative(s) to vaex


minouHub:
### Description
Vaex is a fast memory mapping engine; better than Pandas for Dataframe management. However it is not available in `Python >=3.11` as of June 2024, (https://github.com/vaexio/vaex/issues/2397). **This issue/to-do proposes to implement an alternative to `vaex`.** 

The user would be able to select an engine at will taking in considerations what is compatible with its python version. It would be good for RADIS in terms of safety. It could also help to further increase the performances of RADIS if another package is better suited for our use.

### What has been done so far
_On the vaex side:_ 
- [ ] the vaex developer are working on it: https://github.com/vaexio/vaex/pull/2331

_On the RADIS side:_ 

- [ ] #656 allows to install RADIS with any Python version. However, if `vaex` is not compatible, the installation continues without vaex and `pytables` is the default engine.

### What are the alternatives

- [ ] [polar](https://pypi.org/project/polars/)               **===> preferred option <===**
Polars is a fast DataFrame library implemented in Rust and designed for high performance, with a strong focus on efficiency and speed. Thanks @ddelange for this suggestion in #656.

- [ ] [PySpark](https://spark.apache.org/docs/latest/api/python/)
PySpark is the Python API for Spark, a powerful open-source distributed computing system that can handle big data processing.

- [ ] [Dask](https://dask.org/)
Dask is a flexible parallel computing library for analytics that enables performance at scale. It integrates seamlessly with NumPy, pandas, and other Python libraries.

erwanp:
Just a few comments.

- PySpark and Dask are distributed librairies; i.e. they scale by using more nodes on clusters. On a single laptop (which is what most RADIS users use) they won't help. 

[...]

PR #786, opened 2025-03-27
Minor fix for 0.16 release

minouHub:
- Include some comments of @dcmvdbekerom 
- Remove tests of Cantera

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,9 +64,9 @@ after_success:
 
 
 stages: #This sequence defines the order of the stages
-  - name: Py39
-    if: branch = master OR branch = develop
   - name: Py311
+    if: branch = master OR branch = develop
+  - name: Py39
     if: branch = master OR branch = develop
   - name: Py310
     if: branch = master OR branch = develop
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,6 +64,8 @@ after_success:
 
 
 stages: #This sequence defines the order of the stages
+  - name: Py39
+    if: branch = master OR branch = develop
   - name: Py311
     if: branch = master OR branch = develop
   - name: Py310
@@ -71,8 +73,6 @@ stages: #This sequence defines the order of the stages
   - name: lint
     if: branch = master OR branch = develop
   - name: deploy
-    if: branch = master
-  - name: Py39
     if: branch = master
   - name: Py312
     if: branch = master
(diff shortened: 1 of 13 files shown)


minouHub:
Ok so there are still problems with Python 3.9 and some exomol queries

minouHub:
@erwanp and @HajimeKawahara I'm quite confused by the difference between the arguments `output` and `engine` when dealing with pandas and vaex Dataframes. I had trouble passing the tests because of that, see https://github.com/radis/radis/pull/786/commits/779ad500dafa70bdf37a295b7715d7efa498c199
I understand that `output` will set the type of `output`, but why do we need situations where `output!=engine`? Could you explain the difference?

minouHub:
Still on the question of `output` vs `engine`, with [1d3ccfb](https://github.com/radis/radis/pull/786/commits/1d3ccfbd3b871440c65be663d05c5627515486b5) I feel that I'm cancelling all the benefits of #754 because even on my own Python 3.10 version, I have `output==pandas`

minouHub:
I did something wrong with pytest and Travis, see #794

minouHub:
Done in #796

erwanp:
"Engine" is how the DataFrame is stored on your computer ; Output is how it's returned in your Python console. 

For instance you can have Dataframes stored with Vaex format (which uses Memory-mapping to loaded almost instantly in your RAM, however not all Pandas API functions are available in Vaex; which is a different DataFrame engine just like Polars is) ;  and the calculations within Radis can be done using Pandas (using Pandas API functions, groupby() functions for instance). So you would have

```
engine='vaex'
output='pandas'
```

For a long time, the option above was the only option. To be able to Calculate with Vaex took a long time and was introduced in Radis 0.15 : (beforehand, only loading of database was performed with Vaex) : https://github.com/radis/radis/pull/576, https://github.com/radis/radis/pull/580 . Then you could have: 

```
engine='vaex'
output='vaex'
```

But in Recent Python versions we know that Vaex is not available (see https://github.com/radis/radis/issues/658), then you must have: 

```
engine='pandas'
output='pandas'
```

Hope it's more clear. These important details might need be added somewhere in the function documentations.

minouHub:
If vaex is available I do not see the point of doing a calculation with pandas. Is there a problem with enforcing output == engine?

Issue #781, opened 2025-03-25
H2O HITEMP download error?


HajimeKawahara:
### 🐛 Describe the bug

I am currently reviewing for release v1.6 #780. In the current develop version, CO and CH4 from HITEMP can be loaded, but the following error occurs with H2O.

```sh
Traceback (most recent call last):
  File "/home/kawahara/exojax/tests/integration/moldb/mdbhitemp_test_h2o.py", line 55, in <module>
    test_moldb_hitemp()
  File "/home/kawahara/exojax/tests/integration/moldb/mdbhitemp_test_h2o.py", line 7, in test_moldb_hitemp
    mdb = MdbHitemp(".database/H2O/01_HITEMP2010/",
  File "/home/kawahara/exojax/src/exojax/spec/api.py", line 757, in __init__
    local_files, urlnames = self.get_filenames()
  File "/home/kawahara/anaconda3/lib/python3.10/site-packages/radis/api/dbmanager.py", line 224, in get_filenames
    urlnames = self.fetch_urlnames()
  File "/home/kawahara/anaconda3/lib/python3.10/site-packages/radis/api/hitempapi.py", line 532, in fetch_urlnames
    with open(
FileNotFoundError: [Errno 2] No such file or directory: '/home/kawahara/anaconda3/lib/python3.10/site-packages/radis/db/H2O/HITRANpage_january2025.htm'
```

The code is as follows:

```python
from exojax.spec.api import MdbHitemp
mdb = MdbHitemp(".database/H2O/01_HITEMP2010/",
                    nurange=[4200.0, 4300.0],
                    crit=1.e-30,
                    isotope=None,
                    inherit_dataframe=True)
```

### 💡 Possible solutions

@sh-tada suggests that the issue might be partly due to the absence of `__init__.py` in `radis/db/H2O`?

### 🎲 Radis version

0.16 (develop)

### 💻 Operating system

Linux

HajimeKawahara:
This is not directly related, but is it intentional that the HITEMP files are downloaded every time, even when they already exist?

minouHub:
Indeed, I got a similar problem 
```python
from radis import calc_spectrum
s = calc_spectrum(100, 110,         # cm-1
                  molecule='H2O',
                  isotope='1',
                  pressure=1.01325,   # bar
                  Tgas=1000,
                  mole_fraction=0.1,
                  databank='hitemp',  # or 'hitemp'
[...]

PR #790, opened 2025-03-29
Update hitempapi.py

prabhat-kumar96:
#781 

Fixed HITEMP Download Issues

Now requires HITRAN login - no more HTML errors

Uses direct 2024 download URLs - no HTML scraping

Better error messages - clear auth/download failures

Works for all molecules (H₂O, CO₂, CH₄ etc.)

User Action Needed:
Run setup_credentials() first to save HITRAN username/password.

### Description

Fixes #<Issue Number>


prabhat-kumar96:
I will merge every commit into one

PR #795, opened 2025-03-31
Fix typo: Change "username" to "email" in HITRAN authentication

mohyware:
### Description
HITRAN requires an email for authentication, but it was referred to it as "username" which was misleading. This change avoid confusion.

I myself first used the HITRAN database. I searched for a "username" field on the HITRAN website, only to realize later that it actually asks for the email.

### DIFF
--- a/radis/api/hitempapi.py
+++ b/radis/api/hitempapi.py
@@ -133,7 +133,7 @@ def _prompt_password(user):
     Parameters
     ----------
     user : str
-        Username.
+        Email.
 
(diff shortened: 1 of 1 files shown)


mohyware:
Oh, this caused doc checks to fail.
image

minouHub:
Hello, 
This is a sound argument but rather low priority issue. You can focus on other PR

mohyware:
Sure! no problems.

PR #799, opened 2025-04-03
Fixing kwargs Handling Using Plotly's Validation

### LINKED ISSUE #760: `Spectrum.plot`'s kwargs are not being pass down to `plotly` when it is used as backend

darshvn:
### Description

This pull request addresses issue #760. After I highlighted [issues](https://github.com/radis/radis/pull/762#issuecomment-2711387021) in PR #762, @erwanp suggested to choose more general approach to handling `kwargs` [here](https://github.com/radis/radis/pull/762#issuecomment-2764460511). I believe this method is a better approach because, This method leverages Plotly's built-in validation for parameter handling, automatically supports all Plotly parameters, eliminates the need for maintaining parameter mappings, and ensures clear expectations for parameter usage.

Fixes #760


minouHub:
@erwanp reminder

minouHub:
Closing as this issue was solved by #762

Issue #804, opened 2025-04-08
Auto deploy on Pypi is broken


erwanp:
### 🐛 Describe the bug

setup.py was removed with 0.16  #734

Therefore the Deploy pipeline is broken : https://github.com/radis/radis/blob/develop/.travis.yml#L150 

### 💡 Possible solutions

An alternative is to use Build, see https://stackoverflow.com/questions/58753970/how-to-build-a-source-distribution-without-using-setup-py-file

```
pip install build
python -m build --sdist --wheel
```

It will be required to include this in our Continuous integration Pipeline

In the meantime, a temp fix is to ask @erwanp  to do it manually and deploy from a local installation. This was done successfully for 0.16 https://pypi.org/project/radis/
. Manual upload is done with  

```
twine upload dist/* --repository radis
``` 

### 🎲 Radis version

0.16

### 💻 Operating system

Windows

erwanp:
Tempfix @minouHub @dcmvdbekerom @HajimeKawahara 0.16 was succesfully deployed manually and is available from Pypi https://pypi.org/project/radis/

dcmvdbekerom:
The description is missing atm https://pypi.org/project/radis/0.16/

erwanp:
indeed. We'll deploy it with 0.16.1 after a few weeks and some feedbacks on this version

minouHub:
Agreed

pierrelux:
Probably related to this: 

```
pip show radis
Name: radis
Version: 0.16
```

Getting error due to tests missing: 

```
% /path/to/.venv/bin/python /path/to/project/src/simulation/lbl.py
Traceback (most recent call last):
  File "/path/to/.venv/lib/python3.13/site-packages/radis/lbl/base.py", line 80, in <module>
    from .loader import KNOWN_LVLFORMAT, DatabankLoader, df_metadata
  File "/path/to/.venv/lib/python3.13/site-packages/radis/lbl/loader.py", line 87, in <module>
    from radis.levels.partfunc_cdsd import PartFuncCO2_CDSDcalc, PartFuncCO2_CDSDtab
  File "/path/to/.venv/lib/python3.13/site-packages/radis/levels/partfunc_cdsd.py", line 41, in <module>
    from radis.test.utils import getTestFile
ModuleNotFoundError: No module named 'radis.test.utils'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/path/to/project/src/simulation/lbl.py", line 1, in <module>
[...]

PR #805, opened 2025-04-11
Fix PyPi autodeploy for 0.16

minouHub:
### Description
Remove the need to import a function from a test file. As of version 0.16, the tests files are no longer deployed in the user version via PyPi.

This pull request is to address #804 

@erwanp please review as you were the one who asked for the line that was posing problem. Also put a little bibliographic joke/correction for you in the code


minouHub:
Success on a test version 0.16.1 on test.pypi https://test.pypi.org/project/radis/

PR #809, opened 2025-04-14
Fix NaN issue for Nblower 

### LINKED ISSUE #807: Error when downloading files for AlO and CrH from ExoMol
### 🐛 Describe the bug

I am using `radis==0.15.2` through `ExoJAX==2.0.0`. When I try to download data for molecules like AlO from ExoMol, the following error occurs. I confirmed that this error happens not only with AlO but also with CrH.
```
  File "/home/yuikasagi/.pyenv/versions/exojax2/lib/python3.11/site-packages/exojax/spec/multimol.py", line 117, in multimdb
    api.MdbExomol(
  File "/home/yuikasagi/.pyenv/versions/exojax2/lib/python3.11/site-packages/exojax/spec/api.py", line 190, in __init__
    self.activate(df)
  File "/home/yuikasagi/.pyenv/versions/exojax2/lib/python3.11/site-packages/exojax/spec/api.py", line 301, in activate
    self.set_broadening_coef(df[mask], add_columns=False)
  File "/home/yuikasagi/.pyenv/versions/exojax2/lib/python3.11/site-packages/radis/api/exomolapi.py", line 1340, in set_broadening_coef
    j2alpha_ref, j2n_Texp = make_j2b(
                            ^^^^^^^^^
  File "/home/yuikasagi/.pyenv/versions/exojax2/lib/python3.11/site-packages/radis/api/exomolapi.py", line 672, in make_j2b
    j2alpha_ref = np.full(Nblower, alpha_ref_default)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/yuikasagi/.pyenv/versions/exojax2/lib/python3.11/site-packages/numpy/core/numeric.py", line 329, in full
    a = empty(shape, dtype, order)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: expected a sequence of integers or a single integer, got '157.5'
```

The minimal code that causes this error is as follows:
```
from exojax.spec.multimol import MultiMol

# setting molecule & path to database
mols = [["AlO"]] 
db = [["ExoMol"]]
path_data = "./.database"

# get database
nu_grid, _, _ = wavenumber_grid(x0=14000., x1=15500., N=5000, 
                                unit="AA", xsmode="premodit", wavelength_order="ascending")
nu_grid_list = [nu_grid]
mul = MultiMol(molmulti=mols, dbmulti=db, database_root_path=path_data)
multimdb = mul.multimdb(nu_grid_list, crit=1.e-27, Ttyp=1000.) # <-- failed when mols = [["AlO"]] and [["CrH"]]
```

### 💡 Possible solutions

_No response_

### 🎲 Radis version

0.15.2

### 💻 Operating system

Linux

mohyware:
### Description

This pull request is to Fix NaN issue for Nblower in make_j2b.

Follow-up to: #754 

Fixes #807

### DIFF
--- a/radis/api/exomolapi.py
+++ b/radis/api/exomolapi.py
@@ -692,6 +692,9 @@ def make_j2b(
     else:
         Nblower = np.max([jlower_max, np.max(jlower_arr)]) + 1
 
+    if np.isnan(Nblower):
+        raise ValueError("Invalid value encountered: 'Nblower' is NaN.")
+
     if output == "vaex":
         # Initialize arrays with default alpha_ref and n_Texp
         df_defaults = vaex.from_arrays(
@@ -711,8 +714,8 @@ def make_j2b(
         j2n_Texp = df_defaults["n_Texp"].values
     else:
         # Initialize arrays with default alpha_ref and n_Texp
-        j2alpha_ref = np.full(Nblower, alpha_ref_default)
-        j2n_Texp = np.full(Nblower, n_Texp_default)
+        j2alpha_ref = np.full(int(Nblower), alpha_ref_default)
+        j2n_Texp = np.full(int(Nblower), n_Texp_default)
 
         # Populate the mapping arrays using known broadening coefficients
         j2alpha_ref[jlower_arr] = alpha_ref_arr
@@ -1514,8 +1517,8 @@ class MdbExomol(DatabaseManager):
                     jlower_max=df["jlower"].max(),
                     output=output,
                 )
-                self.alpha_ref = j2alpha_ref[df["jlower"].values]
-                self.n_Texp = j2n_Texp[df["jlower"].values]
+                self.alpha_ref = j2alpha_ref[df["jlower"].values.astype(int)]
+                self.n_Texp = j2n_Texp[df["jlower"].values.astype(int)]
             elif codelv == "a1":
                 j2alpha_ref, j2n_Texp = make_j2b(
                     bdat,
--- a/radis/api/exomolapi.py
+++ b/radis/api/exomolapi.py
@@ -688,12 +688,9 @@ def make_j2b(
 
     # Determine the array size based on jlower_max
     if jlower_max is None:
-        Nblower = np.max(jlower_arr) + 1
+        Nblower = np.nanmax(jlower_arr) + 1
     else:
-        Nblower = np.max([jlower_max, np.max(jlower_arr)]) + 1
-
-    if np.isnan(Nblower):
-        raise ValueError("Invalid value encountered: 'Nblower' is NaN.")
+        Nblower = np.nanmax([jlower_max, np.max(jlower_arr)]) + 1
 
     if output == "vaex":
         # Initialize arrays with default alpha_ref and n_Texp
--- a/radis/lbl/broadening.py
+++ b/radis/lbl/broadening.py
@@ -1818,7 +1818,7 @@ class BroadenFactory(BaseFactory):
         # ------------------------------------
 
         def _init_w_axis(w_dat, log_p):
-            w_min = w_dat.min()
+            w_min = np.nanmin(w_dat)
             if w_min == 0:
                 self.warn(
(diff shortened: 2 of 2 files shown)


radis/api/exomolapi.py
```diff
@@ -692,6 +692,9 @@ def make_j2b(
     else:
         Nblower = np.max([jlower_max, np.max(jlower_arr)]) + 1
 
+    if np.isnan(Nblower):
+        raise ValueError("Invalid value encountered: 'Nblower' is NaN.")
+
```
minouHub, line 697: Shouldn'we use `np.nanmax` instead of `np.max`?

minouHub (commented):
The general comment is more about how we should adapt the code in case there is a non-integer value in the broadening file. In the example of @YuiKasagi in #809, NBlower is 157.5. Should we have the code assuming this should be 157 or 158? Maybe @YuiKasagi has an idea?

minouHub (commented):
Still got an error running
```python

s= calc_spectrum(1900, 2300,         # cm-1
                  molecule='AlO',
                  isotope=1,
                  pressure=1.01325,   # bar
                  Tgas=1000,
                  mole_fraction=0.1,
                  databank='exomol',  # or 'hitemp'
                  )
```
```

  File ~\anaconda3\envs\radis_py310\lib\site-packages\spyder_kernels\py3compat.py:356 in compat_exec
    exec(code, globals, locals)

  File c:\users\nicolas minesi\python\radis\test\levels\sanstitre1.py:10
    s= calc_spectrum(1900, 2300,         # cm-1

  File ~\Python\radis\lbl\calc.py:586 in calc_spectrum

  File ~\Python\radis\lbl\calc.py:967 in _calc_spectrum_one_molecule
    s = eq_spectrum_mode(

  File ~\Python\radis\lbl\factory.py:898 in eq_spectrum
    wavenumber, abscoeff_v = self._calc_broadening()

  File ~\Python\radis\lbl\broadening.py:2969 in _calc_broadening
    (wavenumber, abscoeff) = self._broaden_lines(df)

  File ~\Python\radis\lbl\broadening.py:2639 in _broaden_lines
    ) = self._calc_lineshape_LDM(df)

  File ~\Python\radis\lbl\broadening.py:1847 in _calc_lineshape_LDM
    wL = _init_w_axis(wL_dat, log_pL)  # FWHM

  File ~\Python\radis\lbl\broadening.py:1835 in _init_w_axis
    return w_min * np.exp(log_p * np.arange(N))

ValueError: arange: cannot compute length
```

YuiKasagi:
Thank you for the fix! @mohyware 
I've confirmed that the problem described in #809 has been solved and the code now works properly.
At least in this code, Nblower is only used for setting indices, so converting actual jlower values such as [0.5, 1.5, …] into integers [0, 1, …] seems like a good idea.

YuiKasagi:
In my environment, I was also able to reproduce the error commented by @minouHub. However, I found that it does not occur for molecule='CrH' (after changing np.max to np.nanmax).
The issue likely arises from the following part:
https://github.com/radis/radis/blob/54a27e1799c35c59dda57c6bebde681a673c54ae/radis/api/exomolapi.py#L1579-L1591
because in the data from .broad file, bdat["jlower"] = 0.5 does not exist. As a result, the values of alpha_ref and n_Texp corresponding to df["m"] = 0.5 to become NaN. (For CrH, this wasn't a problem as its jlower values are all ≥ 11.5.)

I am unsure why m=0.5 is not included in the .broad file, and at the moment, I'm not certain how to resolve this...

mohyware:
@minouHub  @YuiKasagi 

Try it now it seems **nan** functions fixed it.

minouHub:
> At least in this code, Nblower is only used for setting indices, so converting actual jlower values such as [0.5, 1.5, …] into integers [0, 1, …] seems like a good idea.

Yes you're right. Sorry for the confusion.

PR #823, opened 2025-05-13
removed test module import in __init__.py

HajimeKawahara:
### Description

This pull request aims to remove the test module import from `__init__.py` because the tests were removed from the distributed package.


minouHub:
Hello,
Rebased on develop and moved function `spectrum_test` outside of the `radis\test` folder

PR #831, opened 2025-06-13
Fixes `PytestReturnNotNoneWarning` by modifying tests to use `np.allclose` and not return True 

### LINKED ISSUE #830: `PytestReturnNotNoneWarning`: Expected None but returning True
### 💭 Description

Pytest had been warning about this, but now it's directly throwing errors in the CI, so we have to refactor the testing infrastructure first. 
```
 PytestReturnNotNoneWarning: Expected None, but radis/test/test_misc.py::test_nantrapz returned 44.5, which will be an error in a future version of pytest.  Did you mean to use `assert` instead of `return`?
```
```
FAILED radis/test/test_misc.py::test_utils - Failed: Expected None, but test returned True. Did you mean to use `assert` instead of `return`?
FAILED radis/test/test_misc.py::test_progress_bar - Failed: Expected None, but test returned True. Did you mean to use `assert` instead of `return`?
FAILED radis/test/test_misc.py::test_nantrapz - Failed: Expected None, but test returned np.float64(44.5). Did you mean to use `assert` instead of `return`?
```
@minouHub @erwanp, I would like to know about this. Maybe I could open a PR, and we can all review it?

Prtm2110:
Fixes #830

### DIFF
--- a/docs/spectrum/spectrum.rst
+++ b/docs/spectrum/spectrum.rst
@@ -188,9 +188,9 @@ Use :py:meth:`~radis.spectrum.spectrum.Spectrum.from_hdf5`  ::
 Calculate a test spectrum
 -------------------------
 
-You need a spectrum immediatly, to run some tests ? Use :py:func:`~radis.test.utils.test_spectrum` ::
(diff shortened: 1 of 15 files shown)


minouHub (approved):
Fine by me!

PR #852, opened 2025-07-04
Accelerate CI tests 

minouHub:
### Description
@dcmvdbekerom I changed the molecule from CO2 to CO to avoid downloading this large database. Is it ok for you?


minouHub:
The longest test is now `test_abscoeff_continuum`. Is it finally time to deprecate this feature and remove the test? @erwanp ?

```
44.86s call     radis/test/lbl/test_broadening.py::test_abscoeff_continuum
```

PR #854, opened 2025-07-07
Bug: Fixes mol is not passed down to `hit2df`

### LINKED ISSUE #849: `ValueError: Cant cast data to specific dtype` when `hit2df` is parsing HITEMP2 CO2
### 🐛 Describe the bug

For the record.
https://github.com/Prtm2110/radis/blob/0c7450e028cd737654e3795ce6008e60ad5d07c9/radis/api/tools.py#L126 is causing error because `iso` is `int` for other molecules but for `CO2` it's string and the check below is not being executed as which molecule is passed down is not transferred
 ```
if molecule == "CO2":
        dtype[1] = (
            "iso",
            "<U1",
        )
``` 
I have fixed this bug in https://github.com/radis/radis/pull/833/commits/8731e1bf702102766a1a1d62d5d3e8379dfe5eff 

### 💡 Possible solutions

_No response_

### 🎲 Radis version

Editable

### 💻 Operating system

Linux

Prtm2110:
Fixes #849

### DIFF
--- a/radis/api/hitranapi.py
+++ b/radis/api/hitranapi.py
@@ -292,8 +292,7 @@ def hit2df(
         raise ValueError(
             f"You're trying to read a binary file {fname} instead of an HITRAN file"
         ) from err
-
-    df = parse_hitran_file(fname, columns, output=output)
+    df = parse_hitran_file(fname, columns, output=output, molecule=mol)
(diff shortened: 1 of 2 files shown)


Prtm2110:
Thank you, it's good to go from my end!

PR #834, opened 2025-06-16
Solving multiple minor issues

minouHub:
### Description
Multiple issues solved here.

- [ ] HITRAN login required for other databases
- [ ] Missing txt files needed for atomic databases

Fixes #712

PR #855, opened 2025-07-14
Fix HITEMP detection by checking urlname.lower()

sh-tada:
### Description

Fixes a bug in HITEMP download.   
Related to #834.

I encountered the following error when trying HITEMP download:
```
  File ".../transmission.py", line 89, in <module>
    mdb = MdbHitemp(path, nu_grid, gpu_transfer=False, isotope=1, engine="vaex")
  File ".../exojax/database/api.py", line 776, in __init__
    self.download_and_parse(download_urls, download_files)
  File ".../radis/api/dbmanager.py", line 508, in download_and_parse
    download_and_parse_one_file(urlname, local_file, Ndownload)
  File ".../radis/api/dbmanager.py", line 416, in download_and_parse_one_file
    if "hitemp" in self.local_databases:
TypeError: argument of type 'PosixPath' is not iterable
```

This happens because `self.local_databases` is a `Path` object, not a string or list.

Changing `self.local_databases` to `str(self.local_databases)` would resolve the immediate error,  
but using `urlname.lower()` seems more appropriate for detecting HITEMP.

### DIFF
--- a/radis/api/dbmanager.py
+++ b/radis/api/dbmanager.py
@@ -413,7 +413,7 @@ class DatabaseManager(object):
                 )
 
             # Download file with requests
-            if "hitemp" in self.local_databases:
+            if "hitemp" in urlname.lower():
                 # Get session from HITEMP API
                 from radis.api.hitempapi import login_to_hitran
 


minouHub:
Hello,
Sorry for that, I think I introduced this bug. No problem for me. Should it be merged urgently?

PR #858, opened 2025-08-21
Add C++ SIMD Parser 

### LINKED ISSUE #510: 🚀 Feature: write a Numba parser for HITEMP Files
### 🔖 Feature description

Parsing HITRAN/HITEMP file is currently very slow. 

#505 improved it a bit for HITRAN files, using the HAPI implementation which is faster than RADIS's initial implementation. 
However, HITEMP is unchanged, and remains a large bottleneck. 

We use `np.fromfile` which is very slow. 
A detailed study was done few months ago : https://stackoverflow.com/questions/71411907/dramatic-drop-in-numpy-fromfile-performance-when-switching-from-python-2-to-pyth/71505529#71505529 

Turns out that a pure-Python implementation of the parser, accelerated by Numba, could be 1,000 faster than the `np.fromfile` approach 

---

### Implementation

Re-write the [_read_hitran_file()](https://github.com/radis/radis/blob/e85cb96b6984c0f60e9362df63c15b81221489a2/radis/io/tools.py#L120) function with a pure Python, then jit-Numba it. 

We currently spend ~3hrs parsing HITEMP CO2 & H2O files (!)
Could really help to speed this up.

prathamhole14:


Prtm2110:
> @Prtm2110 if you think this can be merged within 1-2 weeks, this will make it in version 0.16.4

Yes sure I will try best to get this merge ready in a week or so!

minouHub:
hello @Prtm2110 are you still working on this?

prathamhole14:
Closing this since #904 is fast enough considering the maintainability of this code.

PR #859, opened 2025-08-26
New example on different broadening mechanisms

minouHub:
### Description

This pull request provides graphs similar to those produced in the paper describing the LDM optimization, see https://doi.org/10.1016/j.jqsrt.2020.107476


minouHub:
@erwanp @dcmvdbekerom 
I changed the name of the broadening method "voigt" to "voigt_poly" because it is a polynomial approximation. Could I have your opinion on this?
Also @erwanp, could you make "more appropriate" more explicit?

> Because the LDM convolves all lines at the same time,
        and thus operates on large arrays, ``'fft'`` becomes more appropriate than
        convolutions in real space (``'voigt'``, ``'convolve'`` )

dcmvdbekerom:
> I changed the name of the broadening method "voigt" to "voigt_poly" because it is a polynomial approximation. Could I have your opinion on this?

I think it's a good idea

>>and thus operates on large arrays, 'fft' becomes more appropriate than

More appropriate = faster

dcmvdbekerom:
Hello, I wanted to comment on the discrepancy between convolve and fft: it is due to fft not taking into account truncation. Because for FFT the lineshapes are computed in FT space, it would be counterproductive to implement truncation here: it is a multiplication by a rectangle, in other words a convolution with a sinc in FT space, which we are trying to avoid in order to be faster than convolve.

Easiest solution would be that if truncation and fft are used, a userwarning is printed that trunc is ignored and with the suggestion to use convolve instead

dcmvdbekerom (approved):

minouHub:
Thank you Dirk!

Remaining to-do
- [x] Add explanation for the differences between fft and convolution
- [x] More appropriate = faster

minouHub:
Travis failing but sorted on develop branch. Merging

PR #860, opened 2025-08-26
Add multigrid2208 rebased august25 unbuggued

erwanp:
### Description

Fixes #<Issue Number>


radis/lbl/broadening.py
```diff
@@ -1895,8 +1940,30 @@ def _init_w_axis(w_dat, log_p):
...
+            ### ============> Replacement by Nicolas
+            if wavenumber_group is not None:  # i.e. if sparse
+                raise NotImplementedError(
+                    "broadening_method == 'fft'. Only broadening_method='voigt' is implemented with Multigrids"
+                )
+            else:  # normal, like before
+                w = self.wavenumber_calc
+                wstep = self.params.wstep
+            ### ============> Replacement by Nicolas
+
```
minouHub, line 1966: @erwanp please confirm that my fix makes sense. Thanks

minouHub:
@dcmvdbekerom my tests in `test_sparse.py` indicate that the fastest optimization mode with `SPARSE_WAVEGRID=multi_sparse_grid` is `optimization=None` and `broadening_method=Voigt_poly`. I started implementing changes using this configuration by default. Using another `optimization` or `broadening_method` raises a `ValueError`.

In a way, I like to not give too many options for the user. There is one technique to generate a spectrum, and that's it. Rarely, an advanced user can play with the parameters `optimization`, `broadening_method`, and `SPARSE_WAVEGRID` but it remains marginal.

Still have a lot of tests to resolve induced by the change of the default parameter but let me know what you think.

PR #862, opened 2025-08-27
GitHub actions for CI and linting

minouHub:
### Description

This pull request adds GitHUub Actions to RADIS! The tests are as fast as on Travis and for the moment it seems that the credits are unlimited.

### What's new compared to Travis

- **Unlimited credits?**
- **GPU cannot be tested freely on GitHub actions.** Thus, we will need to keep Travis but only for GPU tests. Are you good with that @dcmvdbekerom?
- **Conda is not employed for the install.** RADIS is installed via `pip install -e .`. Are you ok with that @erwanp? It makes the install way faster. Conda leads to a few problems that I did not bother to solve, see https://github.com/minouHub/radis/pull/1

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -1,0 +1,24 @@
+name: Linting with pre-commit
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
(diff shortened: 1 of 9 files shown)


minouHub:
Euh... Merged by mistake

PR #869, opened 2025-09-19
[hot fix] Restriction to numpy<2 for vaex compatibility

minouHub:
The `vaex-core` package is now compatible with `numpy 2.0` which is great news! More info in https://github.com/vaexio/vaex/pull/2449 and https://github.com/vaexio/vaex/pull/2466
Thus, in an empty environment, `numpy 2.0` is now installed with `radis`. However, `vaex-hdf5` is still not compatible with `numpy 2.0` and this incompatibility is not in the requirements of `vaex-hdf5`. This hot fix imposes `numpy < 2.0` to make sure vaex-hdf5 still works.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -23,7 +23,8 @@ dependencies:
 - joblib  # for parallel loading of SpecDatabase
 - lmfit  # for new fitting modules
 - matplotlib
-- numpy
+- numpy ; python_version > '3.10' # numpy 2.0 is not compatible with vaex-hdf5, see https://github.com/vaexio/vaex/issues/2468
+- numpy<2 ; python_version < '3.11'
 - numba  # just-in-time compiler
 - pandas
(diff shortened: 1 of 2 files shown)


minouHub:
@erwanp and @dcmvdbekerom requesting your review for this fix. The numpy 2.0 fix is critical for python 3.10 configurations. Hope to deploy it soon.

minouHub:
Following discussion with @pmariotto, it seems dangerous to 

- In the conda loop, install everything compatible with `numpy 2.0`
- In the pip loop, install `numpy<2.0` but keeping other packages the same. For instance, `scipy` installed in conda loop might require `numpy 2.0`

I'm going to do another attempt forcing `numpy<2.0` for any installation of radis.

HajimeKawahara:
@minouHub Thank you! Yes `exojax` will be compatible with `numpy>=2.0` once `radis` supports it.

PR #885, opened 2025-11-14
Added parallel CO2 chunk downloader with ThreadPoolExecutor

### LINKED ISSUE #881: Download chunks parallelly for HITEMP CO2
### 🔖 Feature description

In #833 partial download was implemented for HITEMP CO2, where each chuck is downloaded one at a time, this can be be done in parallel if `number of chunks are > 1` . 

_No response_

merintheres:
Fixes  #881 

Adds multithreaded CO₂ HITEMP download support in `hitempapi.py` using `ThreadPoolExecutor` for faster processing of multiple chunks.
- Added `_download_single_chunk()` helper for cleaner logic.
- Enabled parallel downloads (max 4 threads).
- Falls back to single-thread for one chunk.
- Maintains existing cache, verbose, and error handling behavior.


Prtm2110:
Hi thank you for the PR, it is working and I have few suggestions to make it better 
- [x] Move `_download_single_chunk` outside `read_and_write_chunked_for_CO2` this will make the code cleaner 
- [x] The progress bars are not showing up for individual chuck downloads

radis/api/hitempapi.py
```diff
@@ -690,38 +725,44 @@ def _append_dataframe(df_to_append):
...
+                        out_file,
+                        session,
+                        engine,
+                        verbose,
+                        idx,
+                    )
+                    for idx, (start, end, out_file) in enumerate(files_to_download)
+                ]
+                for future in as_completed(futures):
+                    future.result()
```
Prtm2110: Throw and show the error, it will be easier for debugging.

minouHub:
As of now (https://github.com/radis/radis/commit/d8c81956841ac1c47dabc6f5bf55f84fecc86560), the CI should pass.

radis/api/hitempapi.py
```diff
@@ -611,6 +612,40 @@ def parse_one_CO2_block(
     return df
 
 
+def _download_single_chunk(
+    start_wavno, end_wavno, out_file, session, engine, verbose=True, position=0
+):
+    """Download and process a single CO2 chunk."""
```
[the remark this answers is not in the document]
Prtm2110: Since it's a private function no need.

radis/api/hitempapi.py
```diff
@@ -611,6 +612,40 @@ def parse_one_CO2_block(
...
+    try:
```
Prtm2110: I think it would make more sense to wrap only `partial_download_co2_chunk` in try expect block.

merintheres:
@Prtm2110 , I noticed the parallel CO₂ downloader’s tqdm progress bars become unstable during multithreaded downloads. Rich’s Progress provides clean, thread-safe multi-task bars. Would you be open to switching the parallel-download progress bars from tqdm to Rich?

Prtm2110:
I meant the individual progress bar for each chuck download which is inside the `partial_download_co2_chunk` function (it is already been implemented in `get_bz2`). You can turn it on by setting the internal individual download versobe to true. 
Here: ` verbose=False,  # Suppress verbose to avoid cluttered output`. 
That's all we require for now as its already implemented and working no need to use `rich`.

merintheres:
Understood, thanks @Prtm2110. I’ll enable the existing per-chunk progress bar by setting the internal verbose=True and update the PR.

Prtm2110:
Hey @merintheres thanks for the changes they look really good to me, it's ready to merge from my end. Just add a test (or modify the one which already exists), this will not increase testing time by much as github hosted runners have at least 2 cores.     
Note: Test should run 2 chunks.

merintheres:
Thanks @Prtm2110 !  
I’ve modified the existing test to trigger a 2-chunk CO₂ partial download.

radis/test/io/test_hitemp.py
```diff
@@ -241,17 +241,19 @@ def test_fetch_hitemp_partial_download_CO2(verbose=True, *args, **kwargs):
...
+        load_wavenum_max=2500,
         verbose=3,
         return_local_path=True,
     )
 
     assert df.shape[1] == 24
-    assert len(local_files) == 1
-    assert basename(local_files[0]).startswith("CO2_02_00000-00261_HITEMP2024.")
+    assert len(local_files) == 2, f"Expected 2 files, got {len(local_files)}"
+    for f in local_files:
```
Prtm2110: I think we should explicitly check for both file names so that we know it is serving the correct ones.

Prtm2110 (changes requested):

radis/test/io/test_hitemp.py
merintheres: I’ve updated the test to explicitly check both expected chunk filenames.

Prtm2110 (approved):

radis/test/io/test_hitemp.py
Prtm2110: Does it require sorting? I thought they are already sorted or is it because they are now being processed in parallel now. If there is no need for sorting then please remove it.
Thanks!!

merintheres: I wasn’t sure if local_files is always returned in a fixed order when downloads run in parallel, so I added sorting for safety.

radis/test/io/test_hitemp.py
```diff
@@ -241,17 +241,25 @@ def test_fetch_hitemp_partial_download_CO2(verbose=True, *args, **kwargs):
...
-    assert basename(local_files[0]).startswith("CO2_02_00000-00261_HITEMP2024.")
+    assert len(local_files) == 2, f"Expected 2 files, got {len(local_files)}"
+    expected_files = [
+        "CO2_02_02273-02446_HITEMP2024.par",
+        "CO2_02_02446-02659_HITEMP2024.par",
+    ]
+    actual_files = sorted([basename(f) for f in local_files])
+    assert actual_files == sorted(
+        expected_files
+    ), f"Expected files {expected_files}, got {actual_files}"
```
Prtm2110:
```suggestion
    actual_files = [basename(f) for f in local_files]
    assert actual_files == expected_files, f"Expected files {expected_files}, got {actual_files}"
```

radis/api/hitempapi.py
```diff
@@ -611,6 +612,32 @@ def parse_one_CO2_block(
     return df
 
 
+def _download_single_chunk(
+    start_wavno, end_wavno, out_file, session, engine, verbose=True
+):
+    """Download and process a single CO2 chunk."""
```
Prtm2110:
```suggestion
    """Download a single CO2 chunk."""
```

radis/test/io/test_hitemp.py
Prtm2110: They are returning in the fix order so therefore you can remove the sorted.

merintheres: Got it, thanks for clarifying. I’ll remove the sorting logic.

Prtm2110:
Thanks a lot @merintheres it's merge ready from my side, @minouHub please have a look. Thanks!

radis/api/hitempapi.py
```diff
@@ -611,6 +612,40 @@ def parse_one_CO2_block(
...
+            pbar.update(1)
+        if verbose:
+            print(f"Completed chunk {start_wavno:.0f}-{end_wavno:.0f} cm⁻¹")
+        return True
+    except Exception as e:
+        if verbose:
+            print(
+                f" Error downloading chunk {start_wavno:.0f}-{end_wavno:.0f} cm⁻¹: {e}"
+            )
+        raise
```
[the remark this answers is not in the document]
minouHub: I agree that you should keep the initial error raised. In the current status of the code, the error message is empty.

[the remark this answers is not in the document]
minouHub: I'm honestly not sure the print message is needed. There is already one print before the processing of the chunk. Hence, you could probably do without the `try`

radis/api/hitempapi.py
```diff
@@ -695,33 +722,40 @@ def _append_dataframe(df_to_append):
...
-
-            partial_download_co2_chunk(
-                start_wavno,
-                end_wavno,
-                session,
-                out_decompressed_file,
-                verbose=verbose,  # Show internal download progress
+                print("Starting parallel downloads...\n")
+
+            max_threads = min(4, len(files_to_download))
```
minouHub: Is there a problem if someone has less than 4 processors?

merintheres: From my understanding, this is an I/O-bound operation, so using up to 4 threads should be fine even on machines with fewer cores, since threads are mostly waiting on network I/O and are scheduled by the OS.

merintheres: If you prefer, we can also switch this to cap the thread count using `os.cpu_count()`

merintheres:
@minouHub ,I’ve removed the try/except block around the chunk download as suggested.

merintheres:
I also benchmarked the performance with and without parallel downloads. For a small range (2300–2500 cm⁻¹, 2 chunks), the total time reduced from ~2.4 min to ~2.2 min (~8% improvement). For a wider range (2300–3300 cm⁻¹, 11 chunks), it reduced from ~14.8 min to ~12.4 min (~16% improvement).

radis/api/hitempapi.py
minouHub: Yes. Having at most `os.cpu_count()-1` sounds more reliable in my opinion. You don't want to overload the CPU

merintheres: I’ve updated the thread cap to use os.cpu_count() - 1

minouHub:
Hello,
Thank you for the benchmark. A speed increase of 16% is good! 
Another comment, could you rebase on `develop`? A fix in #896 should help having all the tests in green, especially the tests needing logging to HITRAN/HITEMP. This is necessary as you modify this part of the code.

Issue #840, opened 2025-06-25
Spectrum.recalc_gpu() doesn't work when neighbour_lines is not 0


cWetaski:
### 🐛 Describe the bug

I noticed that when I create a SpectrumFactory where I provide the argument "neighbour_lines" with a nonzero value, this leads to some issues with gpu spectrum calculations and specifically recalc_gpu(). 

For example, if I want to calculate the absorption spectra for a gas in the range [2300, 2310] cm-1, and I include the argument neighbour_lines=50, then the first time calculating the spectrum with eq_spectrum_gpu(), there is no issue, and it's clear that the neighbour lines have been accounted for. However, when I try to call recalc_gpu() on the resulting spectrum, something goes wrong. I believe that the absorption spectrum is subsequently calculated for the entire range including the range spanned by neighbour_lines (i.e., from 2250 to 2360 cm-1). However, the spectrum object is still associated with the specified wavenumber range of [2300, 2310]. This leads to errors being raised when trying to plot or analyze the resulting spectrum. See below a MWE:

```
from radis import SpectrumFactory, plot_diff
import matplotlib.pyplot as plt

wn1 = 2300
wn2 = 2310
T = 1500.0  # K
p = 1.0  # bar
x = 0.8
l = 0.2  # cm
w_slit = 1  # cm-1
sf1 = SpectrumFactory(
    wn1,
    wn2,  # cm-1
    molecule="CO2",
    isotope="1",
    wstep=0.002,
    neighbour_lines=0, # <---- No neighbour_lines
)
sf1.fetch_databank(
    source="hitran"
) 
sf2 = SpectrumFactory(
    wn1,
    wn2,  # cm-1
    molecule="CO2",
    isotope="1",
    wstep=0.002,
    neighbour_lines=50 # <----- Nonzero neighbour_lines
)
sf2.fetch_databank(
    source="hitran"
) 
s1 = sf1.eq_spectrum_gpu(
    name="without neighbour_lines",
    Tgas=T,
    pressure=p,
    mole_fraction=x,
    path_length=l,
)
s2 = sf2.eq_spectrum_gpu(
    name="with neighbour_lines",
    Tgas=T,
    pressure=p,
    mole_fraction=x,
    path_length=l,
)
plot_diff(s1, s2, "abscoeff", show=False)
s1.recalc_gpu(Tgas=T+800)
s1.plot("abscoeff", show=False, plotting_library="matplotlib") # No problem
s2.recalc_gpu(Tgas=T+800)
[...]

PR #891, opened 2025-12-13
fix: Resolve recalc_gpu plotting error caused by padded GPU output

PavanendraBaahubali:
### Description
This PR fixes a plotting error in GPU mode when neighbour_lines is enabled.
gpu_app.iterate() returns a padded absorption array larger than the user-defined wavespace.
Trimming the GPU output to the wavespace length restores consistent array sizes and fixes plotting.

This PR Fixes issue #840

### DIFF
--- a/radis/spectrum/spectrum.py
+++ b/radis/spectrum/spectrum.py
@@ -6504,6 +6504,11 @@ class Spectrum(object):
         self.conditions["NwL"] = iter_N_L
         self.conditions["NwG"] = iter_N_G
 
+        # Trim GPU output to match wavespace size
+        if "wavespace" in self._q:
+            n = len(self._q["wavespace"])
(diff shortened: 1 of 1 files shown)


minouHub:
Some tests will fail due to an update on the HITRAN side. Please rebase on develop

minouHub:
Hello,
Could you merge (or rebase on) develop? The tests should pass now that https://github.com/radis/radis/pull/910 is merged

minouHub:
Sounds weird to me than it has to be trimmed on the right and not on both sides. @dcmvdbekerom what do you think?

dcmvdbekerom:
I agree, I'd think it needs trimming from either side

minouHub:
Inactive. Re-open if you wish to finish the job :)

PR #910, opened 2026-01-29
Fix: Update non-regression test values for HITRAN2024

tanishadahale:
### Description
This PR addresses the failure in 
test_calc_spectrum_overpopulations
 caused by the recent release of HITRAN2024 (Jan 2026).

The HITRAN database has updated the line intensities and broadening parameters for CO2 in the 4.3 µm region (Row 2 of Table 2 in the HITRAN2024 paper). As a result, the calculated radiance values in RADIS have shifted slightly (~0.01% - 5%), exceeding the strict rtol=1e-3 tolerance of the non-regression test.

This PR updates the hardcoded reference values (I_ref) in 
radis/test/lbl/test_calc.py
 to match the new physically correct output from HITRAN2024.

Verified locally:

Running the test with the old values FAILED (confirming issue).
Running the test with the new values PASSED (confirming fix).


minouHub:
already pursued in #906. 
The new problem is the doc failing. If you can take a look to the doc problem, I'll merge your PR. I think it's a Python version problem but not sure

tanishadahale:
Oh okay; I'll close this PR then.

tanishadahale:
The ReadTheDocs tests were failing due to RADIS being updated to align with newer software and data standards. The issues and fixes are summarized below:

### 1. Numba Compatibility

**Issue:** Older code required *writable* arrays, but modern Pandas often provides *read-only* arrays.
**Fix (`radis/misc/arrays.py`):** Removed explicit type signatures from `@numba.njit` decorators so Numba can infer array types. This resolves errors with read-only arrays in newer Numba and Python 3.14, while preserving JIT performance.

### 2. NIST ASD Parsing

**Issue:** A logic bug failed to detect modern Arrow-backed string dtypes, causing required data cleanup to be skipped.
**Fix (`radis/api/nistapi.py`):** Replaced the broken dtype check with `pd.api.types.is_string_dtype` to reliably detect string columns and apply proper parsing.

### 3. Test Reporting

**Issue:** A programming error attempted to concatenate a dictionary with a string, causing the error reporter itself to crash.
**Fix (`radis/test/utils.py`):** Converted `entries` to a string when raising `ValueError`, ensuring clear and informative error messages instead of a crash.

minouHub:
Hello,
Nice! I had to remove some unused import to pass linting.

PR #915, opened 2026-02-13
Fix false HTML HEAD failures for HITEMP downloads

MJO7:
**Summary**
This fixes a false-positive download failure in DatabaseManager.download_and_parse for HITEMP/HITRAN URLs.

**Bug**
Some HITEMP endpoints can return text/html on HEAD (redirect/login-like response) even when the actual GET returns the correct binary [.par.bz2](https://file+.vscode-resource.vscode-cdn.net/Users/meghjoshi/.vscode/extensions/openai.chatgpt-0.4.74-darwin-arm64/webview/#) file.
Previous logic raised immediately on HEAD content-type: text/html, which blocked valid downloads.

**Changes**

Updated [dbmanager.py](https://file+.vscode-resource.vscode-cdn.net/Users/meghjoshi/.vscode/extensions/openai.chatgpt-0.4.74-darwin-arm64/webview/#):
Keep HEAD as an availability check only.
Raise on HEAD only for HTTP error statuses (>= 400).
Move HTML/login-page detection to the actual GET response.
Only raise if GET indicates HTML in both header and payload sniffing (prevents false positives).
Added regression tests in [test_dbmanager.py](https://file+.vscode-resource.vscode-cdn.net/Users/meghjoshi/.vscode/extensions/openai.chatgpt-0.4.74-darwin-arm64/webview/#):
Allows HEAD: text/html when GET is valid binary.
Raises when GET is actually HTML.
Raises on HEAD error status.
Allows redirect-like HEAD status when GET succeeds.
Verifies HITEMP path uses login_to_hitran session and still succeeds.

**Validation**

[test_dbmanager.py -vv](https://file+.vscode-resource.vscode-cdn.net/Users/meghjoshi/.vscode/extensions/openai.chatgpt-0.4.74-darwin-arm64/webview/#)
Result: 5 passed
[test_dbmanager.py](https://file+.vscode-resource.vscode-cdn.net/Users/meghjoshi/.vscode/extensions/openai.chatgpt-0.4.74-darwin-arm64/webview/#)
Result: 12 passed

**Impact**
Fixes intermittent/first-run-like HITEMP download failures without weakening protection against actual HTML/login-page responses.


minouHub:
This PR is indeed much needed! 
As I understand, you want to have a clearer error message when HITRAN or HITEMP download provides an empty page. This will not change the problem but at least the error message will be more explicit.
Proposition: could you re-trigger the download attempt a second time if this kind of issue occurs?

MJO7:
Thank you for you response! I am working on the proposition as suggested and looking for pathways that re-trigger the attempt. Let me get back to you in some time. Thank you for the guidance.

minouHub:
@MJO7 are you related to the #940 ?

MJO7:
@minouHub Thanks for pointing this out. Yes, this PR relates to the same issue. My implementation focuses on moving the HTML detection to the GET response and only raising on actual HTTP error status codes from HEAD. Working forward, I'd be happy to coordinate and consolidate approaches.

minouHub:
Alright, I'll go for #940 but you will be marked as contributor as well

PR #918, opened 2026-02-17
Fix inconsistent spectral_points reporting

### LINKED ISSUE #613: Inconsistent reporting of number of spectral points
### 🐛 Describe the bug

When the spectral axis is generated it is done so with `wmax` inclusive; this leads to numbers of spectral points such as `150001`. This is also the number you get when printing `len(sf.wavenumber)`. However, `s.conditions['spectral_points']` yields the exclusive answer, i.e. `150000`. It would be nice if these values were consistent.

I used the following script:

```

sf = SpectrumFactory(
    2150,
    2450,  # cm-1
    molecule="CO2",
    isotope="1,2,3",
    wstep=0.002,
)

sf.fetch_databank("hitemp")

s = sf.eq_spectrum(
    Tgas=1100.0,  # K
    pressure=1,  # bar
    mole_fraction=0.8,
    path_length=0.2,  # cm
)
```

### 💡 Possible solutions

Wouldn't it be better to use spectral axes with the `wmax` exclusive?

Alternatively, when updating the `sf.conditions['spectral_points']`, rather than calculating the number of points from input parameters the integer value should be read from elsewhere.

### 🎲 Radis version

0.14

### 💻 Operating system

Windows

apoorvdarshan:
## Summary
- Fixes `s.conditions['spectral_points']` reporting incorrect value (e.g. `150000.0` instead of `150001`)
- The old formula `int(wavenum_max_calc - wavenum_min_calc) / wstep` used the calc range (with neighbour_lines buffer) and computed intervals instead of points (off-by-one)
- Replaced with `len(self.wavenumber)` at all 4 locations in `radis/lbl/factory.py`

Fixes #613

## Test plan
- [x] Reproduced the issue from #613 with `SpectrumFactory(2150, 2450, wstep=0.002)` — confirmed `spectral_points` now matches `len(s.get_wavenumber())`
- [x] All 18 factory tests pass

### DIFF
--- a/radis/lbl/factory.py
+++ b/radis/lbl/factory.py
@@ -949,10 +949,7 @@ class SpectrumFactory(BandFactory):
                 "thermal_equilibrium": True,
                 "diluents": self._diluent,
                 "radis_version": version,
-                "spectral_points": (
-                    int(self.params.wavenum_max_calc - self.params.wavenum_min_calc)
-                    / self.params.wstep
-                ),
+                "spectral_points": len(self.wavenumber),
                 "profiler": dict(self.profiler.final),
             }
         )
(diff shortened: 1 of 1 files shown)


minouHub (approved):
Nice!

minouHub:
HITRAN is under maintenance, which explains why the tests might be failing.

PR #926, opened 2026-02-20
Refactor: Remove mutable default arguments in fitting and factory modules

vishwas-droid:
- Replaced mutable default arguments (e.g., `{}`) with `None`.
- Added explicit initialization inside function bodies.
- Preserved existing default values and behavior.
- Applied changes in fitting and factory-related modules only.
- No functional changes introduced.

### DIFF
--- a/radis/lbl/factory.py
+++ b/radis/lbl/factory.py
@@ -1392,7 +1392,7 @@ class SpectrumFactory(BandFactory):
         var="transmittance",
         slit_function=0.0,
         mpl_backend="",
-        plotkwargs={},
+        plotkwargs=None,
         *vargs,
         **kwargs,
     ) -> Spectrum:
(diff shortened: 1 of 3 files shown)


minouHub:
Not sure I understood why these minor changes are needed. Re-opn if you can make a case about this PR

vishwas-droid:
Thanks for taking another look.

The main reason for this change is to eliminate mutable default arguments such as {} and [], which in Python are instantiated once at function definition time and then shared across calls.

Even if the current implementation does not mutate these objects, the pattern is considered unsafe because:

Future modifications may introduce mutation without realizing the shared-state behavior.

Bugs caused by shared defaults are subtle and difficult to trace.

The fix is low-cost and makes the function behavior strictly safer without altering logic.

Since these functions are part of fitting and factory-related paths, which may evolve over time, I felt it was better to remove this potential footgun proactively rather than wait for a hard-to-diagnose issue later.

PR #930, opened 2026-02-21
Add file-like object support to spectrum.store() (fixes #506)

### LINKED ISSUE #506: 🚀 Feature: Add ability to write spectrum file to a buffer instead of to disk
### 🔖 Feature description

We're working on a feature for RADIS App to allow the user to download the spectrum file (https://github.com/suzil/radis-app/pull/619). To do this, we do `spectrum.store(file_path)`, load the file from disk to serve to the user, and then delete the file. If we could instead pass a [buffer object](https://docs.python.org/3.8/library/io.html#buffered-streams) to write the file to memory, we could simplify our implementation of this feature and improve performance by skipping the write to disk. The buffer would have the [same interface](https://stackoverflow.com/a/18550157) as someone doing `fp = open(file_path)` and passing `fp` to `spectrum.store(fp)`, so RADIS would just write the data there and the client can take care of whether it goes to disk or to an in-memory buffer.

I propose adjusting `spectrum.store(file_path)` to accept either a string file path (as it is now) or a file handle-like object. If passed a file handle-like object, we would skip the `open(...)` lines like here: https://github.com/radis/radis/blob/91bfcd0bd2c317a17fc80e8150322370ccba321d/radis/tools/database.py#L211

Simplify RADIS App implementation of serving spectrum download files to users and improve performance.

tanishadahale:
## Description

Allow spectrum.store() to accept file-like objects (BytesIO, StringIO) instead of only file paths.

```python
from io import BytesIO
buffer = BytesIO()
spectrum.store(buffer)
```
Fixes #506

### DIFF
--- a/radis/spectrum/spectrum.py
+++ b/radis/spectrum/spectrum.py
@@ -4104,12 +4104,18 @@ class Spectrum(object):
 
         Parameters
         ----------
-        path: path to folder (database) or file
-            if a folder, file is saved to database and name is generated automatically.
-            if a file name, then Spectrum is saved to this file and the later
-            formatting options dont apply
-        file: str
-            explicitly give a filename to save
+        path: str or file-like object
+            If a string: path to folder (database) or file. If a folder, file
+            is saved to database and name is generated automatically. If a file
+            name, then Spectrum is saved to this file and the later formatting
+            options dont apply.
+
+            If a file-like object (e.g., ``io.BytesIO``, ``io.StringIO``):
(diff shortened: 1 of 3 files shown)


Prtm2110 (approved):
Looks good, thanks @tanishadahale

Prtm2110:
@minouHub I am merging it now. If you have any follow-ups on this, let @tanishadahale know; she can create another PR.

Prtm2110:
Thanks again! @tanishadahale

PR #936, opened 2026-02-23
Refactor: switch from os.path to pathlib (partial)

tanishadahale:
### Description

 Fix for #53 - switching os.path to pathlib in a few files.

Changed:
- radis/db/utils.py
- radis/db/hitemp_co2.py
- radis/misc/utils.py

### DIFF
--- a/radis/db/hitemp_co2.py
+++ b/radis/db/hitemp_co2.py
@@ -9,7 +9,7 @@ database using block-aligned bzip2 decompression for specific wavenumber ranges.
 
 import bz2
 import io
-import os
+from pathlib import Path
 
 import numpy as np
 from tqdm import tqdm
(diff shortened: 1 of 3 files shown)


Prtm2110 (commented):
Thanks @tanishadahale, did you add commits of Add file-like object support to spectrum.store() by mistake? 
It seems that these are unrelated issues and have separate PR (or else you can mention it in the PR description and title).

PS: No need to add test for `os.path to pathlib` change.

tanishadahale:
Done! I rebased and removed the extra commits. Now the PR only contains the pathlib changes for #53.

Prtm2110:
Thanks! Did some minor change - removed `if path.parent != Path() else Path(".")` as the if condition is  always true. 
Since these are trivial changes I will go ahead and merge this myself.
Also hoping the doc build fail is unrelated!

PR #940, opened 2026-02-24
Fix spurious HITEMP HTTPError when HEAD returns HTML

### LINKED ISSUE #893: `download_and_parse` raises spurious HTTPError when HEAD request returns HTML for HITEMP URLs
### 🐛 Describe the bug

Hello @minouHub, thank you for your efforts in maintaining `radis`!
I encountered an issue when using `HITEMPDatabaseManager` through `ExoJAX`, and I think the current error logic in download_and_parse may be too strict.

🔍 Problem Summary

In `DatabaseManager.download_and_parse()`, the following check triggers an HTTPError:

if "text/html" in head_response.headers.get("content-type", "").lower():
    raise requests.HTTPError(
        "Received an HTML response instead of the expected file content. ..."
    )

However, for HITEMP URLs, the HEAD request often returns HTML (e.g., a generic redirect page), while the actual GET request succeeds and correctly returns the .par.bz2 file.

As a result, RADIS refuses to download files that are actually accessible.

📌 Observed Behavior

`head_response.status_code == 200`

But `head_response.headers["content-type"] == "text/html"`

`session.get(url)` afterwards does download the real .bz2 file correctly

The HTTPError prevents using the file—even though the download would succeed.

This makes the current HEAD-based detection too strict.

After comment out [L429-439](https://github.com/radis/radis/blob/d8c81956841ac1c47dabc6f5bf55f84fecc86560/radis/api/dbmanager.py#L429) in `radis.api.dbmanager`, it completely worked.

- ExoJAX side code

```python
from exojax.database.hitemp.api import MdbHitemp

nus, wav, res = wavenumber_grid(22920.0, 23100.0, 10000, unit="AA", xsmode="modit")
mdb = MdbHitemp(".database/CH4", nurange=nus, gpu_transfer=False, isotope=1)

```
### Update (Dec 17, 2025)

In some environments, the current version may still work correctly. The conditions under which this error occurs have not yet been identified.

### 💡 Possible solutions

removes L429-439 in `radis.api.dbmanager`?

### 🎲 Radis version

0.16.4

### 💻 Operating system

Linux

Flamki:
### Description
This pull request is to address false-positive download failures in `DatabaseManager.download_and_parse` (issue #893).

Problem:
- Some HITEMP URLs return `text/html` on `HEAD` even when `GET` returns the correct binary payload.
- Current code treats `HEAD content-type == text/html` as a hard error, causing spurious `HTTPError`.

What changed:
- Keep `HEAD` as a lightweight availability check only:
  - raise only when `HEAD` status is `>= 400`.
- Move HTML/login-page detection to the actual `GET` response:
  - inspect `GET` `content-type` and first payload chunk,
  - raise only when both indicate HTML (e.g., login page),
  - otherwise continue download/parse.
- Preserve streamed downloading behavior by peeking first non-empty chunk and then continuing over the same iterator.

Tests added:
- `test_hitemp_head_html_get_binary_does_not_fail`
- `test_get_html_payload_raises_httperror`
- `test_head_http_error_raises_before_get`
- `test_head_redirect_status_is_accepted`

Validation run:
- `python -m pytest radis/test/api/test_dbmanager_download.py -q`

Fixes #893

### DIFF
--- a/radis/api/dbmanager.py
+++ b/radis/api/dbmanager.py
@@ -441,10 +441,9 @@ class DatabaseManager(object):
                 )
 
             def _looks_like_html_payload(chunk):
-                snippet = chunk[:1024].lstrip().lower()
-                return snippet.startswith(b"<!doctype html") or snippet.startswith(
-                    b"<html"
-                )
+                snippet = chunk[:2048].lstrip().lower()
+                html_markers = (b"<!doctype html", b"<html", b"<head", b"<body")
+                return any(marker in snippet for marker in html_markers)
 
             try:
                 # Now download the file
@@ -462,9 +461,9 @@ class DatabaseManager(object):
                         first_chunk = chunk
                         break
 
-                if "text/html" in content_type and _looks_like_html_payload(
-                    first_chunk
-                ):
+                # Reject HTML responses from GET directly. Keep payload sniffing as
+                # fallback when servers mislabel an HTML page as octet-stream.
+                if "text/html" in content_type or _looks_like_html_payload(first_chunk):
                     raise requests.HTTPError(
                         "Received HTML content from GET request instead of the expected file payload. "
                         f"This may indicate authentication is required or access is restricted for URL: {urlname}."
--- a/radis/api/dbmanager.py
+++ b/radis/api/dbmanager.py
@@ -434,9 +434,9 @@ class DatabaseManager(object):
 
             # First check if we can access the file
             head_response = session.head(urlname, headers=headers, allow_redirects=True)
-            if head_response.status_code >= 400:
+            if head_response.status_code != 200:
                 raise requests.HTTPError(
-                    f"Unable to access the resource (HEAD request). Received HTTP status code {head_response.status_code} for URL: {urlname}. "
+                    f"Unable to access the resource (HEAD request). Expected HTTP status code 200, got {head_response.status_code} for URL: {urlname}. "
                     "Please verify the URL and your network access permissions."
                 )
 
--- a/radis/api/dbmanager.py
+++ b/radis/api/dbmanager.py
@@ -434,16 +434,16 @@ class DatabaseManager(object):
 
             # First check if we can access the file
             head_response = session.head(urlname, headers=headers, allow_redirects=True)
-            if head_response.status_code != 200:
+            if head_response.status_code >= 400:
                 raise requests.HTTPError(
-                    f"Unable to access the resource. Received HTTP status code {head_response.status_code} for URL: {urlname}. "
+                    f"Unable to access the resource (HEAD request). Received HTTP status code {head_response.status_code} for URL: {urlname}. "
                     "Please verify the URL and your network access permissions."
                 )
 
-            # Check if we got redirected to login page
-            if "text/html" in head_response.headers.get("content-type", "").lower():
-                raise requests.HTTPError(
-                    "Received an HTML response instead of the expected file content. This may indicate authentication is required or access to the resource is restricted. Please verify your credentials and permissions for the requested URL: ({urlname})."
+            def _looks_like_html_payload(chunk):
+                snippet = chunk[:1024].lstrip().lower()
+                return snippet.startswith(b"<!doctype html") or snippet.startswith(
+                    b"<html"
                 )
 
             try:
@@ -452,6 +452,23 @@ class DatabaseManager(object):
                     urlname, headers=headers, stream=True, allow_redirects=True
                 )
                 response.raise_for_status()  # Raise an error if request fails
+                content_type = response.headers.get("content-type", "").lower()
+                chunks = response.iter_content(chunk_size=8192)
+
+                # Peek the first non-empty chunk to detect HTML/login pages.
+                first_chunk = b""
+                for chunk in chunks:
+                    if chunk:
+                        first_chunk = chunk
+                        break
+
+                if "text/html" in content_type and _looks_like_html_payload(
+                    first_chunk
+                ):
+                    raise requests.HTTPError(
+                        "Received HTML content from GET request instead of the expected file payload. "
+                        f"This may indicate authentication is required or access is restricted for URL: {urlname}."
+                    )
 
                 # Create a temporary file to store the downloaded content
                 temp_file_path = urlname.split("/")[-1]
@@ -474,7 +491,11 @@ class DatabaseManager(object):
                         desc="Downloading",
                         disable=not verbose,
                     ) as pbar:
-                        for chunk in response.iter_content(chunk_size=8192):
+                        if first_chunk:
+                            f.write(first_chunk)
+                            pbar.update(len(first_chunk))
+
+                        for chunk in chunks:
                             if chunk:  # filter out keep-alive new chunks
                                 f.write(chunk)
                                 pbar.update(len(chunk))
--- a/.github/workflows/tests-develop.yml
+++ b/.github/workflows/tests-develop.yml
@@ -13,7 +13,7 @@ jobs:
   build:
     strategy:
       fail-fast: false
-      max-parallel: 3
+      # max-parallel: 3
       matrix:
         include:
           # - os: ubuntu-latest
--- a/radis/test/api/test_dbmanager_download.py
+++ b/radis/test/api/test_dbmanager_download.py
@@ -63,9 +63,37 @@ class _DummyDownloadManager(DatabaseManager):
         return 1
 
 
+@pytest.fixture
+def manager(tmp_path):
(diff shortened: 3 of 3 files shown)


Flamki:
Update on #940

I pushed a follow-up commit to fix the CI lint failure:

- Commit: `864592a5`
- Change: applied `isort` import ordering in `radis/test/api/test_dbmanager_download.py`
- No functional logic changes to the HEAD/GET fix itself.

Local validation run:
- `python -m isort radis/test/api/test_dbmanager_download.py radis/api/dbmanager.py`
- `python -m pytest radis/test/api/test_dbmanager_download.py radis/test/spectrum/test_spectrum_io.py::test_specutils_io -q`
- Result: `5 passed`

Note on the previous Ubuntu failure (`test_specutils_io` with `ValueError: invalid literal for int() with base 10: '  '`):
- I could not reproduce it locally after installing `specutils`.
- On this branch, the target test passes consistently in my environment.
- Since #940 only touches `radis/api/dbmanager.py` and new tests under `radis/test/api/`, that spectrum I/O failure appears unrelated to this PR’s code path.

minouHub:
To maximize the probability to trigger a false positive, I recommend you to **comment** this line in `tests-develop.yml`
```
max-parallel: 3
```

Flamki:
Addressed the Ubuntu 3.10 flaky failure with a follow-up hardening commit: `56cdcdc`.

What changed
- Tightened GET-side HTML rejection in `download_and_parse`:
  - now fails if GET `content-type` is `text/html` (even when payload sniff is inconclusive), and
  - still fails when payload sniff looks like HTML even if `content-type` is misleading.
- Expanded HTML sniff markers (`<!doctype html`, `<html`, `<head`, `<body`) and increased snippet size.
- Added regression test for the edge case where GET reports HTML content-type but the first chunk is non-HTML bytes (`\x00\x00`), to avoid passing bad payloads to parsers.

Why
- This prevents malformed/non-file GET responses from slipping through and failing later with parser errors like `invalid literal for int() ...`.
- Keeps the original #893 fix intent (HEAD HTML is tolerated; GET must be validated by real payload response).

Validation run
- `python -m pre_commit run --files radis/api/dbmanager.py radis/test/api/test_dbmanager_download.py`
- `python -m pytest radis/test/api/test_dbmanager_download.py -q`
- `python -m pytest radis/test/io/test_hdf5.py::test_local_hdf5_lines_loading -q -s`

Flamki:
Applied your suggestion exactly in this PR branch:
- commented max-parallel: 3 in .github/workflows/tests-develop.yml (commit 7acff2f).

This is CI-only and separate from the download/parsing logic fix, to maximize reproducibility of flaky failures on Ubuntu 3.10.

radis/api/dbmanager.py
```diff
@@ -434,24 +434,40 @@ def download_and_parse_one_file(urlname, local_file, Ndownload, printer):
 
             # First check if we can access the file
             head_response = session.head(urlname, headers=headers, allow_redirects=True)
-            if head_response.status_code != 200:
+            if head_response.status_code >= 400:
```
minouHub: I do not understand why you changed this condition. If `head_response.status_code != 200` then it is not what we expect from HITRAN and an error should be raised. If you are convinced this should be changed, explain me why

minouHub (commented):
This looks like an enhanced duplicate of #915. Are you guys related?

minouHub:
@Prtm2110 can I have your opinion on this one? I'm hoping it will solve or at least make the errors related to the downloading of database more explicit.

Examples:

- in https://github.com/radis/radis/actions/runs/22457766547/job/65043476124?pr=949#step:6:1224

`FAILED radis/test/lbl/test_calc.py::test_calc_spectrum_multiple_molecules_wstep_auto - radis.misc.warning.EmptyDatabaseError: CO has no lines on range 1999.45-2400.31 cm-1`

- in https://github.com/radis/radis/actions/runs/22441630358/job/64985252977?pr=948#step:6:425

`FAILED radis/test/io/test_query.py::test_fetch_hitran_CO_vaex - ValueError: invalid literal for int() with base 10: '\x00\x00'`

Flamki:
@minouHub Thanks for the review, good point.

On `HEAD` status handling (`!=200` vs `>=400`):
the intent of changing to `>=400` was to treat `HEAD` as a lightweight pre-check only, because `HEAD` can be inconsistent on some endpoints (redirect/CDN/auth flows) while `GET` still succeeds with the actual file. In this PR, `GET` remains the strict source of truth (`raise_for_status()` + HTML/login payload detection), so we still fail hard on real download errors.

So the goal was to avoid false negatives from `HEAD` while keeping strict validation on `GET`.

About #915:
I’m not related to that PR; this one was opened from #893 independently. If you prefer to keep only one PR, I’m happy to close this and move any useful delta to #915 (or vice versa), whichever is easier for maintainers.

Flamki:
Applied your requested change in this PR:
- restored strict HEAD preflight check to require status 200 (commit e8140bf).
- updated regression test accordingly: redirect HEAD (302) now raises before GET.

Validation run:
- python -m pre_commit run --files radis/api/dbmanager.py radis/test/api/test_dbmanager_download.py
- python -m pytest radis/test/api/test_dbmanager_download.py -q (5 passed)
- python -m pytest radis/test/io/test_hdf5.py::test_local_hdf5_lines_loading -q -s (1 passed)

minouHub:
@MJO7 will be marked as contributor as well, because his PR was extremely similar

minouHub:
Clearly does not solve the problem of database fetching, see [this line](https://github.com/radis/radis/actions/runs/22524823620/job/65366184477?pr=940#step:6:1258)
```
FAILED radis/test/lbl/test_calc.py::test_calc_spectrum - ValueError: invalid literal for int() with base 10: '1,2'
FAILED radis/test/lbl/test_calc.py::test_calc_spectrum_overpopulations - ValueError: invalid literal for int() with base 10: '1,2'
FAILED radis/test/lbl/test_calc.py::test_eq_vs_noneq_isotope - assert 0
FAILED radis/test/validation/test_RADIS_vs_HAPI_line_broadening.py::test_line_broadening - assert np.False_
```

Flamki:
Thanks for the clarification 👍

No problem at all regarding marking @MJO7 as contributor — the approaches are indeed very similar and the objective is aligned: making the download handling more explicit and robust.

Regarding the failing Ubuntu 3.10 / 3.11 jobs, I’m currently reviewing the logs to determine whether the failures are:

- directly related to the HEAD/GET preflight logic, or  
- side effects of database state / CI download flakiness.

Since macOS, Windows and 3.12 are passing, this may point to a platform-specific or timing-related issue, but I’ll confirm after isolating the failing tests locally in a 3.10/3.11 environment.

On the Codecov project drop:
The patch coverage is ~99%, so the new logic is well covered. I’ll check which existing lines are now reported as uncovered and see whether a small additional test or minor refactor is needed to avoid affecting global coverage.

I’ll report back once I’ve narrowed down the Ubuntu failure cause.

minouHub:
> Regarding the failing Ubuntu 3.10 / 3.11 jobs, I’m currently reviewing the logs to determine whether the failures are: directly related to the HEAD/GET preflight logic, or side effects of database state / CI download flakiness.
Since macOS, Windows and 3.12 are passing, this may point to a platform-specific or timing-related issue, but I’ll confirm after isolating the failing tests locally in a 3.10/3.11 environment.

No it's just a problem with the download from HITRAN.

> On the Codecov project drop:
The patch coverage is ~99%, so the new logic is well covered. I’ll check which existing lines are now reported as uncovered and see whether a small additional test or minor refactor is needed to avoid affecting global coverage.

Not a problem. 

I accept the PR but I'm waiting for @Prtm2110 review to merge it.

minouHub (approved):

Prtm2110 (commented):
Hi the test 2 and 3 are redundant - you can  merge them into one. Same goes for test 4 and 5. Basically can you make the tests concise and easier to maintain.

Prtm2110:
Use can use `@pytest.fixture` to combine the common part of tests. This will surely make the test suit smaller.

Flamki:
@minouHub @Prtm2110

Implemented the test refactor requested in review.

- Consolidated redundant cases using fixtures + parametrization in `radis/test/api/test_dbmanager_download.py`.
- Merged former test pairs (2+3 and 4+5) into concise parameterized tests.
- Kept behavior coverage the same while reducing duplication and maintenance overhead.
- No functional change in `dbmanager` logic in this commit.

Local validation:
- `python -m pytest radis/test/api/test_dbmanager_download.py -q` -> 6 passed

Prtm2110 (approved):
Looks better, merging since @minouHub has already approved!

Issue #629, opened 2023-11-01
Monitor Radis Import time


erwanp:
### 💭 Description

Reminder that we can monitor radis imports with 

```
python -X importtime -c 'import radis' 
```

Results can also be analysed with `tuna`   (```pip install tuna```)

```
python -X importtime -c 'import radis' 2> radis-imports
tuna radis-imports
```

![image](https://github.com/radis/radis/assets/16088743/0c696b19-1585-41af-bf39-64f7d0e5334e)

Zeyadzahran:
This also must be closed @minouHub 
It's done via #941

PR #941, opened 2026-02-24
perf: optimize startup time by lazy loading heavy dependencies

### LINKED ISSUE #937: Profiling import radis startup time: Current status vs #629
### Description 

I was looking into the `import radis` startup time mentioned in #629. I ran some profiling on the current version, and it seems the import time hasn't improved. It currently averages around 2.0s, which is slightly slower than the 1.9s reported back in 2023.

### Profiling the bottlenecks

To see what is taking so much time, I ran the import profiler and extracted a sorted list of the slowest modules:

```bash
conda run -n radis-env python -X importtime -c 'import radis' 2>&1 | grep "radis\." | sort -t'|' -k2 -rn | head -20
```

Image

I used `tuna` to visualize the full import tree

```Bash
conda run -n radis-env python -X importtime -c 'import radis' 2> radis-imports
conda run -n radis-env tuna radis-imports
```

Image

### Breakdown of the bottlenecks

Based on the profiling, here are the main issues:

1. **radis.misc.config (~45% of total time)**: This loads right at the beginning, even if the user never uses any config options.
    
2. **scipy.signal via radis.tools.slit (~14%)**: This is a large library that loads on startup, but we probably only need it to load when someone actually calls `apply_slit()`.
    
3. **pandas (~18%) and numba (~15%)**: Both of these also load immediately at startup, even if the user's script doesn't need them right away.
    

### Exact time verification

To make sure these numbers are accurate and avoid caching issues, I ran a script across 5 independent Python processes (RADIS v0.16.4, Python 3.x, .pyc cache warm):
```python
import subprocess, sys
times = []
for i in range(5):
    r = subprocess.run(
        [sys.executable, '-c', '''
import time
t0 = time.perf_counter()
import radis
print(f"{time.perf_counter()-t0:.4f}")
'''],
        capture_output=True, text=True
    )
    t = float(r.stdout.strip())
    times.append(t)
    print(f'Run {i+1}: {t:.3f}s')

print(f'---')
print(f'Min:  {min(times):.3f}s')
print(f'Max:  {max(times):.3f}s')
print(f'Mean: {sum(times)/len(times):.3f}s')
```

This is the results 

Image

### Benchmark between the results of #629 and the current results

Image

### Radis version

0.16.4

### Operating system

Linux

cc @Prtm2110 @minouHub @erwanp 

I'd like to share these findings with you since this was tracked before. I would love to hear your thoughts on these bottlenecks. What are your opinions on how we should approach a solution for these eager imports to bring the startup time down?

I was thinking about converting these heavy top-level imports to lazy imports as a potential fix. For example, we could defer `scipy.signal` to only load inside `apply_slit()`.

Zeyadzahran:
### Description

This PR addresses the slow startup time reported in #937 (and previously #629).

Profiling showed that `import radis` was taking about 2.2s because it was eagerly loading heavy dependencies like `scipy`, `astropy`, `plotly`, and the `radis.io` module right at the start.

I have updated these to use lazy loading so they are only imported when actually used.

### Changes

* **Function-Level Deferral:** Moved top-level imports (`scipy`, `astropy`, `plotly`, `bs4`, `requests`) inside the specific functions that need them (mostly in `radis/tools`, `radis/lbl`, and `radis/misc`).
* **Module-Level Lazy Loading:** Updated `radis/io/__init__.py` to use `__getattr__` so fetch functions are only loaded when accessed.

### Benchmarks

I measured the startup time using `python -X importtime` (averaging 5 runs on Linux, Python 3.14).
The mean startup time dropped from **~2.17s to ~1.02s** (~53% faster).
image

**Analysis:**
The chart above breaks down the time spent per module. You can see the heavy bottlenecks in the current version (Red bars):
* `radis.io` (~721ms) and `scipy.signal` (~376ms) were the biggest costs.
* In this PR (Green bars), `scipy.signal`, `astropy`, and `bs4` are completely deferred (0ms).
* `radis.io` is reduced by 98% (down to ~13ms).

**Tuna Profile**
The profile below confirms a clean import tree. The heavy modules are no longer blocking the main thread.

image

### How to Verify

```bash
# Quick benchmark 
 for i in 1 2 3 4 5; do python -c "import time; s=time.perf_counter(); import radis; print(f'{time.perf_counter()-s:.3f}s')"; done

# Per-module profiling
python -X importtime -c "import radis" 2>&1 | sort -t'|' -k2 -rn | head -15
```

**Note:** The numbers reported here are the **mean** (average) of my test runs, not the absolute best or worst case. While exact timings may vary slightly on different machines, you should get results very near to these figures.

Fixes #937

### DIFF
--- a/radis/lbl/base.py
+++ b/radis/lbl/base.py
@@ -63,7 +63,6 @@ Most methods are written in inherited class with the following inheritance schem
 
 
 """
-
 # TODO: move all CDSD dependant functions _add_Evib123Erot to a specific file for CO2.
 
 import numpy as np
--- a/radis/lbl/base.py
+++ b/radis/lbl/base.py
@@ -63,11 +63,11 @@ Most methods are written in inherited class with the following inheritance schem
 
 
 """
+
 # TODO: move all CDSD dependant functions _add_Evib123Erot to a specific file for CO2.
 
 import numpy as np
 import pandas as pd
-from astropy import units as u
 from numpy import exp, pi
 from psutil import virtual_memory
 
@@ -3979,6 +3979,8 @@ def get_wavenumber_range(
     """
 
     # Checking consistency of all input variables
+    from astropy import units as u
+
     assert medium in ["air", "vacuum"]
 
     w_present = wmin is not None and wmax is not None
--- a/radis/lbl/factory.py
+++ b/radis/lbl/factory.py
@@ -77,16 +77,12 @@ for Developers:
 ----------
 """
 
-from __future__ import annotations
-
-from typing import TYPE_CHECKING, Union
+from typing import Union
 from warnings import warn
 
 import numpy as np
 from numpy import arange, exp, expm1
-
-if TYPE_CHECKING:
-    from scipy.optimize import OptimizeResult
+from scipy.optimize import OptimizeResult
 
 from radis import version
 from radis.db import MOLECULES_LIST_EQUILIBRIUM, MOLECULES_LIST_NONEQUILIBRIUM
--- a/radis/lbl/factory.py
+++ b/radis/lbl/factory.py
@@ -76,13 +76,17 @@ for Developers:
 
 ----------
 """
-from typing import Union
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Union
 from warnings import warn
 
-import astropy.units as u
 import numpy as np
 from numpy import arange, exp, expm1
-from scipy.optimize import OptimizeResult
+
+if TYPE_CHECKING:
+    from scipy.optimize import OptimizeResult
 
 from radis import version
 from radis.db import MOLECULES_LIST_EQUILIBRIUM, MOLECULES_LIST_NONEQUILIBRIUM
@@ -593,6 +597,8 @@ class SpectrumFactory(BandFactory):
             )
 
         # Initialize input conditions
+        import astropy.units as u
+
         self.input.wavenum_min = wavenum_min
         self.input.wavenum_max = wavenum_max
         self.input.Tref = convert_and_strip_units(Tref, u.K)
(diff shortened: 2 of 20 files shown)


Zeyadzahran:
I just finished the PR  I loved working on it, it was very interesting. Waiting for your review, @Prtm2110, @minouHub

radis/lbl/base.py
```diff
@@ -63,11 +63,11 @@
 
 
 """
+
```
minouHub:
```suggestion
```

radis/lbl/factory.py
```diff
@@ -76,13 +76,17 @@
...
+from typing import TYPE_CHECKING, Union
 from warnings import warn
 
-import astropy.units as u
 import numpy as np
 from numpy import arange, exp, expm1
-from scipy.optimize import OptimizeResult
+
+if TYPE_CHECKING:
+    from scipy.optimize import OptimizeResult
```
minouHub, line 86: Seems a bit convoluted. Do we need the "annotations" import?
Also, do we gain significant time with lazy import of scipy.optimize?

minouHub (approved):

radis/lbl/factory.py
Zeyadzahran: hi @minouHub 

Yes, `from __future__ import annotations` is required without it, the `OptimizeResult` type hint in `fit_spectrum()` would crash at runtime because it's only imported under `TYPE_CHECKING`. The `annotations` import makes Python treat all type hints as strings so they're never actually evaluated.

For `scipy.optimize` honestly  the time savings are tiny (~5ms) since scipy usually gets loaded anyway. I mostly did it for consistency. Happy to revert this one and keep it as a normal import if you think it's overwhelming. The big wins are from deferring `astropy` and `radis.io`.

minouHub: Alright yes then let's revert it to keep it simple if you don't mind

Zeyadzahran:
Is this failing Tests related @minouHub

minouHub (approved):

minouHub:
RDT is failing and it is likely related to your PR (other PR pass). Check the logs of RDT to find the problem

Zeyadzahran:
RDT passed, but other tests failed. What should I do now @minouHub

minouHub:
The tests are ok, they failed because of bad download of database. For travis, something messed up with the upload to codecov but you're working on this in #948

PR #949, opened 2026-02-26
fix: replace all bare except: with specific exception types

### LINKED ISSUE #945: Bare `except`: blocks swallow KeyboardInterrupt and mask errors
### 🐛 Describe the bug

I was working through the codebase and noticed a recurring pattern of bare `except:` blocks (just `except:` without specifying an exception type).

This is dangerous because in Python, a bare `except:` catches **everything**, including `SystemExit` and `KeyboardInterrupt`.

1.  **Hanging processes:** If the code hangs, a user cannot stop it using **Ctrl+C** because the signal is swallowed.
2.  **Hidden bugs:** It masks legitimate errors (like `NameError`) without printing a traceback.

I noticed this pattern in several modules, for example:
* `radis/misc/utils.py`
* `radis/tools/database.py`
* `radis/misc/arrays.py` 
and more but I still try to address them
### Minimal code to reproduce

```python
from radis import load_spec

# Load a test spectrum
s = load_spec(getTestFile("CO_Tgas1500K_mole_fraction0.01.spec"), binary=True)

# radis/misc/utils.py uses a bare 'except: pass' inside this function.
# If an error happens here (or if the user tries to Ctrl+C), 
# it is silently ignored.
s.print_conditions() 
```

### 💡 Possible solutions

We should do a sweep of the codebase to replace these:

1.  **General Fix:** Replace `except:` with `except Exception:`. This is safer as it catches standard program errors but lets `KeyboardInterrupt` and `SystemExit` propagate.
2.  **Specific Fix:** In places like `arrays.py` (where it checks for scalars), we should replace the `try/except` block with a proper check (e.g., `if np.ndim(searched) == 0`) for better performance.

### 🎲 Radis version

0.16.4

### 💻 Operating system

Linux

Zeyadzahran:
### Description
I removed all bare `except:` blocks across the codebase. 

Bare `except:` blocks are dangerous as they catch `KeyboardInterrupt` and `SystemExit`, preventing users from safely stopping hanging processes and masking legitimate underlying bugs by swallowing all tracebacks.

**Key Changes:**
* Carefully analyzed each `try` block to replace the bare `except:` with the most appropriate, narrow exception type (e.g., `ImportError`, `KeyError`, `TypeError`, `ValueError`).
* Where a broad catch was genuinely intended (e.g., non-critical printing or plotting fallbacks), replaced it with `except Exception:` to maintain functionality while allowing system-level exits to propagate.

Fixes #945


radis/io/exomol.py
```diff
@@ -157,7 +157,7 @@ def fetch_exomol(
     # Ensure isotope format:
     try:
         isotope = int(isotope)
-    except:
+    except (ValueError, TypeError):
         raise ValueError(
             f"In fetch_exomol, ``isotope`` must be an integer. Got `{isotope}` "
             + "Only one isotope can be queried at a time. "
```
minouHub: This should simply be a "if not int() then raise error"

radis/lbl/factory.py
```diff
@@ -1160,7 +1160,7 @@ def eq_spectrum_gpu(
 
             else:
                 mol_id = self.df0.attrs["id"]
-        except:
+        except (KeyError, AssertionError, AttributeError):
```
minouHub, line 1169: AssertionError?

radis/misc/basics.py
```diff
@@ -314,7 +314,7 @@ def compare_lists(
             out = out, string
 
         return out
-    except:
+    except Exception:
```
minouHub: Clearly, there is no need for try except here. It should just be removed

radis/misc/utils.py
```diff
@@ -51,7 +51,7 @@ class Chdir:
         cd = Chdir(os.path.dirname(__file__))
         try:
             (...)
-        except:
+        except Exception:
```
minouHub: What is this doing?

minouHub (changes requested):
Positive:

- Hard work. Very good to avoid unexpected behavior of the code.
- This is needed for the long term.

Improvement to do:

- Propose actual solutions these empty `except`. Given the few remarks I gave, you should be able to either replace them by "if ... then raise Error"
- In most cases, replacing `except` by `except Exception` does not solve the entire problem and only "lets KeyboardInterrupt and SystemExit propagate" as you mentioned in #945 [EDITED]

Zeyadzahran:
Hi @minouHub, I truly appreciate the thorough review. Thanks for pointing these 
I've pushed the fixes based on your feedback. I removed the unnecessary try/except blocks and replaced them with explicit if checks where appropriate

radis/io/exomol.py
Zeyadzahran: done

radis/lbl/factory.py
Zeyadzahran: It was originally there to catch the `assert len(id_set) == 1` failure. However, taking your advice, I completely refactored `_is_at_equilibrium()` to use standard if checks instead of the whole `try/assert/except` pattern.

radis/misc/basics.py
Zeyadzahran: done

radis/misc/utils.py
Zeyadzahran: It was just an example in the docstring. I removed it.

radis/misc/basics.py
```diff
@@ -234,8 +234,6 @@ def compare_dict(
             out = out, string
 
         return out
-    except:
-        raise
```
minouHub, line 236: Either you keep the full try except block, or you fully remove it. This does not make sense

radis/misc/utils.py
```diff
@@ -51,8 +51,6 @@ class Chdir:
         cd = Chdir(os.path.dirname(__file__))
         try:
             (...)
-        except:
-            (...)
```
minouHub, line 55: Does not make sense to remove a comment part

radis/io/exomol.py
```diff
@@ -155,13 +155,14 @@ def fetch_exomol(
...
-            + "Only one isotope can be queried at a time. "
-        )
+    if not isinstance(isotope, (int, np.integer)):
+        try:
+            isotope = int(isotope)
+        except (ValueError, TypeError):
+            raise ValueError(
+                f"In fetch_exomol, ``isotope`` must be an integer. Got `{isotope}` "
+                + "Only one isotope can be queried at a time. "
+            )
```
minouHub: Good. This makes sense

radis/lbl/bands.py
```diff
@@ -1063,9 +1061,7 @@ def _calc_broadening_noneq_bands(self):
 
         self.profiler.start("calc_broadening_noneq_bands", 2)
         # Just some tests
-        try:
-            assert len(df.shape) == 2
-        except:
+        if len(df.shape) != 2:
```
minouHub: Good. This makes sense

radis/lbl/bands.py
```diff
@@ -1015,9 +1015,7 @@ def _calc_broadening_bands(self):
 
         self.profiler.start("calc_broadening_eq_bands", 2)
         # Just some tests
-        try:
-            assert len(df.shape) == 2
-        except AssertionError:
+        if len(df.shape) != 2:
```
minouHub: Good. This makes sense

radis/lbl/factory.py
```diff
@@ -2072,24 +2072,18 @@ def predict_time(self):
...
+                    return False
+            if "Trot" in self.input:
+                if not (self.input.Trot is None or self.input.Trot == self.input.Tgas):
+                    return False
+            if "overpopulation" in self.input:
+                if not (
+                    self.input.overpopulation is None or self.input.overpopulation == {}
+                ):
+                    return False
+            return True
```
minouHub, line 2094: Good. This makes sense

radis/los/slabs.py
```diff
@@ -410,7 +410,7 @@ def same_wavespace(wl):
             if all([allclose(w, wl[0]) for w in wl[1:]]):
                 return True
             return False
-        except:  # ex:  different lengths
+        except Exception:  # ex:  different lengths
             return False
         else:
             return True
```
minouHub, line 416: This could be improved for sure.

radis/tools/database.py
```diff
@@ -85,7 +85,7 @@ def is_jsonable(x):
         #        json.dumps(x)
         json_tricks.dumps(x)
         return True
-    except:
+    except (TypeError, ValueError):
```
minouHub, line 87: Are you sure this is the expected error? Why two?

minouHub (changes requested):
- Although not ideal, the conversion of `except:` to `except Exception:` has some value
- In some parts, you actually remove the expect statement. Good
- Some parts do not make sense. Please adjust

Zeyadzahran:
Is this good @minouHub

radis/io/exomol.py
```diff
@@ -155,13 +155,14 @@ def fetch_exomol(
...
-    try:
-        isotope = int(isotope)
-    except (ValueError, TypeError):
-        raise ValueError(
-            f"In fetch_exomol, ``isotope`` must be an integer. Got `{isotope}` "
-            + "Only one isotope can be queried at a time. "
-        )
+    if not isinstance(isotope, (int, np.integer)):
+        try:
+            isotope = int(isotope)
```
minouHub: you forgot to remove the try

radis/io/exomol.py
minouHub: I wonder what happened here. the `try` is now in the `if` instance.

radis/misc/utils.py
minouHub: Again: Why did you remove this part?

radis/test/validation/test_RADIS_vs_HAPI_line_broadening.py
```diff
@@ -116,9 +116,6 @@ def calc_hapi():
                 name="HAPI",
             )
 
-        except:
-            raise
-
```
minouHub, line 121: Why is this removed?

minouHub (changes requested):

Zeyadzahran:
Sorry about that, I think all your comments are addressed now, thanks @minouHub

radis/io/exomol.py
```diff
@@ -155,13 +155,8 @@ def fetch_exomol(
...
     # Ensure isotope format:
-    try:
-        isotope = int(isotope)
-    except:
-        raise ValueError(
-            f"In fetch_exomol, ``isotope`` must be an integer. Got `{isotope}` "
-            + "Only one isotope can be queried at a time. "
-        )
+    if not isinstance(isotope, (int, np.integer)):
+        isotope = int(isotope)  # raises ValueError/TypeError if not convertible
```
minouHub:
```suggestion
         raise ValueError(
            f"In fetch_exomol, ``isotope`` must be an integer. Got `{isotope}` "
            + "Only one isotope can be queried at a time. "
        )
```

radis/lbl/bands.py
```diff
@@ -1019,9 +1019,7 @@ def _calc_broadening_bands(self):
...
```
minouHub:
```suggestion
        if len(df.shape) == 2:
```

minouHub: I think there was an error in the initial test. If df is very small (i.e. two lines) than there might be a problem

radis/lbl/bands.py
```diff
@@ -1067,9 +1065,7 @@ def _calc_broadening_noneq_bands(self):
 
         self.profiler.start("calc_broadening_noneq_bands", 2)
         # Just some tests
-        try:
-            assert len(df.shape) == 2
-        except:
+        if len(df.shape) != 2:
```
minouHub:
```suggestion
        if len(df.shape) == 2:
```
Same as above

radis/lbl/broadening.py
```diff
@@ -1350,7 +1350,7 @@ def _add_Lorentzian_broadening_HWHM(
             )
             try:
                 assert bool(shift) == False
-            except:
+            except (AssertionError, ValueError):
```
minouHub: You could replace the three lines by "if shift is not None:"

radis/misc/basics.py
minouHub: Did you ignore my comment or I forgot something?

radis/misc/basics.py
```diff
@@ -312,8 +310,6 @@ def compare_lists(
...
```
minouHub, line 316: Same. Either keep the entire try except structure or remove it **entirely** (i.e. "try" included")

radis/misc/printer.py
```diff
@@ -52,8 +52,6 @@ def _capture_print(*args, **kwargs):
     try:
         print(*args, **kwargs)
         string = newstdout.getvalue()  # Get string output
-    except:
```
minouHub, line 55: Same comment again

radis/misc/printer.py
```diff
@@ -84,8 +82,6 @@ def get_print_full(x):
         pd.set_option("display.max_rows", len(x))
         print(x)
         string = newstdout.getvalue()  # Get string output
-    except:
```
minouHub, line 87: Same comment again

minouHub (changes requested):
Still a few ajustements. Apply suggested corrections when you can

Zeyadzahran:
@minouHub I addressed the changes, I think everything is good now

radis/test/validation/test_RADIS_vs_HAPI_line_broadening.py
minouHub: This was not resolved nor answer.

Zeyadzahran: Hey @minouHub, really sorry for the late reply! I completely missed the notification for this comment.

I removed the bare `except: raise` because it's redundant. I couldn't remove the entire `try` block (like you suggested in `basics.py`) because we still need it for the `finally` cleanup (`shutil.rmtree`).

minouHub: @Zeyadzahran please account for this comment

Zeyadzahran:
I have a problem in my local machine I will fix this

radis/test/validation/test_RADIS_vs_HAPI_line_broadening.py
Zeyadzahran: done @minouHub

minouHub (approved):

PR #951, opened 2026-02-27
ci: cache HITRAN databases to fix rate-limiting and mark network tests

### LINKED ISSUE #350: HITRAN connection error in CI-test : replace test database

### LINKED ISSUE #785: [CI-Tests] Coverage, acceleration, and Windows
There are three tasks related to CI on Travis.

## Tests on Windows

When performing a PR we test RADIS on 
- Linux Python 3.10
- Linux Python 3.11

When merging develop to master, we also test

- Linux Python 3.9
- Mac OS, Python 3.10

To make sure we do not release a new version that fails on Windows, the following test must be added 

- [ ] Windows, Python 3.10

## Accelerate the tests

At the moment, each test costs approximately 7 minutes of computation on Travis. Any solution to accelerate these tests would be great because 1/ we have to ask Travis for new free credits regularly and 2/ we have to wait to see the results of our PR. Potential to accelerate the tests:

- [ ] Reduce time of install (I don't know how)
- [ ] Find redondant tests (a function tested several times)
- [ ] Find deprecated functions that are still tested
- [ ] Reduce the number of databases that are downloaded (we just need to test once that our fetching functions are still working for each type of database)
- [ ] Any suggestion to reduce the duration of the longer tests

**Note:** These actions should ne reduce the test coverage.

## Test coverage

Coverage can be tracked there: https://app.codecov.io/gh/radis/radis. See also https://radis.readthedocs.io/en/develop/dev/developer.html#code-coverage
- [ ] Increase code coverage

Zeyadzahran:
### Description

This PR implements database caching in GitHub Actions to reduce HITRAN API calls and prevent rate-limiting, addressing the remaining fetching optimization mentioned in the closed issue #785.

**The Problem:**
RADIS CI tests intermittently fail when downloading databases from HITRAN servers. Because CI runners start fresh on every run and execute five jobs in parallel, they simultaneously spam HITRAN with download requests, triggering rate limiting. Since HITRAN doesn't return a standard HTTP error code when blocking us, adding retry logic isn't a practical fix.

**The Solution:**
1. **GitHub Actions Cache:** Added `actions/cache@v4` to the CI workflows to persist the downloaded databases (`~/.radisdb` and `~/radis.json`) between runs. The cache key includes the OS and Python version to avoid cross-platform HDF5 incompatibility. After the initial cache is built, future CI runs will reuse the databases instead of re-downloading them, fixing the rate-limiting at its root.
2. **Missing Network Markers:** I tracked down 8 test functions that make network calls to HITRAN but were missing the `@pytest.mark.needs_connection` marker, and added them for correctness.

Fixes #785 

[Related Slack discussion](https://radis-radiation.slack.com/archives/C9F7X0ZAT/p1772135413531809?thread_ts=1772135413.531809&cid=C9F7X0ZAT)

### DIFF
--- a/.github/workflows/test-fetch-databases.yml
+++ b/.github/workflows/test-fetch-databases.yml
@@ -1,0 +1,35 @@
+# Separate workflow to test actual database fetching from HITRAN, HITEMP, ExoMol, GEISA.
+# No cache - every run downloads fresh from the servers.
+# Known to fail intermittently due to server-side rate-limiting - this is NOT
+# a required check, but gives visibility into fetch regressions before merge.
+
+name: Test fetch from databases
+
+on:
+  push:
+    branches: [ "develop" ]
+  pull_request:
+    branches: [ "develop" ]
+
+jobs:
+  fetch:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python 3.12
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install pytest specutils
+          pip install -e .
+      - name: Test database fetching
+        env:
+          HITRAN_EMAIL: ${{ secrets.HITRAN_EMAIL }}
+          HITRAN_PASSWORD: ${{ secrets.HITRAN_PASSWORD }}
+        run: |
+          pytest -m "needs_connection and not download_large_databases" --durations=10 --tb=short
--- a/.github/workflows/tests-develop.yml
+++ b/.github/workflows/tests-develop.yml
@@ -38,6 +38,13 @@ jobs:
         uses: actions/setup-python@v3
         with:
           python-version: ${{ matrix.python-version }}
+      - name: Cache downloaded databases
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.radisdb
+            ~/radis.json
+          key: radisdb-v1-${{ matrix.os }}-py${{ matrix.python-version }}
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
--- a/.github/workflows/tests-master.yml
+++ b/.github/workflows/tests-master.yml
@@ -49,6 +49,13 @@ jobs:
         uses: actions/setup-python@v3
         with:
           python-version: ${{ matrix.python-version }}
+      - name: Cache downloaded databases
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.radisdb
+            ~/radis.json
+          key: radisdb-v1-${{ matrix.os }}-py${{ matrix.python-version }}
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
--- a/.github/workflows/uncached-fetch-tests.yml
+++ b/.github/workflows/uncached-fetch-tests.yml
@@ -1,7 +1,5 @@
-# Separate workflow to test actual database fetching from HITRAN, HITEMP, ExoMol, GEISA.
-# No cache - every run downloads fresh from the servers.
-# Known to fail intermittently due to server-side rate-limiting - this is NOT
-# a required check, but gives visibility into fetch regressions before merge.
+# Verifies live database fetching without cache (HITRAN, HITEMP, ExoMol, GEISA).
+# Non-blocking workflow: Expect occasional rate-limit failures from remote servers.before merge.
 
 name: Test fetch from databases
 
--- a/radis/test/lbl/test_broadening.py
+++ b/radis/test/lbl/test_broadening.py
@@ -926,6 +926,7 @@ def test_noneq_continuum(plot=False, verbose=2, warnings=True, *args, **kwargs):
     assert res < 5.2e-6
 
 
+@pytest.mark.needs_connection
 def test_broadening_chunksize_eq(verbose=True, plot=False, *args, **kwargs):
     """
     Test equilibrium spectra with and without chunksize,
@@ -990,6 +991,7 @@ def test_broadening_chunksize_eq(verbose=True, plot=False, *args, **kwargs):
 
 
 # @pytest.mark.fast #not fast due to connection, Nicolas Minesi 08/04/2024
+@pytest.mark.needs_connection
 def test_non_air_diluent(verbose=True, plot=False, *args, **kwargs):
     """Test collisional broadening by other species than air and self (resonant)
 
@@ -1055,6 +1057,7 @@ def test_non_air_diluent(verbose=True, plot=False, *args, **kwargs):
 
 
 # @pytest.mark.fast #not fast due to connection, Nicolas Minesi 08/04/2024
+@pytest.mark.needs_connection
 def test_diluents_molefraction(verbose=True, plot=False, *args, **kwargs):
     """
     Assert an error is raised when Molefraction (molecule + diluent) < 1 or > 1
--- a/radis/test/lbl/test_calc.py
+++ b/radis/test/lbl/test_calc.py
@@ -866,6 +866,7 @@ def test_check_wavelength_range(verbose=True, warnings=True, *args, **kwargs):
     assert np.isclose(w.max(), 5000, atol=wstep)
 
 
+@pytest.mark.needs_connection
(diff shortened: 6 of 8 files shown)


Zeyadzahran:
Can you take a look? I made the changes we [discussed](https://radis-radiation.slack.com/archives/C9F7X0ZAT/p1772135413531809?thread_ts=1772135413.531809&cid=C9F7X0ZAT), so tell me if this sounds good for you @minouHub, @Prtm2110

minouHub:
Quick review:

- Is the cache actually loaded by the next commit?
- Is the fetch of HITRAN still tested at least once?

Zeyadzahran:
Yes,`actions/cache@v4` restores both `~/.radisdb` (the `.h5` files) and `~/radis.json` (the registry) before the tests start. When `fetch_hitran()` runs, it checks `is_registered()`  finds the entry in `~/radis.json`  then `get_missing_files()` sees the files already in `~/.radisdb`  skips the download. You can see the logic in `radis/io/hitran.py`  `if local_files and not ldb.get_missing_files(local_files): download_files = False`.

Yes, the CI pytest commands don't filter out `needs_connection`, so all the HITRAN fetch tests still run. On cached runs, `fetch_hitran()` is still called end-to-end (registry check, file validation, loading), it just skips the HTTP download since files exist. And on the first run of a new cache (or after GitHub's 7-day eviction), there's no cache yet, so the full download still happens. 

Does this make sense @minouHub

minouHub:
Here is the complex issue.

- We want to avoid downloading to many times. Why? Because often, the downloading fails due to HITRAN.
- We want to test if the RADIS actually downloads from HITRAN correctly. So we need to test it. Imagine a new PR messes with the fetch. The admin could miss it during the review process because the cache is still here and skipped the download part.

I honestly have no solution to this!

Zeyadzahran:
I have a solution for this What about we keep the cache for normal PRs and pushes to avoid the rate-limiting, but we add a weekly scheduled CI run (like a cron job) that runs without the cache?
This way, daily PRs stay fast and don't spam HITRAN, but we still force a real, end-to-end download test once a week. If a PR breaks the fetch logic, the weekly run will catch it and fail.
what do you think about this @minouHub

minouHub:
With your suggestion, this test will fail once the bad PR has been merged to develop. Thus it would be too late.

Suggestion: 

- Keep the cache as you suggest here.
- Run another GitHub Actions that does the fetch for all databanks. This new GitHub action is called "Test fetch from databases" and we know it will fail sometimes

Zeyadzahran:
Can you take a look @minouHub , @Prtm2110

Prtm2110:
> Suggestion:
> 
> * Run another GitHub Actions that does the fetch for all databanks. This new GitHub action is called "Test fetch from databases" and we know it will fail sometimes

Yes, but run it on only one python version (3.14) so this will reduce fetches to one fourth. 

> GitHub Actions Cache: Added actions/cache@v4 to the CI workflows to persist the downloaded databases (~/.radisdb and ~/radis.json) between runs. The cache key includes the OS and Python version to avoid cross-platform HDF5 incompatibility. After the initial cache is built, future CI runs will reuse the databases instead of re-downloading them, fixing the rate-limiting at its root.

Does this mean that all parallel jobs will have shared `.radisdb` storage @Zeyadzahran?

Zeyadzahran:
Hey @Prtm2110  Thanks for your review 
To answer your question about the storage: no, they don't share a live directory. Each parallel job runs on its own isolated runner, and `actions/cache` just restores a local copy of the artifact for that specific job. There's no real-time shared filesystem between them, so we don't have to worry about race conditions or data corruption.

Plus, the cache is strictly partitioned to avoid any cross-platform HDF5 headaches. Notice the key setup here:
`key: radisdb-v1-${{ matrix.os }}-py${{ matrix.python-version }}`

Every OS + Python combo gets its own distinct snapshot.
As for the new fetch workflow, it's a good point, and  I  already set it up to run without a matrix to save bandwidth:

```YAML
- name: Set up Python 3.12
  uses: actions/setup-python@v5
  with:
    python-version: "3.12"
```
I currently have it on `3.12` If you want I can update it to `3.14`

Zeyadzahran:
@minouHub , @prathamhole14

prathamhole14 (changes requested):
Please change the file name to `uncached-fetch-tests.yml`

.github/workflows/test-fetch-databases.yml
```diff
@@ -0,0 +1,35 @@
+# Separate workflow to test actual database fetching from HITRAN, HITEMP, ExoMol, GEISA.
+# No cache - every run downloads fresh from the servers.
+# Known to fail intermittently due to server-side rate-limiting - this is NOT
+# a required check, but gives visibility into fetch regressions before merge.
```
prathamhole14:
```suggestion
# Verifies live database fetching without cache (HITRAN, HITEMP, ExoMol, GEISA).
# Non-blocking workflow: Expect occasional rate-limit failures from remote servers.
```

prathamhole14 (approved):

prathamhole14 (commented):
Please rebase this on develop.

Zeyadzahran:
done @prathamhole14 sorry for the late reply

dcmvdbekerom:
Hi, what is the status of this PR? Are all tests passing?
And for my understanding; the needs_connection tests are ran only during the first caching step or also afterwards?

dcmvdbekerom:
May I propose another solution, namely to use a lazy caching reverse proxy server.

There is a program called **nginx** with the `proxy_cache_lock` feature. It captures outgoing http requests, forwards only the first one to the external server, caches it, and refers all following requests to the cache. This way the python code would remain completely untouched, we only have to set up the CI to run the proxy server.

**EDIT**: oh i forgot that the 5 instances run from 5 different virtual machines, so the reverse proxy is not going to work and the caching as [Zeyadzahran](https://github.com/Zeyadzahran) proposed is indeed the best way to do it

Zeyadzahran:
> Hi, what is the status of this PR? Are all tests passing? And for my understanding; the needs_connection tests are ran only during the first caching step or also afterwards?

Hi @dcmvdbekerom, thanks for checking in!

The PR is up to date and all the cached CI jobs are passing. The only failing check is the separate Test fetch from databases workflow, which isn't related to the cache changes themselves. It's currently hitting the existing HITRAN login issue.
Regarding needs_connection, they still run as part of the normal test matrix on every run. There isn't a separate caching step in this PR. On a cache miss they download the databases and populate the cache, while on a cache hit they simply reuse the restored cache. The separate fetch workflow always runs without the cache so the download path is still exercised.

dcmvdbekerom:
The issue I see is that on a cache miss, all runners will miss at the same time, send download request to hitran.org at the same time, and we're back to the original issue. Is that correct? If so, I have the proxy working now, so that could be a solution for this.

Zeyadzahran:
Yes, that's correct. This PR optimizes the common case after the cache has been populated, but it doesn't prevent concurrent downloads on an initial cache miss. We also know that this situation will still happen sometimes.

dcmvdbekerom:
Ok in that case I suggest we move ahead with https://github.com/radis/radis/pull/1029 and close this one?

PR #954, opened 2026-03-03
Register partial HITEMP CO2 downloads in radis.json

### LINKED ISSUE #953: HITEMP CO₂ downloads are not registered in radis.json under "database"
Automatically downloaded HITEMP CO₂ databases stored in ~/.radisdb are not registered under the "database" section in radis.json, unlike other database clients (e.g., HITRAN).

Other database clients create entries in:
```
"database": {
    "DB_NAME": {
        "info": "...",
        "path": "...",
        "format": "...",
        "levelsfmt": "..."
    }
}
```
However, automatically downloaded HITEMP CO₂ databases do not appear to create similar entries.

merintheres:
Fixes #953 

Adds registration support for partial HITEMP CO2 2024 chunked downloads.
-Registers partial HITEMP CO₂ downloads in radis.json
-Introduces register_partial_hitemp_co2() helper function
-Stores per-file metadata under a new files field: 
(path, wavenumber_min, wavenumber_max, download_date, size_mb, last_used)

Example Structure:
```
"HITEMP-CO2": {
  "info": "...",
  "format": "hitemp-radisdb",
  "path": ["chunk1.h5", "chunk2.h5"],
  "wavenumber_min": "",
  "wavenumber_max": "",
  "files": [
    {
      "path": "chunk1.h5",
      "wavenumber_min": 2273,
      "wavenumber_max": 2446,
      "download_date": "2026-03-04",
      "size_mb": 270.53
      "last_used": "2026-03-06 12:31"
    }
  ]
}
```


Prtm2110:
@merintheres can't we use [HITEMPDatabaseManager.register](https://github.com/Prtm2110/radis/blob/c9c4baf732e9391d9706e7a82a0e1fc799817c10/radis/api/hitempapi.py#L1251) I understand this might not be convenient/possible given how the code is structured for HITMEP CO2 but it is creating some redundant code (its fine it we don't have any other option).

Prtm2110:
Oh you have used `register_database(...)` this is fine as well - as it will make things easier to implement. thanks!

merintheres:
@Prtm2110 @minouHub could you please review this PR

radis/io/hitemp.py
```diff
@@ -160,6 +164,18 @@ def fetch_hitemp(
             output=output,
         )
 
+        # Register the CO2 database in radis.json
+        databank_name_fmt = databank_name
+        if r"{molecule}" in databank_name_fmt:
+            databank_name_fmt = databank_name_fmt.format(**{"molecule": molecule})
```
Prtm2110: If the code is reaching here then it already means the databank_name_fmt is going to fixed something like `HITEMP-CO2-2024` so this logic can be removed.

radis/misc/config.py
```diff
@@ -619,7 +619,13 @@ def addDatabankEntries_configformat(dbname, dict_entries, verbose=True):
 
     # Allow some informative parameters
     # ... "download_url", "download_date", "wavenumber_min", "wavenumber_max" generated by fetch_hitemp
-    for entry in ["download_url", "download_date", "wavenumber_min", "wavenumber_max"]:
+    for entry in [
+        "download_url",
+        "download_date",
```
Prtm2110, line 624: `addDatabankEntries_configformat` where is this function used? Could you please explain why this is done?

radis/test/io/test_hitemp.py
```diff
@@ -592,6 +592,41 @@ def test_parse_hitemp_missing_labels_issue280(*args, **kwargs):
     print(df.dtypes["v3u"])  # Worked
 
 
+@pytest.mark.fast
+def test_register_partial_hitemp_co2():
```
Prtm2110: A better way to test this is to directly integrate the test in `test_fetch_hitemp_partial_download_CO2`. Simple test to check if the file appears in radis.json will work.

Prtm2110 (changes requested):
Thanks @merintheres I have added few minor comments. One major change I think you should do is use `register_partial_hitemp_co2` in middle of the `read_and_write_chunked_for_CO2` just before  combining dataframe)  - this will make the code more readable - I understand that `read_and_write_chunked_for_CO2` is getting too big - you can create another PR to do some refactoring of hitemp CO2. 

Also make sure you look for case where a particular file entry is already there then don't add it again but add  other files.

radis/misc/config.py
merintheres: The CO₂ registration implemented in this PR writes metadata using the JSON-based addDatabankEntries (used with radis.json). I added support for the files field there so that per-chunk metadata can be stored for partial CO₂ downloads. The INI config path is not modified.

minouHub:
This is indeed needed! Thanks for contributing, I'll let @Prtm2110 review and I'll give a look in the end.

Prtm2110:
@minouHub please have a look at this. @merintheres I have made the code little more readable and modified tests, for this sort of work you can take help of LLMs but be careful of the changes it makes. Thanks!

Prtm2110:
Yes `@minouHub`, I am done reviewing - it's good to go from my end (given the tests will pass).

Prtm2110 (approved):

merintheres:
Thanks for the improvements to the code and for the suggestions. I’ll go through the modifications to make sure I fully understand them. Thank you for the guidance!

PR #956, opened 2026-03-04
Fix TODOs: add error/warning messages and fix bare except

tanishadahale:
## Summary

Implements 4 TODO items from the codebase:

1. **nistapi.py**: Add `ValueError` when NIST database returns no spectral lines
2. **spectrum.py**: Add warning when slit FWHM >= wstep/5 (may cause inaccurate convolution)
3. **overp.py**: Add warning when rescaling factor > 30% (large rescaling may be inaccurate)
4. **exomolapi.py**: Replace bare `except:` with specific `(OSError, pd.errors.ParserError)`

### DIFF
--- a/radis/lbl/overp.py
+++ b/radis/lbl/overp.py
@@ -677,7 +677,19 @@ def rescale_updown_levels(
     when inferring emission quantities from absorption quantities this may
     ends up in error in overpopulation rescaling.
     """
-    # TODO: Add warning when too large rescaling
+    # Warn if rescaling is too large (>30% change)
+    if old_nu != 0 and old_nl != 0 and not ignore_warnings:
+        rescale_nu = abs(new_nu / old_nu - 1)
+        rescale_nl = abs(new_nl / old_nl - 1)
+        if rescale_nu > 0.3 or rescale_nl > 0.3:
+            from warnings import warn
+
+            warn(
+                f"Large rescaling detected: nu factor={new_nu/old_nu:.2f}, "
+                f"nl factor={new_nl/old_nl:.2f}. For large changes, "
+                "calculate a new spectrum instead.",
+                UserWarning,
+            )
 
     # Check inputs
     # ---------
--- a/radis/lbl/overp.py
+++ b/radis/lbl/overp.py
@@ -682,8 +682,6 @@ def rescale_updown_levels(
         rescale_nu = abs(new_nu / old_nu - 1)
         rescale_nl = abs(new_nl / old_nl - 1)
         if rescale_nu > 0.3 or rescale_nl > 0.3:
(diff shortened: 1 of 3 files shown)


radis/spectrum/spectrum.py
```diff
@@ -3365,6 +3365,15 @@ def apply_slit(
         w = self._q["wavespace"]  # non convoluted wavespace
         wstep = abs(diff(w)).min()
         assert wstep > 0
+
+        # Warn if slit FWHM is large compared to spectral sampling
+        if isinstance(slit_function, (int, float)) and slit_function >= wstep / 5:
```
minouHub: Is it not already checked [here](https://github.com/radis/radis/blob/b69333f969ea8f2f5473303a8053c0c228f37e6d/radis/tools/slit.py#L463)?

minouHub: here is a check for you
```
s = calc_spectrum(
    3500, 3600,         # cm-1
                  molecule='CO2',
                  isotope='1',
                  pressure=1.01325,   # bar
                  Tgas=500,           # K
                  mole_fraction=1,
                  path_length=5,      # cm
                  name='500 K',
                  wstep=0.02,
                  )
s.apply_slit(0.02, 'cm-1')       # simulate an experimental slit
s.plot()
```

minouHub (changes requested):
See comments in the code

radis/spectrum/spectrum.py
tanishadahale: You're right, there's already a check in slit.py:461 that raises an error when the slit has < 5 points. I've removed this TODO from the PR since it's redundant.

radis/lbl/overp.py
```diff
@@ -677,7 +677,19 @@ def rescale_updown_levels(
     when inferring emission quantities from absorption quantities this may
     ends up in error in overpopulation rescaling.
     """
-    # TODO: Add warning when too large rescaling
+    # Warn if rescaling is too large (>30% change)
+    if old_nu != 0 and old_nl != 0 and not ignore_warnings:
+        rescale_nu = abs(new_nu / old_nu - 1)
+        rescale_nl = abs(new_nl / old_nl - 1)
+        if rescale_nu > 0.3 or rescale_nl > 0.3:
+            from warnings import warn
```
Prtm2110: warn should be a module level import.

tanishadahale: Done, moved the warn import to the module level.

Prtm2110 (approved):
@tanishadahale thanks for the PR it looks good to me @minouHub please have a look.

minouHub:
Nice job

PR #957, opened 2026-03-04
Fix LineSurvey to respect lineinfo for unknown database formats

tanishadahale:
`lineinfo` was ignored for unknown database formats in `LineSurvey`, which caused all columns to appear in hover labels. This change updates `get_label_all` to accept a `columns` parameter so that only the columns specified in `lineinfo` are displayed. Now the behavior for unknown database formats is consistent with HITRAN/CDSD formats.

### DIFF
--- a/radis/tools/line_survey.py
+++ b/radis/tools/line_survey.py
@@ -301,19 +301,19 @@ def LineSurvey(
                 val = f"{row[k]:.3g}"
             else:
                 val = f"{row[k]}"
-            label += f"<br>{k} {name}: {val} {unit}"
+            label += f"<br><b>{k}</b> {name}: {val} {unit}"
             # If lower value, also show upper value on same line
             if k[-1] == "l" and k[:-1] + "u" in details:
                 k_up = k[:-1] + "u"
                 if is_float(row[k_up]):
                     label += (
                         "&nbsp;" * (30 - len(k) - len(val))
-                        + f"{k_up} {name}: {row[k_up]:.3g} {unit}"
+                        + f"<b>{k_up}</b> {name}: {row[k_up]:.3g} {unit}"
                     )
                 else:
                     label += (
                         "&nbsp;" * (30 - len(k) - len(val))
-                        + f"{k_up} {name}: {row[k_up]} {unit}"
+                        + f"<b>{k_up}</b> {name}: {row[k_up]} {unit}"
                     )
 
         return label
@@ -359,7 +359,7 @@ def LineSurvey(
                     iso = "?"
 
                 label = (
-                    "{molec}[iso{iso}] [{branch}{jl:.0f}]({vl:.0f})->({vu:.0f})".format(
+                    "{molec}[iso{iso}] {branch}({vl:.0f},{jl:.0f})".format(
                         **dict(
                             [(k, row[k]) for k in add]
                             + [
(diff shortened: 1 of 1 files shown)


minouHub:
Could you show a printscreen of the situation before and now? And the code to reproduce it please?

tanishadahale:
code to reproduce:
```
from radis.tools.line_survey import LineSurvey

# Create a spectrum with an unknown database format
s = calc_spectrum(2000, 2100, molecule='CO', Tgas=700, pressure=1, export_lines=True)
s.conditions['dbformat'] = 'unknown'

LineSurvey(s, lineinfo=['wav', 'S', 'A'], writefile="result.html")
```

### Before
image

### After
image

minouHub (commented):
I got this Warning. Can you take a look? [EDIT: sorry, problem on my side]

minouHub:
This is very nice. I'd like to ask for improvement I've been hoping to see for some time as a user. If I can ask ...🙏 

1/ In the tool "line survey" of spectraplot.com, they have a very nice layout of the quantities: **name of quantity (in bold)**: value of quantity. Is there a way to do that on our side as well?
image

2/ We could even do better and show the name of the line. 

- Branch. -1 is P, 0 is Q, +1 is R
- jl is low vibrational state
- vl is low rotational state

For the example, in HITRAN, it gives "P[31(0)->(1)]" when it should show "P(0,31)" @erwanp maybe you can tell us why you used this convention? 
image


but not for ExoMol

image

tanishadahale:
Hi Nicolas, I’ve added the requested features in #957
-  Bold labels
-  Improved spectroscopic notation for HITRAN/CDSD.
-  ExoMol labels improved: now show P(0,31) header with bold key:value instead of raw dumps

I also had a few enhancement ideas inspired by the earlier SpectraPlot examples you shared, let me know if you’d like to include them as well:

1. Show both wavenumber and wavelength in the tooltip
SpectraPlot displays ν (cm⁻¹) and λ (μm) together, while RADIS currently shows only one.

2. Show lower energy (E'') in the tooltip for all formats
SpectraPlot highlights E''. RADIS already has El, but it’s not easily visible.

3. Add “Copy to clipboard” on click
Similar to SpectraPlot’s feature that lets users copy plot data by clicking.

4. Add export buttons (Save PNG / Save CSV)

minouHub:
We are in the improvement zone now. Using this code, I got:
```
from radis.tools.line_survey import LineSurvey

s = calc_spectrum(2007, 2009, molecule='CO', Tgas=700, pressure=1, export_lines=True)

LineSurvey(s, 
           # lineinfo=['wav', 'S', 'A'], 
           writefile="result.html")
```
image

- I do not think that the first line `(wav, S)` is necessary anymore as these info are repeated below
- I do not think that the explanation of the quantities is necessary either
- Isotope number is an int, not a float
- Can you put the grec letter "ν" instead of "wavenumber"? It would be more compact (btw, there is no "vacuum wavenumber". A wavenumber is always the same) 
- "int" should be "S" for linestrength
- "El" should be displayed as E'' (more conventional in spectroscopy)
- Exomol: can you apply the same fix?

minouHub:
I think it should be the last comment:

**General**

- [ ] Don't you find that, in the exponent -1 or -2, the number is a bit below the "-" sign? Could you move it up a bit?

**HITRAN**
image

- [ ] Could you print E'' as a float "{:.0f}" and not as an exponent?

**ExoMol**
image

- [ ] Could you add "CO P(0,31)" like with HITRAN?

tanishadahale:
Hello @minouHub 
I've implemented all the requested changes,

HITRAN
image

EXOMOL
image

- HTML `<sup>` for superscript alignment 
- E″ as integer 
- molecule name + branch for ExoMol tooltips
- hint text moved inside hover tooltip, subtle "✓ Copied" toast at the bottom of the page.
image

minouHub:
This is great!

PR #948, opened 2026-02-26
ci: enable codecov upload for travis

Zeyadzahran:
### Description
This pull request addresses the missing coverage reporting from Travis CI and the documentation examples, completing the CI tracking architecture for CodeCov.

**Key Changes:**

1. **Travis CI Upload Fix:** Replaced the deprecated and broken bash uploader (`bash <(curl -s https://codecov.io/bash)`) in `.travis.yml` with  `codecov-cli` Python package. Travis coverage is now successfully uploaded under the `travis` flag.
2. **Docs Examples Coverage:** Instead of modifying ReadTheDocs (which is not natively a test runner), I introduced a new GitHub Actions workflow (`docs-coverage.yml`). This workflow replicates the RDT `sphinx-gallery` build, wraps it in `coverage run`, and uploads the resulting XML to CodeCov under the `docs` flag. This captures all the "free tests" run during documentation generation.
3. **Codecov Configuration:** Updated `codecov.yml` to define individual flags (`unittests`, `travis`, `docs`) and display them in the PR comment. Crucially, I enabled `carryforward: true` for all flags. This ensures that if a specific CI doesn't run (e.g., Travis skipping on fork PRs due to missing secrets), Codecov will gracefully use the last known coverage instead of artificially dropping the coverage percentage.

Fixes #947

PR #960, opened 2026-03-06
ci: fix Travis coverage upload to radis/radis Codecov project

### LINKED ISSUE #947: Codecov covers several CI tools
### 💭 Description

Radis relies on GitHub actions, Travis and RDT for CI. We use three different CI because:

- GitHub actions are extremely practical
- Travis can handle secret variables (for HITEMP login) and GPU calculation. 
- RDT is a good documentation tool (obviously) and running examples is also an opportunity to run tests, even though this is not the main objective.

Since PR #909, [CodeCov](https://app.codecov.io/gh/radis/radis) only uses the tests run by GitHub actions. What has to be done:

- [x] Include the coverage of Travis in CodeCov
- [ ] Include the coverage of ReadTheDocs in CodeCov

Zeyadzahran:
### Description

This is clean pr replace part of #948

Fixes #947

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,7 +57,10 @@ script:
   - xvfb-run -a env PYTHONPATH="." pytest -m "needs_gpu" --durations=10 --cov=radis --cov-report=xml:coverage.xml
   - echo ">>> Run tests require HITRAN/HITEMP credentials (write access only)"
   - xvfb-run -a env PYTHONPATH="." pytest -m "needs_HITRAN_credentials and not download_large_databases" --durations=10 --cov=radis --cov-report=xml:coverage.xml --cov-append
-  - bash <(curl -s https://codecov.io/bash) -f coverage.xml -t $CODECOV_TOKEN -F travis -v
+  # Upload coverage to Codecov. Clear CODECOV_TOKEN to prevent routing to the
+  # wrong project (the token in Travis settings belongs to a fork).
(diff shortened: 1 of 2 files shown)


Zeyadzahran:
Take a look @minouHub

Zeyadzahran:
I checked https://app.codecov.io/gh/radis/radis/pull/960/blob/radis/lbl/factory.py
The function name is green, but the body is red. I have an explanation for this, but I don't know if it's correct
Python executes def eq_spectrum_gpu() when the module is imported (it creates the function object). GitHub Actions imports factory.py during its tests, so this line shows as covered. The function body only runs when eq_spectrum_gpu() is actually called. Only Travis GPU tests call it (@pytest.mark.needs_gpu). GitHub Actions explicitly skips GPU tests (not needs_gpu).
Could you confirm if my guess here is right, @minouHub, @Prtm2110?

minouHub:
> The function name is green, but the body is red. I have an explanation for this, but I don't know if it's correct

It worked!
https://app.codecov.io/gh/radis/radis/blob/Zeyadzahran%2Fradis%3Afix-travis-codecov-upload/radis%2Flbl%2Ffactory.py
image

You explanation makes sense. I guess you had to wait for Travis to upload the report?

Zeyadzahran:
Yes exactly. Travis just needed some time. Glad it finally worked

PR #965, opened 2026-03-09
Implement LRU cache eviction for HITEMP CO2 chunk downloads

### LINKED ISSUE #962: Implement LRU cache eviction for HITEMP CO2 chunk downloads
Downloaded HITEMP CO2 chunks stored in the local .radisdb directory can accumulate over time since they are not automatically removed, which may lead to increased disk usage.
An LRU (Least Recently Used) cache mechanism can be added to manage these downloads using metadata already stored in radis.json (path, size_mb, last_used). When the cache size exceeds a configurable limit, older unused chunks should be removed while keeping recently used data.

merintheres:
Fixes #962

- Introduces an LRU (Least Recently Used) cache eviction mechanism for HITEMP CO2 chunk downloads stored in the local `.radisdb` directory.
- Downloaded chunks are tracked in `radis.json`.
- Adds a new configuration key `HITEMP_CO2_CACHE_LIMIT_MB` to control the cache size limit (default: 5 GB).
- After chunk processing completes, the total cache size is computed and compared against the configured limit.
- If the cache exceeds the limit, cached chunks are sorted by their `last_used` timestamp and the least recently used files are evicted.
- Chunks overlapping the current user-requested wavenumber range are protected from eviction.
- Metadata is cleaned by removing entries whose files no longer exist on disk.
- The updated metadata is then written back to `radis.json`.

### DIFF
--- a/radis/api/hitempapi.py
+++ b/radis/api/hitempapi.py
@@ -843,6 +843,13 @@ def read_and_write_chunked_for_CO2(
 
             pbar.update(1)
 
+    # Run LRU cache eviction
+    evict_lru_cache(
+        user_wmin=load_wavenum_min,
+        user_wmax=load_wavenum_max,
+        verbose=verbose,
+    )
+
     # Combine DataFrames
     if dataframes:
         printer.info("Combining parsed data from all chunks...", indent=1)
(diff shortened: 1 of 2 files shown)


merintheres:
@Prtm2110, Could you please take a look at the implementation and let me know if the approach looks good?

Prtm2110:
@merintheres Your code might be functionally correct but what makes this change difficult is the fact that the code should be potentially extendable to other clients as well. That is either you should create a separate new class for handling LRU cache system or integrate this functionality in the `DatabaseManager` so the code is clean and good enough so that it can be easily maintained later. @minouHub what are your views on this?

Prtm2110:
Are you sorting the list every time when a new entry is added? If so then are you adding the new file at the top?

merintheres:
Currently new chunks are appended during registration, so the order in radis.json follows the download_date order and not last_used.
During eviction, the entries are copied into a temporary list and sorted by last_used to determine the eviction order. Since the number of chunks is small, sorting once during eviction is easier than keeping the list always sorted and updating it every time. Since you mentioned future extensibility, we could also implement the registration to store the entries sorted by last_used if that would be preferable.

merintheres:
Regarding the architecture, making the cache reusable would be better. I’ll refactor the implementation to integrate it into DatabaseManager.

minouHub:
Is this PR still active given our last conversation in #962 ?

minouHub:
Closing. For the moment, the policy will be to let the user clean up the space by hand

PR #973, opened 2026-03-14
refactor: deprecate optically_thin parameter

### LINKED ISSUE #972: Remove `optically_thin` option from `Spectrum.rescale`
Hi, 

I was looking at the long-term todo list in #53, and I noticed the task : 
clean code : remove optically_thin option in Spectrum.rescale ? --> useless, complicates the code for nothing. Could be simulated by setting transmittance_noslit=0 in the user code.

I will work on this todo

Zeyadzahran:
### Description
This pull request is to address the clean-up task for the `optically_thin` parameter while ensuring full backward compatibility for existing users in the LONG TERM TODO #53 

**Key changes:**

* **Graceful Deprecation:** The `optically_thin` parameter in `Spectrum.update()` (`spectrum.py`) and `rescale.py` now defaults to `None` and raises a `DeprecationWarning` if used. Legacy code will still run perfectly under the hood!
* **API Simplification:** Users are now guided to use the single source of truth (`conditions['self_absorption'] = False`) instead of relying on a hidden side-effect parameter.
* **Internal Updates:** Updated internal logic (like `MergeSlabs` in `slabs.py`) and test suites (`test_spectrum.py`) to natively use the condition-based approach.

Fixes #972

### DIFF
--- a/radis/los/slabs.py
+++ b/radis/los/slabs.py
@@ -823,9 +823,9 @@ def MergeSlabs(*slabs, **kwargs) -> Spectrum:
                 )
 
         # Add the rest of the spectral quantities afterwards:
+        s.conditions["self_absorption"] = not optically_thin
         s.update(
             [k for k in requested if k not in ["emisscoeff", "abscoeff"]],
-            optically_thin=optically_thin,
             verbose=verbose,
         )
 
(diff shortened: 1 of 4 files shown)


Zeyadzahran:
Can you take a look @minouHub, @Prtm2110

PR #977, opened 2026-03-16
Fix temporary database downloads being written to working directory

### LINKED ISSUE #976: Temporary downloads written to working directory
While downloading database chunks in DatabaseManager.download_and_parse(), temporary files are written to the current working directory because only the filename is used when creating the file path. This causes raw downloaded files (e.g., .par.bz2, .zip) to appear in the repository or execution directory.

merintheres:
Temporary downloads were written to the current working directory because only the filename was used when constructing the temporary file path in DatabaseManager.download_and_parse(). This change saves temporary downloads in the existing downloads__can_be_deleted directory inside the local RADIS database folder.

Fixes #976

### DIFF
--- a/radis/api/dbmanager.py
+++ b/radis/api/dbmanager.py
@@ -456,10 +456,11 @@ class DatabaseManager(object):
                 response.raise_for_status()  # Raise an error if request fails
 
                 # Create a temporary file to store the downloaded content
-                temp_file_path = urlname.split("/")[-1]
-                temp_file_path = re.sub(
-                    r'[<>:"/\\|?*&=]', "_", temp_file_path
+                temp_fname = urlname.split("/")[-1]
+                temp_fname = re.sub(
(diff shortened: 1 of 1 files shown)


Prtm2110:
@merintheres can you provide example code to reproduce this issue? I guess you are talking about H20 right - I have faced this similar bug before.

Prtm2110 (approved):
Thank you this was needed! This PR also solves the problem of those .zip file not being deleted automatically. 

```
from radis import fetch_hitemp
df = fetch_hitemp(
    "H2O",
    load_wavenum_min=0,  # cm-1
    load_wavenum_max=100,
    # verbose=3,
    # return_local_path=True,
)
df
```

merintheres:
Yes, the code you shared triggers the issue. Thank you!

minouHub:
Good catch!

PR #980, opened 2026-03-18
Add warning for out-of-range wavenumber requests

merintheres:
Adds a warning in hdf2df when the requested wavenumber range is outside the database range.
Uses existing metadata and skips cases where range is unavailable.

Implements : add Warning if using a line database outside its min/max range ( #53 )

### DIFF
--- a/radis/api/hdf5.py
+++ b/radis/api/hdf5.py
@@ -6,6 +6,7 @@ Defines the :py:class:`~radis.api.hdf5.DataFileManager` class
 import os
 import pathlib
 import sys
+import warnings
 from os.path import abspath, exists, expanduser, splitext
 from time import time
 
@@ -915,6 +916,24 @@ def hdf2df(
 
     # Read and add metadata in the DataFrame
     metadata = manager.read_metadata(fname)
+
+    # warning when requested wavenumber range lies outside DB metadata
+    md = metadata[0] if isinstance(metadata, list) and len(metadata) > 0 else metadata
+    if isinstance(md, dict):
+        db_min = md.get("wavenumber_min") or None
+        db_max = md.get("wavenumber_max") or None
+
+        req_min = load_wavenum_min
+        req_max = load_wavenum_max
+
+        if db_min is not None and db_max is not None:
+            if (req_min is not None and req_min < db_min) or (
+                req_max is not None and req_max > db_max
+            ):
+                warnings.warn(
+                    f"Requested range [{req_min}, {req_max}] outside database range [{db_min}, {db_max}]",
+                    UserWarning,
+                )
 
     # Sanity Checks if loading the full file
     selection = isotope or load_wavenum_min or load_wavenum_max
--- a/radis/api/hdf5.py
+++ b/radis/api/hdf5.py
@@ -920,8 +920,8 @@ def hdf2df(
     # warning when requested wavenumber range lies outside DB metadata
     md = metadata[0] if isinstance(metadata, list) and len(metadata) > 0 else metadata
     if isinstance(md, dict):
-        db_min = md.get("wavenumber_min") or None
-        db_max = md.get("wavenumber_max") or None
+        db_min = md.get("wavenumber_min", None)
+        db_max = md.get("wavenumber_max", None)
 
         req_min = load_wavenum_min
         req_max = load_wavenum_max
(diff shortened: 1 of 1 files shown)


minouHub:
Hello,
If I use this code, I already have an error message. Can you give an example where your warning is actually raised?
```python
s = calc_spectrum(0, 3,
                  molecule='CO',
                  isotope='1',
                  pressure=1.01325 * u.bar,
                  Tgas=350 * u.K,
                  mole_fraction=1.0,
                  path_length=5.5 * u.cm,
                  databank='GEISA',
                  verbose=False,
                  )
```

Also, I skimmed quickly #53 and didn't see the to-do related to this PR. Can you help?

minouHub:
> Hello, If I use this code, I already have an error message. Can you give an example where your warning is actually raised?
> 
> ```python
> from astropy import units as u
> s = calc_spectrum(0, 3,
>                   molecule='CO',
>                   isotope='1',
>                   pressure=1.01325 * u.bar,
>                   Tgas=350 * u.K,
>                   mole_fraction=1.0,
>                   path_length=5.5 * u.cm,
>                   databank='GEISA',
>                   verbose=False,
>                   )
> ```
> 
> Also, I skimmed quickly #53 and didn't see the to-do related to this PR. Can you help?

minouHub:
Sorry, wrong click! Just wanted to know if you had an update

radis/api/hdf5.py
```diff
@@ -923,6 +917,24 @@ def hdf2df(
     # Read and add metadata in the DataFrame
     metadata = manager.read_metadata(fname)
 
+    # warning when requested wavenumber range lies outside DB metadata
+    md = metadata[0] if isinstance(metadata, list) and len(metadata) > 0 else metadata
+    if isinstance(md, dict):
+        db_min = md.get("wavenumber_min") or None
+        db_max = md.get("wavenumber_max") or None
+
```
Zeyadzahran, line 932: Watch out for the or None trap here! Since 0 is falsy in Python, if wavenumber_min happens to be exactly 0, this will accidentally evaluate to None. 

Using the default argument in .get() is much safer and avoids this bug.

```suggestion
        db_min = md.get("wavenumber_min", None)
        db_max = md.get("wavenumber_max", None)
```

radis/api/hdf5.py
```diff
@@ -923,6 +917,24 @@ def hdf2df(
...
+        if db_min is not None and db_max is not None:
+            if (req_min is not None and req_min < db_min) or (
+                req_max is not None and req_max > db_max
+            ):
+                warnings.warn(
+                    f"Requested range [{req_min}, {req_max}] outside database range [{db_min}, {db_max}]",
+                    UserWarning,
+                )
+
```
Zeyadzahran, line 945: Adding stacklevel=2 here is good practice so the warning points directly to the user's script instead of this internal line in hdf5.py.

```suggestion
            warnings.warn(
                f"Requested range [{req_min}, {req_max}] outside database range [{db_min}, {db_max}]",
                UserWarning,
                stacklevel=2,
            )

**Review by Zeyadzahran — CHANGES_REQUESTED** [review]
I pulled this branch and tested it locally to investigate the issue raised by @minouHub.

The crash reported earlier occurs because the test range (0-3 cm-1) triggers an EmptyDatabaseError during the online fetch stage, before ever reaching `hdf2df`. Additionally, I found that local `.par` files converted to .h5 usually lack the `wavenumber_min/max` metadata keys, which is why the warning doesn't trigger for them.

The PR logic **is functional** for HDF5 files that contain the proper metadata. I verified this with the following snippet:

```python
import pandas as pd
from radis.api.hdf5 import hdf2df, DataFileManager

h5 = "test_warning.h5"
df = pd.DataFrame({"wav": [2000, 2100], "int": [1e-20, 2e-20], "iso": [1, 1]})

manager = DataFileManager("pytables")
manager.write(h5, df, append=False)
manager.add_metadata(h5, {"wavenumber_min": 2000.0, "wavenumber_max": 2300.0})

hdf2df(h5, load_wavenum_min=1900, load_wavenum_max=2400)
```
result
image

 
I've also included a few minor comments on the code itself to handle edge cases and improve the user experience.

Zeyadzahran:
> Also, I skimmed quickly #53 and didn't see the to-do related to this PR. Can you help?

It's the last todo mentioned under new (physical) features

merintheres:
@minouHub sorry for the late reply!
I checked this further and confirmed that the warning logic works when hdf2df is reached with valid HDF5 metadata. In the example it fails earlier during the fetch stage, so the warning isn’t triggered. 
I also applied @Zeyadzahran’s suggestions  to improve the behavior.

minouHub:
Alright, the PR does what you wanted but ... could you provide an example

PR #982, opened 2026-03-19
fix: Spectrum.plot(wunit='nm') respects medium for correct wavelength range

### LINKED ISSUE #707: Wavelength range calculated wrong by `Spectrum.get(wunit='nm')` in some cases
### 🐛 Describe the bug

The wavelength array is wrongly calculated in some cases. For instance, modifying the basic example in the README slightly to this, where `2937.6340295525983 cm-1` is equal to `3404.1 nm` and `2937.9792578664396 cm-1` is equal to `3403.7 nm`:
```python
s = calc_spectrum(2937.6340295525983, 2937.9792578664396,         # cm-1
                  molecule='CH4',
                  isotope='1',
                  pressure=1.01325,   # bar
                  Tgas=400,           # K
                  mole_fraction=0.02,
                  path_length=2,      # cm
                  databank='hitran',   # or use 'hitemp'
                  verbose=False
                  )
s.plot('transmittance_noslit', wunit='nm') 
```
Gives the following plot:

bugradis

Note that not even the labels in `cm-1` at the top of the figure are the ones requested in the code. On the other hand, if `nm` units are not requested, the spectrum is calculated correctly:

bugradis2

### 💡 Possible solutions

I just discovered `radis`, so I'm not very familiar with the codebase. I have no idea as to why this happens, but at the moment I am doing the conversion between `cm-1` and `nm` myself with this function and it works very well:

```python
def wavelength_to_wavenumber(wl: float) -> float:
    """
    Convert wavelength to wavenumber.

    Parameters
    ----------
    wl : float
        Wavelength in nm.

    Returns
    -------
    float
        Wavenumber in cm^-1.
    """
    return 1 / wl * 1e7


def wavenumber_to_wavelength(wn: float) -> float:
    """
    Convert wavenumber to wavelength.

    Parameters
    ----------
    wn : float
        Wavenumber in cm^-1.

    Returns
    -------
    float
        Wavelength in nm.
    """
    return 1 / wn * 1e7
```

It could probably also be done using `astropy` quantities, but for this fits my needs nicely at the moment.

### 🎲 Radis version

0.15.2

### 💻 Operating system

Linux

tanishadahale:
### Description

The issue was that `medium` wasn’t stored in `SpectrumFactory.input`, so it never reached `Spectrum.conditions`. As a result, `get(wunit='nm')` always assumed `"air"`, leading to incorrect conversions and plot axes.

This PR fixes it by propagating `medium` correctly and using it in conversions. It:

* Stores `medium` in `SpectrumFactory.input`
* Uses `conditions["medium"]` in `get(wunit='nm')`
* Applies correct air/vacuum conversions in plots
* Adds support for `'nm_vac'` and `'nm_air'`

Also added a regression test and verified across cm⁻¹, nm (air), nm (vacuum), and `nm_vac`.

Fixes #707

image

### DIFF
--- a/radis/lbl/base.py
+++ b/radis/lbl/base.py
@@ -3963,6 +3963,8 @@ def get_wavenumber_range(
         If unitless, wunit is assumed as the accompanying unit.
     wunit: string
         The unit accompanying wmin and wmax. Cannot be passed without passing values for wmin and wmax.
+        Accepted values: ``'cm-1'``, ``'nm'``, ``'nm_air'``, ``'nm_vac'``.
+        ``'nm_air'`` and ``'nm_vac'`` override the ``medium`` parameter.
         Default: cm-1
     wavenum_min, wavenum_max: float, or `~astropy.units.quantity.Quantity` or ``None``
         wavenumbers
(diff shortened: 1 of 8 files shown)


minouHub (changes requested):
You should update the failing test as the wavelength in vacuum has been corrected
Will be merged in 0.17.1

PR #986, opened 2026-03-20
test: optimize test_eq_vs_noneq_isotope to use local offline database

Zeyadzahran:
### Description

This pull request is to address the inline TODO: "replace `test_eq_vs_noneq_isotope` example not to download CO2 from HITRAN ; use CO HITEMP instead ?".  in the LONG TERM TODO #53  to stop downloading CO2 from HITRAN by switching the test to use the local `HITRAN-CO-TEST` database for CO instead of HITEMP, completely removing the network dependency without needing extra credentials.

### DIFF
--- a/radis/test/lbl/test_calc.py
+++ b/radis/test/lbl/test_calc.py
@@ -580,7 +580,6 @@ def test_all_calc_methods_CO2pcN(
     #    assert np.isclose(s_nq.get_power(), s_eq.get_power(), rtol=rtol)
 
 
-@pytest.mark.needs_connection
 def test_eq_vs_noneq_isotope(verbose=True, plot=False, warnings=True, *args, **kwargs):
     """Test same spectrum for 2 different calculation codes (equilibrium,
     non-equilibrium) in the presence of isotopes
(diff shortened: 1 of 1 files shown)


minouHub:
Non-equilibrium CO2 computation must be tested. You could use HITRAN-CO2 instead

Zeyadzahran:
Understood, I'll update it right away.

minouHub:
Another comment: I'm afraid that this PR won't reduce the duration of tests because HITRAN CO2 is downloaded in several other places. I'll merge if there is a clear demonstration that the duration of the tests is reduced

Zeyadzahran:
Fair enough. I'll run the benchmarks to compare the performance before and after the change. If the results don't show a significant improvement in test duration, I'm okay with closing this PR. I'll ping you once I'm finished.

Zeyadzahran:
Hey @minouHub, 

I ran the benchmarks. Here are the numbers:
- Online Fetch (cached): ~1.8s
- Local DB: ~1.6s

So the speedup is only about 0.2s since the cache is already doing the heavy lifting.
However, since this still fully isolates the test from network issues and ticks off one of the tasks in #53, do you think it’s still worth merging? 

Happy to go with whatever you prefer.

minouHub:
Not an important change compared to https://github.com/radis/radis/pull/994. Closing

PR #990, opened 2026-04-02
Rename basefactory populations

Zeyadzahran:
### Description

This PR addresses the TODO: `rename base.py and BaseFactory to more explicit populations.py and PopulationFactory ?` in LONG TERM TODO #53.

The name `PopulationFactory` makes a lot more sense for what this class actually does. 

**Changes made:**
- Renamed `radis/lbl/base.py` to `radis/lbl/populations.py`.
- Renamed the `BaseFactory` class to `PopulationFactory` globally.
- Renamed `test_base.py` to `test_populations.py` and updated all internal imports, references, and docs.
- **Backward compatibility:** Added a deprecated alias (`from .populations import PopulationFactory as BaseFactory`) in `radis/lbl/__init__.py` so we don't break existing user code after the update.


Zeyadzahran:
I realized we needed a compatibility shim to handle historical imports (like `from radis.lbl.base import ...`). I added `base.py` back, but I used `__getattr__` for the warning. This way, it only triggers if someone specifically imports the deprecated `BaseFactory`, which prevents spamming users who just need unchanged helper functions from that old path.

Zeyadzahran:
Can you please take a look @minouHub , @Prtm2110

minouHub:
This PR introduces important changes to the core of the code for a limited impact on the user side or physics. I prefer to keep this for the long-term not-that-essential to-do list

PR #991, opened 2026-04-03
fix: convert README from RST to Markdown and fix PyPI deployment

tanishadahale:
### Description

<h2>What was broken</h2>
<p>The PyPI deployment had several issues stacking together:</p>
<ol>
<li>
<p><strong>README not included in sdist</strong>: <code>MANIFEST.in</code> included <code>README.rst</code> but then <code>global-exclude *.rst</code> removed it. The package was shipping without a README.</p>
</li>
<li>
<p><strong>Pandoc workaround failing silently</strong>: The workflows were converting <code>README.rst → README.md</code> because <code>pyproject.toml</code> referenced a non-existent <code>README.md</code>. This had two problems:</p>
<ul>
<li><code>apt-get install pandoc</code> was missing <code>-y</code> (could hang in CI)</li>
<li>The generated <code>README.md</code> never made it into the source distribution anyway</li>
</ul>
</li>
<li>
<p><strong>RST rendering errors masked by pandoc</strong>: Once the conversion workaround was removed, <code>twine check</code> revealed errors that pandoc had been hiding:</p>
<ul>
<li>A stray <code>::</code> in the GPU Acceleration section</li>
<li>Emoji headings (<code>🌱</code>, <code>🔬</code>) with underlines that were too short</li>
<li><code>:scale:</code> image directives that require PIL (not supported by PyPI's renderer)</li>
</ul>
</li>
<li>
<p><strong>TestPyPI re-uploading on every push</strong>: The TestPyPI workflow triggered on every push to <code>develop</code>/<code>master</code>, re-uploading the same version repeatedly.</p>
</li>
</ol>
<h2>What this PR does</h2>
<ul>
<li>Converts <code>README.rst</code> to <code>README.md</code> (as suggested by @erwanp — simpler and avoids all RST rendering issues)</li>
<li>Fixes <code>MANIFEST.in</code>: includes <code>README.md</code>, uses <code>prune</code> instead of <code>global-exclude</code> for <code>docs/</code> and <code>examples/</code> (matches original intent, more reliable)</li>
<li>Removes pandoc conversion from both PyPI and TestPyPI workflows</li>
<li>Adds version tag validation step in PyPI workflow</li>
<li>Changes TestPyPI trigger to version tags only</li>
<li>Removes obsolete <code>rst_to_md.py</code> and stale RST comments from <code>pyproject.toml</code></li>
</ul>
<h2>Verification</h2>
<p>Reproduced failures locally, then verified fixes:</p>

Test | Result
-- | --
python -m build + twine check dist/* | PASSED (wheel + sdist)
README.md in sdist, no .rst files | PASSED
Wheel install in fresh venv + import radis | PASSED
Sdist install + metadata check | PASSED (text/markdown)
Pre-commit (black, isort, flake8) | PASSED

@Prtm2110


Zeyadzahran (commented):
I have some comments 

- in pyproject.toml: I'm worried PyPI won't render the Markdown correctly. Should we explicitly set `content-type = "text/markdown"` in the metadata to be safe?
- in MANIFEST.in: Why switch to `prune examples`? Pruning the whole tree seems a bit much if we want users to have access to examples in the sdist. 
- in publish-Pypi.yml: Using `github.event.release.tag_name` is probably more reliable for release events .
- in exomolapi.py: Silent NaN filling could mask some serious data issues. Can we add a `UserWarning` here so users know fallback values are being used? Otherwise, it might lead to physically wrong results without anyone noticing.

tanishadahale:
@Zeyadzahran

1. It’s already included;  [see here:](https://github.com/tanishadahale/radis/blob/26021b05cf712842966e5b41cd3d035cb8ba8d5d/pyproject.toml#L18)

2. That’s incorrect. The original `MANIFEST.in` used `global-exclude examples/*`; we replaced it with `prune examples`, which more reliably removes the entire directory. Both approaches ensure examples aren’t included in the sdist. Examples are maintained in the docs, so there’s no behavior change. they were already excluded before my changes.

3. This would only be a cosmetic change; functionality remains the same.
4. not part of this pr, removed

Zeyadzahran:
1. While `readme = "README.md"` often works, explicitly setting `content-type = "text/markdown"` follows **PEP 621** standards. to ensure PyPI (and other tools) render the README correctly. Without it, we risk the project page showing raw text instead of formatted content. [ckeck this ](https://peps.python.org/pep-0621/)

2. Even if examples are in the docs, researchers often need them offline inside the source download (sdist). `prune` deletes the entire folder, which is too aggressive. Let's keep the folder structure so users have everything they need in one place.

3. Using github.event.release.tag_name is the canonical way to get the tag. GITHUB_REF_NAME is a general variable that could point to a branch instead of a tag if we ever update the workflow triggers.
4. I recommend opening an issue first to define the problem and the proposed fix; this will keep things organized.

@tanishadahale

tanishadahale:
1. Wrong. PEP 621 literally says: "If readme is a file path, the tool MUST set the content-type based on the file extension" 
`.md → text/markdown, .rst → text/x-rst`. 
Setting it explicitly is the old setup.cfg way. With pyproject.toml, `readme = "README.md"` is the correct PEP 621 approach.
I already verified the built package has Description-Content-Type: text/markdown.

2. Wrong. The original MANIFEST.in already had `global-exclude examples/*` ; examples were NEVER in the sdist. We just changed the mechanism. prune is actually more reliable because global-exclude examples/* only matches files, not subdirectories. Also, pip install radis gives a wheel (no examples either way). 

3. agreed, github.event.release.tag_name is more explicit and won't break if someone changes the trigger.

4. Process preference anyway, I had already discussed this issue with the mentors and they recommended opening a PR. The PR description provides sufficient context, but I’ll still include the points discussed with them.

minouHub:
Hello,
I got the upload to TestPypi working. (For info, the problem was the environment name which was incorrect on [testPypi](https://test.pypi.org/project/radis/)).

PR #1001, opened 2026-05-24
fix(#1000): crop() boundary 0.0 truthiness bug

### LINKED ISSUE #1000: crop() silently fails when wmin=0.0 or wmax=0.0
### 🐛 Describe the bug

```python


### 💡 Possible solutions

_No response_

### 🎲 Radis version

The current version is 0.17

### 💻 Operating system

MacOS

### PR DESCRIPTION
**Pull request by Omiii-215**
### Description

- Replaced truthiness checks in crop() with explicit `is not None` checks
- ~~Fixed upstream circular import with track_ref in spectrum.py~~ [EDIT @minouHub]
- Added regression test for zero-value boundaries

Fixes #<Issue Number>
#1000

### CODE REVIEW

File: radis/spectrum/spectrum.py [hunk 1]
```diff
@@ -88,7 +88,20 @@
...
+_RefTracker = None
+
+
+def _get_RefTracker():
+    global _RefTracker
+    if _RefTracker is None:
+        from radis.tools.track_ref import RefTracker
+
+        _RefTracker = RefTracker
+    return _RefTracker
```
minouHub: Why did you need this?

radis/spectrum/spectrum.py
```diff
@@ -559,7 +572,7 @@ def __init__(
         self.profiler = None
 
         # Add references
-        self.references = RefTracker(**references)
+        self.references = _get_RefTracker()(**references)
```
minouHub: Needed? Related?

radis/test/spectrum/test_operations.py
```diff
@@ -66,6 +66,74 @@ def test_crop(verbose=True, *args, **kwargs):
...
+    assert w_crop.min() >= 0.0, f"wmin=0.0 was not applied: got w.min()={w_crop.min()}"
+    assert w_crop.max() <= 5.0
+
+    # --- Test wmax = 0.0 ---
+    s_crop2 = s.crop(wmin=-5.0, wmax=0.0, wunit="cm-1", inplace=False)
+    w_crop2 = s_crop2.get("transmittance_noslit", wunit="cm-1")[0]
+    assert (
+        w_crop2.max() <= 0.0
+    ), f"wmax=0.0 was not applied: got w.max()={w_crop2.max()}"
+    assert w_crop2.min() >= -5.0
```
minouHub: A negative wavenumber does not make sense. Either use another value or prevent a negative value

radis/test/spectrum/test_operations.py
```diff
@@ -66,6 +66,74 @@ def test_crop(verbose=True, *args, **kwargs):
...
+                # The purpose of this call is to hit the `is not None` branch in crop().
+                s_unit.crop(wmin=0.0, wmax=None, wunit=test_wunit, inplace=False)
+            except Exception:
+                pass
+
+            # Use wmin=None, wmax=0.0
+            try:
+                s_unit.crop(wmin=None, wmax=0.0, wunit=test_wunit, inplace=False)
+            except Exception:
+                pass
```
minouHub: Good idea. However, you should check that the proper error is raised, not use try/except, see https://stackoverflow.com/questions/129507/how-do-you-test-that-a-python-function-throws-an-exception

radis/test/spectrum/test_operations.py
```diff
@@ -66,6 +66,74 @@ def test_crop(verbose=True, *args, **kwargs):
...
+    """Regression test for GitHub issue #1000.
+
+    ``crop()`` used ``if wmin:`` instead of ``if wmin is not None:`` to guard
+    boundary checks. Because ``0.0`` is falsy in Python, passing ``wmin=0.0``
+    silently skipped the crop and returned the original (uncropped) spectrum.
+    """
+    from radis.spectrum.spectrum import Spectrum
+
+    # Create a spectrum that spans negative-to-positive wavenumbers
+    w = np.linspace(-10, 10, 201)
```
minouHub: No negative value

minouHub (changes requested):
Thank you. Please apply requested changes

minouHub (approved):
Please rebase on develop (the tests should pass now)

PR #1002, opened 2026-05-26
fix: use is not None checks in crop() to handle wmin=0.0

thakoreh:
## Problem
The `crop()` function in `radis/spectrum/operations.py` uses Python truthiness checks (`if wmin:`, `if wmax:`, `if wmin0:`, `if wmax0:`) to test whether boundary values are provided. When `wmin=0.0` or `wmax=0.0`, these checks evaluate as falsy, causing the crop filter to silently skip that boundary.

This means calling `crop(s, wmin=0.0, wmax=500.0, wunit='nm')` returns the full spectrum unchanged.

## Fix
Changed all truthiness checks to use explicit `is not None` checks.

## Testing
Added test_crop_zero.py with tests for wmin=0.0 and wmax=0.0 edge cases.

Note: Could not run tests locally (Python 3.9, radis requires 3.10+). CI should validate.

### DIFF
--- a/radis/spectrum/operations.py
+++ b/radis/spectrum/operations.py
@@ -393,14 +393,14 @@ def crop(s: Spectrum, wmin=None, wmax=None, wunit=None, inplace=False) -> Spectr
     if stored_waveunit == "cm-1":
         # convert wmin, wmax to wavenumber
         if wunit == "nm":
-            if wmax0:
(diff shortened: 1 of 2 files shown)


thakoreh:
Closing this while I pause open-source contributions. Apologies for the noise.

PR #1011, opened 2026-06-05
fix: use 'is not None' for crop() boundary checks to handle wmin/wmax=0.0

### LINKED ISSUE #1000: crop() silently fails when wmin=0.0 or wmax=0.0

duongynhi000005-oss:
## Problem
 in `radis/spectrum/operations.py` uses bare truthiness checks like `if wmin:` and `if wmax:` to determine whether to apply boundary filtering. Since `0.0` is falsy in Python, passing `wmin=0.0` or `wmax=0.0` silently skips that boundary.

Fixes #1000

## Fix
Replace all 14 occurrences of `if wmin:`, `if wmax:`, `if wmin0:`, `if wmax0:` with their `is not None` equivalents throughout the unit-conversion and crop logic.

## Changes
- `radis/spectrum/operations.py`: 14 truthiness → `is not None` checks (lines 396-449)

### DIFF
--- a/radis/spectrum/operations.py
+++ b/radis/spectrum/operations.py
@@ -393,14 +393,14 @@ def crop(s: Spectrum, wmin=None, wmax=None, wunit=None, inplace=False) -> Spectr
     if stored_waveunit == "cm-1":
         # convert wmin, wmax to wavenumber
         if wunit == "nm":
-            if wmax0:
+            if wmax0 is not None:
                 wmin = nm_air2cm(wmax0)  # note: min/max inverted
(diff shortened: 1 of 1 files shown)


minouHub:
Sorry, I was quite busy and could not review before. This should be solved by https://github.com/radis/radis/pull/1001

minouHub:
Solved by #1001

PR #1014, opened 2026-06-14
fix: handle wmin=0.0 and wmax=0.0 in crop()

### LINKED ISSUE #1000: crop() silently fails when wmin=0.0 or wmax=0.0

Malaybhai11:
Fixes #1000

`crop()` was using `if wmin:` / `if wmax:` to check boundaries, but `0.0` is falsy in Python, so passing `wmin=0.0` silently skips the crop.

Changed all boundary checks to use `is not None` instead of truthiness. Also added a test case that crops with `wmin=0.0` to verify the fix.

### DIFF
--- a/radis/spectrum/operations.py
+++ b/radis/spectrum/operations.py
@@ -393,14 +393,14 @@ def crop(s: Spectrum, wmin=None, wmax=None, wunit=None, inplace=False) -> Spectr
     if stored_waveunit == "cm-1":
         # convert wmin, wmax to wavenumber
         if wunit == "nm":
-            if wmax0:
+            if wmax0 is not None:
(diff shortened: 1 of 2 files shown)


Malaybhai11:
Hey team! Quick PR here that handles the edge case where both wmin=0.0 and wmax=0.0 are passed to crop(). Currently this throws an unexpected error, but with the fix it gracefully returns an empty result range instead. Would love a review when someone gets a chance!'

minouHub:
Sorry, I was quite busy and could not review before. This should be solved by https://github.com/radis/radis/pull/1001

PR #1019, opened 2026-07-01
test: increase relative tolerance for Exomol absorption coefficient integration test

prathamhole14:
As titled: 
To fix: https://app.travis-ci.com/github/radis/radis/jobs/639901897


prathamhole14:
I just need to confirm if this is an actual fix and not just a bandage!

minouHub:
ExoMol recently pushed another recommended CO database... and the difference increased. Thus, I prefer to enforce Li2015 to make this won't happen again next time they push another database

PR #1025, opened 2026-07-05
Prevent Tk error in CI running with windows

minouHub:
Added argument in the test yaml to prevent Tk error in CI running with windows


minouHub:
Windows tests passed twice. Merging

PR #1026, opened 2026-07-11
feat: add option to drop unlabelled CO2 HITEMP lines in calc_spectrum

### LINKED ISSUE #509: todo: add quick-fix for unlabelled HITEMP lines
### 💭 Description

HITEMP-CO2 has some unlabelled lines; therefore the following non-eq calculations won't work : 

```python
calc_spectrum(2000, 3000,
              molecule='CO2',
              Trot=3000,
              isotope=1,
              Tvib=(3000, 3000, 1000),
              databank='hitemp')
```

RADIS already suggests the user a fix, which requires to use SpectrumFactory : 

```
using HITEMP2010 for CO2, some lines are unlabelled and therefore cannot be used at equilibrium. This is a known issue of the HITEMP database and will soon be fixed in the edition. In the meantime you can use:
 'sf.df0.drop(sf.df0.index[sf.df0['v1u']==-1], inplace=True)' where 'sf' is SpectrumFactory object
```
See https://github.com/radis/radis/issues/508 for an example of where it's used 

It works but requires to change the user script. Would be better if an option was added in `calc_spectrum()` to automatically fix this. 

## Implementation

- let's not declare a new parameter in calc_spectrum ; it would add more info for too few use-cases;  instead let's make it a "secret" parameter: look-up if there is a key in ``kwargs`` 
```
drop_unlabelled_CO2_hitemp_lines=kwargs.pop("drop_unlabelled_CO2_hitemp_lines", False) #  False is the default value 
```

Then add this in calc_spectrum code : 
```
if drop_unlabelled_CO2_hitemp_lines: 
    sf.df0.drop(sf.df0.index[sf.df0['v1u']==-1], inplace=True)
```

and also update the suggested tip in the error message (wherever it is) : 
``` 
using HITEMP2010 for CO2, some lines are unlabelled and therefore cannot be used at equilibrium. This is a known issue of the HITEMP database and will soon be fixed in the edition. In the meantime you can use:
 'sf.df0.drop(sf.df0.index[sf.df0['v1u']==-1], inplace=True)' where 'sf' is SpectrumFactory object ;   or use `calc_spectrum(..., drop_unlabelled_CO2_hitemp_lines=True)` 
```

prathamhole14:
Fixes: https://github.com/radis/radis/issues/509

### Description
When computing non-equilibrium spectra with HITEMP for CO2, some lines are unlabelled (quantum numbers like `v1u` or `v1l` are `-1`). These lines cause `Evib1u=NaN` and `Evib1l=NaN` errors during the `assert_no_nan` energy level validation.

### DIFF
--- a/radis/lbl/base.py
+++ b/radis/lbl/base.py
@@ -331,14 +331,16 @@ class BaseFactory(DatabankLoader):
                 "This usually happens during the computation of Evib and Erot. "
                 "To solve this issue, you can perform a dry run in a try block. "
                 "Then, drop incomplete rows in the line dataframe and run again. "
-                "See also https://github.com/radis/radis/pull/942."
+                "See also https://github.com/radis/radis/pull/942.\n"
                 "Example:\n"
                 "\n"
                 "try:\n"
                 "    s1 = sf.non_eq_spectrum(Tvib=9800, Trot=8500) #dry run to compute all energy columns\n"
                 "except AssertionError:\n"
                 '    bad_cols = [c for c in sf.df0.columns if "Evib" in c or "Erot" in c]\n'
-                "    sf.df0.dropna(subset=bad_cols, inplace=True)"
+                "    sf.df0.dropna(subset=bad_cols, inplace=True)\n"
+                "\n"
+                "Alternatively, if you are using `calc_spectrum`, you can simply use the `drop_unlabelled_CO2_hitemp_lines=True` argument."
             )
 
         if self.dataframe_type == "pandas":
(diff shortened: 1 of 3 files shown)


minouHub:
I'm sorry but I'm going to re-close this issue and your PR for the following reason:

> The user must understand what is done to its database. 

Adding another argument to `calc_spectrum`, which is already too furnished, is not the philosophie I want to pursue. `calc_spectrum` should be simple. This option is not really needed and the error message provides a clear instruction on what must be done. Since @erwanp did not defend his idea, I prefer to force the user to copy-paste one line and at least we are sure everybody is well aware of what is done :)

Issue #1003, opened 2026-05-27
recalc_gpu gives different results if multiple GPU objects are initialized


audren-dor:
### 🐛 Describe the bug

The eq_spectrum_gpu() and recalc_gpu() methods work great when used on a single molecule. However, when trying to model the spectra of two or more molecules, strong discrepencies are observed between the expected signal and the resulting one. 

To model the spectra of two molecules, two SpectrumFactory objects are first initialized. Then the method eq_spectrum_gpu is called for one of the two molecules, leading to the expected result. Then, eq_spectrum_gpu is called again for the other molecule. If recalc_gpu is used for the first molecule (in the context of a fitting procedure for example), the resulting spectrum is different from the first one despite the same parameters.

Here is below a script that shows this behavior for a NO absorption spectrum:

```python
import matplotlib.pyplot as plt
from radis import SpectrumFactory

# %%
WN_FIT_MIN = 1700
WN_FIT_MAX = 2200

# %%
def Load_SpectrumFactory_NO_H2O(wavenum_min = 1700,
                                 wavenum_max = 2200,
                                 wstep = 0.1,
                                 pressure = 1.01325,
                                 cutoff = 1e-25,
                                 path_length = 1000):
    
    # NO
    sf_NO = SpectrumFactory(wavenum_min = wavenum_min,
                            wavenum_max = wavenum_max,
                            wstep       = wstep,
                            molecule    = 'NO',
                            isotope     = 'all',
                            pressure    = pressure,
                            path_length = path_length,
                            cutoff      = cutoff,
                            broadening_method = 'voigt',
                            medium      = 'air',
                            warnings    = {'AccuracyError': 'ignore'})
    
    
    # H2O
    sf_H2O = SpectrumFactory(wavenum_min = wavenum_min,
                             wavenum_max = wavenum_max,
                             wstep       = wstep,
                             molecule    = 'H2O',
                             isotope     = 'all',
                             pressure    = pressure,
                             path_length = path_length,
                             cutoff      = cutoff,
                             broadening_method = 'voigt',
                             medium      = 'air',
                             warnings    = {'AccuracyError': 'ignore'})

    #
    sf_NO.load_databank('HITRAN-NO')
    sf_H2O.load_databank('HITRAN-H2O')
    
    return sf_NO, sf_H2O


# initialize the classical cpu spectrum factory
sf_NO, sf_H2O = Load_SpectrumFactory_NO_H2O(wavenum_min = WN_FIT_MIN,
[...]

PR #1027, opened 2026-07-12
fix random segfault in GPU tests by clamping the FFT workgroup count

### LINKED ISSUE #1024: CI - Intermittent bug of Vulkan in Travis
### 🐛 Describe the bug

Segmentation faults occurs randomly but recurrently during the GPU tests in Travis, see [example](https://app.travis-ci.com/github/radis/radis/jobs/639938635).

```
radis/test/gpu/test_gpu.py Fatal Python error: Segmentation fault
Thread 0x000070cd2ee8bb80 (most recent call first):
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/vulkan/_vulkan.py", line 11024 in _callApi
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/vulkan/_vulkan.py", line 11668 in vkWaitForFences
  File "/home/travis/build/radis/radis/radis/gpu/vulkan/vulkan_compute_lib.py", line 898 in runCommandBuffer
  File "/home/travis/build/radis/radis/radis/gpu/vulkan/vulkan_compute_lib.py", line 276 in run
  File "/home/travis/build/radis/radis/radis/gpu/gpu.py", line 357 in iterate
  File "/home/travis/build/radis/radis/radis/lbl/factory.py", line 1265 in eq_spectrum_gpu
  File "/home/travis/build/radis/radis/radis/test/gpu/test_gpu.py", line 49 in test_geisa_cpu_gpu_equivalence
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/_pytest/python.py", line 159 in pytest_pyfunc_call
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/pluggy/_callers.py", line 103 in _multicall
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/pluggy/_manager.py", line 120 in _hookexec
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/pluggy/_hooks.py", line 513 in __call__
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/_pytest/python.py", line 1627 in runtest
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/_pytest/runner.py", line 174 in pytest_runtest_call
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/pluggy/_callers.py", line 103 in _multicall
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/pluggy/_manager.py", line 120 in _hookexec
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/pluggy/_hooks.py", line 513 in __call__
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/_pytest/runner.py", line 242 in <lambda>
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/_pytest/runner.py", line 341 in from_call
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/_pytest/runner.py", line 241 in call_and_report
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/_pytest/runner.py", line 132 in runtestprotocol
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/_pytest/runner.py", line 113 in pytest_runtest_protocol
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/pluggy/_callers.py", line 103 in _multicall
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/pluggy/_manager.py", line 120 in _hookexec
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/pluggy/_hooks.py", line 513 in __call__
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/_pytest/main.py", line 362 in pytest_runtestloop
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/pluggy/_callers.py", line 103 in _multicall
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/pluggy/_manager.py", line 120 in _hookexec
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/pluggy/_hooks.py", line 513 in __call__
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/_pytest/main.py", line 337 in _main
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/_pytest/main.py", line 283 in wrap_session
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/_pytest/main.py", line 330 in pytest_cmdline_main
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/pluggy/_callers.py", line 103 in _multicall
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/pluggy/_manager.py", line 120 in _hookexec
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/pluggy/_hooks.py", line 513 in __call__
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/_pytest/config/__init__.py", line 175 in main
  File "/home/travis/virtualenv/python3.11.2/lib/python3.11/site-packages/_pytest/config/__init__.py", line 201 in console_main
  File "/home/travis/virtualenv/python3.11.2/bin/pytest", line 8 in <module>
Extension modules: charset_normalizer.md, charset_normalizer.cd, requests.packages.charset_normalizer.md, requests.packages.chardet.md, requests.packages.charset_normalizer.cd, requests.packages.chardet.cd, h5py._errors, h5py.defs, h5py._objects, h5py.h5, numpy._core._multiarray_umath, numpy.linalg._umath_linalg, h5py.utils, h5py.h5t, h5py.h5s, h5py.h5ac, h5py.h5p, h5py.h5r, h5py._npystrings, h5py._proxy, h5py._conv, h5py.h5z, h5py.h5a, h5py.h5d, h5py.h5ds, h5py.h5g, h5py.h5i, h5py.h5o, h5py.h5f, h5py.h5fd, h5py.h5pl, h5py.h5l, h5py._selector, pyarrow.lib, numpy.random._common, numpy.random.bit_generator, numpy.random._bounded_integers, numpy.random._mt19937, numpy.random.mtrand, numpy.random._philox, numpy.random._pcg64, numpy.random._sfc64, numpy.random._generator, pandas._libs.tslibs.ccalendar, pandas._libs.tslibs.np_datetime, pandas._libs.tslibs.dtypes, pandas._libs.tslibs.base, pandas._libs.tslibs.nattype, pandas._libs.tslibs.timezones, pandas._libs.tslibs.fields, pandas._libs.tslibs.timedeltas, pandas._libs.tslibs.tzconversion, pandas._libs.tslibs.timestamps, pandas._libs.properties, pandas._libs.tslibs.offsets, pandas._libs.tslibs.strptime, pandas._libs.tslibs.parsing, pandas._libs.tslibs.conversion, pandas._libs.tslibs.period, pandas._libs.tslibs.vectorized, pandas._libs.ops_dispatch, pandas._libs.missing, pandas._libs.hashtable, pandas._libs.algos, pandas._libs.interval, pandas._libs.lib, pyarrow._compute, pandas._libs.ops, numexpr.interpreter, pandas._libs.hashing, pandas._libs.arrays, pandas._libs.tslib, pandas._libs.sparse, pandas._libs.internals, pandas._libs.indexing, pandas._libs.index, pandas._libs.writers, pandas._libs.join, pandas._libs.window.aggregations, pandas._libs.window.indexers, pandas._libs.reshape, pandas._libs.groupby, pandas._libs.json, pandas._libs.parsers, pandas._libs.testing, tables._comp_lzo, tables._comp_bzip2, tables.utilsextension, tables.hdf5extension, tables.linkextension, tables.lrucacheextension, tables.tableextension, tables.indexesextension, yaml._yaml, psutil._psutil_linux, pyarrow._fs, pyarrow._azurefs, pyarrow._hdfs, pyarrow._gcsfs, pyarrow._s3fs, vaex.vaexfast, PIL._imaging, PIL._imagingft, kiwisolver._cext, lxml._elementpath, lxml.etree, _cyutility, scipy._cyutility, scipy._lib._ccallback_c, numba.core.typeconv._typeconv, numba._helperlib, numba._dynfunc, numba._dispatcher, numba.core.runtime._nrt_python, numba.np.ufunc._internal, numba.experimental.jitclass._box, scipy.linalg._fblas, scipy.linalg._flapack, scipy.linalg.cython_lapack, scipy.linalg._cythonized_array_utils, scipy.linalg._solve_toeplitz, scipy.linalg._batched_linalg, scipy.linalg._decomp_lu_cython, scipy.linalg._matfuncs_schur_sqrtm, scipy.linalg._matfuncs_expm, scipy.linalg._linalg_pythran, scipy.linalg.cython_blas, scipy.linalg._decomp_update, _cffi_backend, scipy.sparse._sparsetools, _csparsetools, scipy.sparse._csparsetools, scipy.sparse.linalg._dsolve._superlu, scipy.sparse.linalg._eigen.arpack._arpacklib, scipy.sparse.linalg._propack, scipy.optimize._group_columns, scipy._lib.messagestream, scipy.optimize._trlib._trlib, scipy.optimize._lbfgsb, _moduleTNC, scipy.optimize._moduleTNC, scipy.optimize._slsqplib, scipy.optimize._minpack, scipy.optimize._lsq.givens_elimination, scipy.optimize._zeros, scipy._lib._uarray._uarray, scipy.special._ufuncs_cxx, scipy.special._ellip_harm_2, scipy.special._special_ufuncs, scipy.special._gufuncs, scipy.special._ufuncs, scipy.special._specfun, scipy.special._comb, scipy.linalg._decomp_interpolative, scipy.optimize._bglu_dense, scipy.optimize._lsap, scipy.spatial._ckdtree, scipy.spatial._qhull, scipy.spatial._voronoi, scipy.spatial._hausdorff, scipy.spatial._distance_wrap, scipy.spatial.transform._rotation_cy, scipy.spatial.transform._rigid_transform_cy, scipy.optimize._direct, scipy.integrate._odepack, scipy.integrate._quadpack, scipy.integrate._vode, scipy.integrate._dop, scipy.signal._sigtools, scipy.signal._max_len_seq_inner, scipy.signal._upfirdn_apply, scipy.signal._spline, scipy.interpolate._fitpack, scipy.interpolate._dfitpack, scipy.interpolate._dierckx, scipy.interpolate._ppoly, scipy.interpolate._interpnd, scipy.interpolate._rbfinterp_pythran, scipy.interpolate._rgi_cython, scipy.ndimage._nd_image, scipy.ndimage._rank_filter_1d, _ni_label, scipy.ndimage._ni_label, scipy.signal._sosfilt, scipy.special.cython_special, scipy.stats._stats, scipy.stats._biasedurn, scipy.stats._stats_pythran, scipy.stats._levy_stable.levyst, scipy.stats._ansari_swilk_statistics, scipy.sparse.csgraph._tools, scipy.sparse.csgraph._shortest_path, scipy.sparse.csgraph._traversal, scipy.sparse.csgraph._min_spanning_tree, scipy.sparse.csgraph._flow, scipy.sparse.csgraph._matching, scipy.sparse.csgraph._reordering, scipy.stats._sobol, scipy.stats._qmc_cy, scipy.stats._rcont.rcont, scipy.stats._qmvnt_cy, scipy.signal._peak_finding_utils (total: 202)
Segmentation fault
The command "xvfb-run -a env PYTHONPATH="." pytest -m "needs_gpu" --durations=10 --cov=radis --cov-report=xml:coverage.xml" exited with 139.
```

### 💡 Possible solutions

_No response_

### 🎲 Radis version

0.17.1

### 💻 Operating system

Linux

Zeyadzahran:
`setFFTWorkGroupSize()` is passed a number of batches (`N_G * N_L`), but writes it into the
indirect buffer as a number of workgroups. For short transforms VkFFT puts several batches
in one workgroup, so the surplus workgroups ran past the end of `S_klm_d`  a random SIGSEGV
inside the FFT shader, surfacing as a crash in `vkWaitForFences`.

On the GEISA test the plan is built for 6 batches and VkFFT asks for 1 workgroup; we were
overwriting that with 4. Wide-range spectra use one workgroup per batch, which is why only the
small-range test crashed.

Fix: remember the workgroup counts VkFFT planned (it writes them when the transform is
recorded) and clamp to them. Where VkFFT uses one workgroup per batch this is a no-op, so the
usual path is unchanged.

Fixes #1024

### DIFF
--- a/radis/gpu/vulkan/pyvkfft_vulkan.py
+++ b/radis/gpu/vulkan/pyvkfft_vulkan.py
@@ -320,7 +320,7 @@ class VkFFTApp(VkFFTAppBase):
 
         shape = np.ones(vkfft_max_fft_dimensions(), dtype=vkfft_long_type)
         skip = np.zeros(vkfft_max_fft_dimensions(), dtype=vkfft_long_type)
-        grouped_batch = -np.ones(vkfft_max_fft_dimensions(), dtype=vkfft_long_type)
+        grouped_batch = np.ones(vkfft_max_fft_dimensions(), dtype=vkfft_long_type)
 
         shape[0] = self.shape[-1]
         # skip[1 : len(self.shape)] = 1


Zeyadzahran:
@minouHub

dcmvdbekerom:
Hello! thanks for working on this and good job figuring this out. 
I'd like to review it before it gets merged

radis/gpu/vulkan/pyvkfft_vulkan.py
```diff
@@ -416,21 +424,17 @@ def ifft(self, cmd_buf, src):
...
+
+        N is a number of batches, not a number of workgroups: for short transforms
+        VkFFT puts several batches in a single workgroup. Dispatching more workgroups
+        than the plan was built for makes the FFT shader run past the end of the
+        buffer (issue #1024).
+        """
+        # wg.id is the axis (x, y or z) VkFFT put the batch number on.
+        for wg, n_planned in zip(self.indirectHost, self._plannedWorkGroups):
+            if 0 < wg.id < 3 and n_planned > 0:
+                setattr(wg, "xyz"[wg.id], min(N, n_planned))
```
dcmvdbekerom: Keep in mind that during normal operation the size needs to be able to grow, so min(N, n_planned) will not work in that case.

Moreover, if the current FFT plan is batched in a single workgroup, and the number is increased, the FFT likely needs to be replanned.

radis/gpu/vulkan/pyvkfft_vulkan.py
```diff
@@ -397,6 +398,13 @@ def fft(self, cmd_buf, src):
             None,
         )
 
+        # VkFFT fills the indirect buffer with the workgroup counts of its plan while the
+        # transform is being recorded. Keep them: they are the most we can dispatch.
+        self._plannedWorkGroups = [
+            getattr(wg, "xyz"[wg.id]) if 0 < wg.id < 3 else 0
```
dcmvdbekerom: ~Here we should quickly assess whether VkFFT put multiple batches in a single workgroup, we will need that info later in setFFTWorkGroupSize().~

dcmvdbekerom (commented):
I see the idea of this fix, but it is currently incompatible with a growing batch size. The option of a growing batch size is required since the changing pressure/temperature broadening may necessitate a larger matrix. 

If we have multiple batches in a single workgroup, we need to acommodate the option to replan the FFT if/as needed.

Perhaps instead it is easier solution is to tell VkFFT to never combine multiple batches in a single workgroup. This will make it slightly slower for small spectra, but at these sizes the GPU kernels are likely not the bottleneck anyway (although this would have to be profiled to be sure). I think this may be possible by passing the proper configuration flags to VkFFT.

Edit: setting grouped_batch to 1 instead of -1(max) would do exactly that:
[        grouped_batch = -np.ones(vkfft_max_fft_dimensions(), dtype=vkfft_long_type)](https://github.com/radis/radis/blob/631509b9d7628aea4bfcadba62c9b39ca2caba85/radis/gpu/vulkan/pyvkfft_vulkan.py#L323)
 You can see if it solves the issue without impacting the performance too much

dcmvdbekerom:
Another thought: instead of clipping the FFT size, it might be safer to extend the buffer size to whatever is needed. Could you see if this could provide a solution?

Zeyadzahran:
Hi @dcmvdbekerom, Thanks for the detailed review and the suggestions. I took your comments and tested the growth scenario.

The specific growth case you mentioned actually seems to work with the clamp. When the batch count grows, GPUBuffer.setBatchSize() calls `resizeBuffer()`, which clears _fftAppFwd and re-records the command buffer. That rebuilds the FFT plan, so the _plannedWorkGroups snapshot is refreshed after every resize. Since the plan is built for the full buffer capacity (1.5× the current batch count), min(N, n_planned) never under-dispatched in my tests.

I also checked the parameter generation, and for the GEISA case, the batch count (N_G * N_L) never actually grows over the tested pressure/temperature range.

That said, after trying your suggestion I think grouped_batch = 1 is the cleaner solution. It removes the assumption entirely instead of defending against it, lets setFFTWorkGroupSize() behave as originally intended, and removes the _plannedWorkGroups bookkeeping and the clamp.

I profiled it on lavapipe as well:

The original crash no longer reproduces (0/12 crashes, compared to 6/12 on develop).
The GEISA tests all pass.
The performance impact is very small (~1.38 => 1.46 ms for the short FFT, ~1.91 => 1.97 ms for the long FFT), so the trade-off appears negligible.

I'll update the PR to use the grouped_batch = 1 approach. Thanks again for the suggestion!

dcmvdbekerom:
> That rebuilds the FFT plan, so the _plannedWorkGroups snapshot is refreshed after every resize. Since the plan is built for the full buffer capacity (1.5× the current batch count), min(N, n_planned) never under-dispatched in my tests.

Ahh yep you're right! Indeed with this in mind your solution should work!
But I agree setting grouped_batch = 1 is a little neater, and the performance loss is acceptable.

There's still a test failing in radis/test/validation/test_RADIS_vs_HAPI_line_broadening.py, is that related to this PR? When that one's passing this PR is ready to be merged.

Since you clearly have a good command over both Vulkan and the RADIS code, perhaps you want to take a stab at https://github.com/radis/radis/issues/1003 after this one :)?

Zeyadzahran:
Sure :)  I'd be happy to take a look at #1003.

I've updated the PR to use `grouped_batch = 1`.
Regarding `test_RADIS_vs_HAPI_line_broadening.py`, I don't think it's related to this PR. It's a pre-existing issue caused by HAPI/HITRAN rate limiting, and I've seen the same failure on develop as well. I already opened PR #951 to help reduce this issue.

dcmvdbekerom:
@minouHub Let's prioritize this since it caused the merge to main test to fail

minouHub:
I'm honestly not relevant to review this PR. The GPU world is yours @dcmvdbekerom. If you approve this PR, please merge it.

PR #1029, opened 2026-07-16
Cache with proxy

### LINKED ISSUE #785: [CI-Tests] Coverage, acceleration, and Windows

dcmvdbekerom:
### Description

This is a continuation of https://github.com/radis/radis/pull/951 by @Zeyadzahran.
While in the original pull request the data is cached per runner, this means the data can still be requested in parallel when there's a cache miss, which is to be avoided.

Combining github caching with a reverse proxy server, and caching the cache of the proxy server to github, looks like the most promising solution. In this case the following happens:

1. tests with @needs_connection are ran first. Requests to hitran.org are funneled through the proxy, which stores a cache of the downloaded files. The requests will miss the cache due to the github cache having a unique key per run, in other words, the data is downloaded fresh. The proxy cache is saved at the end of this job.
2. Now the regular test (including the ones marked @needs_connection) are ran. The hitran requests are again funneled through the proxy, but since its cache is still there, they can all be served locally. Because the raw files are cached, it's not a problem they were downloaded with one OS and used by another.

**This is still a draft** but I need to make a PR already to trigger the CI tests

Fixes https://github.com/radis/radis/issues/785

*Main issues:*
- [x] Make working proxy in Ubuntu
- [x] Github Caching on the proxy cache
- [x] Make the Win & Mac runners work with the cached data
- [x] Fix certificate issue with urllib3 / requests
- [x] Test proxy should be a 'real' test
- [x] Fix test_plot_all_CO2_bandheads
- [x] Fix Hitran login issue with proxy server
- [x] Deselect needs_authorization tests for users without credentials
- [x] Make a single nginx_conf file with variable cache path
- [x] Make shell scripts for proxy setup
- [x] Fix Windows charmap issue
- [x] gitignore nginx.conf
- [ ] Fuse mac and linux proxy setup?
- [ ] Fix hitemp partial download issue
- [ ] Add needs_caching decorator
- [ ] Update master test workflow

*Cosmetic issues:*

*Unrelated issues (need separate issue)*:
- [ ] Fix test_calc_geisa_spectrum
- [ ] test_OH_AX_*.py tests throw a `NotImplementedError` when the download failed; this should be specified better.
- [ ] store hitran.org credentials using `keyring`
- [ ] run GPU tests on ubuntu servers
- [ ] Windows does not always enable UTF-8 encoding. It is now turned on in testing but it may not be on a user system


.github/workflows/tests-develop.yml
```diff
@@ -10,6 +10,41 @@ on:
     branches: [ "develop" ]
 
 jobs:
+  fetch:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
```
minouHub, line 19: Do we need 30 minutes to fetch all databases?

.github/workflows/tests-develop.yml
```diff
@@ -10,6 +10,41 @@ on:
...
+      - name: Set-up proxy server
+        run: | # generate SSL certificates, run server in the background, and reroute hitran traffic to the proxy
+          python .github/proxy/generate_certificates.py
+          ${{ runner.os != 'Windows' && 'sudo ' || '' }}python .github/proxy/reverse_proxy_server.py &
+          ${{ runner.os != 'Windows' && 'sudo ' || '' }}python .github/proxy/append_hosts.py
+      - name: Test database fetching
+        env:
+          HITRAN_EMAIL: ${{ secrets.HITRAN_EMAIL }}
+          HITRAN_PASSWORD: ${{ secrets.HITRAN_PASSWORD }}
+        run: |
```
minouHub, line 55: Not sure if you know:

- In GitHub actions, I removed the tests which require HITEMP login with the key "not needs_HITRAN_credentials". On GitHub, only admins can use secret variables in their PR.
- The test of HITRAN login is still done with Travis.

After making this comment... I realize we could still do the tests of HITRAN when merging to the main because only admin can do these PR

minouHub (commented):
Not the best reviewer but let's add a few comments anyway. I understood that you want to two steps:

1. run once a job that downloads the database (Ubuntu x python 3.12) 
2. run all the tests on multiple python versions and OS

Thus ... do you need to do all the tests "require_connection" in fetch?

dcmvdbekerom:
Finally, the concept is working! 

In the ci test, you'll see that the proxy test returns MISS the first time, implying the file is downloaded from hitran.org. In the consecutive parallel tests, you'll see a HIT, meaning it serves from the local cache, no download.

Note that on the test_proxy.py file, there are no instructions to tell the file it is talking to a proxy, so no modifications to radis code are required.

The windows and latest are failing atm because the test is set up for Linux; changing the commands to windows and Mac specific commands will solve that.

.github/workflows/tests-develop.yml
dcmvdbekerom: This is a timeout, just in case it gets stuck. it should be much faster, something like 5 mins.

.github/workflows/tests-develop.yml
dcmvdbekerom: I didn't know, is there an issue for this?

dcmvdbekerom:
> Thus ... do you need to do all the tests "require_connection" in fetch?

No, only the `connection_required` tests need to be ran, I'll update this. We could then even skip these ones in the later test on the platform that was used for the fetch.

minouHub:
**Note for myself:** at this stage, the structure of the tests passes but the tests are not ran.

PR #842, opened 2025-06-27
feat: Add non-equilibrium electronic spectra support (OH A-X from ExoMol)

darshvn:
### What this does

Extends RADIS's existing Telec (electronic temperature) support from atoms to molecules. Previously Telec only worked for atomic spectra. Now you can compute non-equilibrium molecular spectra where electronic, vibrational, and rotational temperatures are all different. Demonstrated on the OH A-X band (~308 nm) using the ExoMol MoLLIST-OH database.

Closes #425

---

### How electronic spectra are calculated

The physics relies on Born-Oppenheimer separation: each molecular energy level splits into electronic, vibrational, and rotational parts, each at its own temperature.

```
E_total = Te + Evib + Erot
```

The population of each state follows a three-temperature Boltzmann distribution:

```
n(state) = g * exp(-Te/kTelec) * exp(-Evib/kTvib) * exp(-Erot/kTrot) / Q
```

where the partition function Q sums this over all states. Here is how RADIS computes this, step by step:

**Step 1: Load and separate energies**
When `sf.fetch_databank("exomol", "MoLLIST-OH", load_energies=True)` is called, RADIS downloads the ExoMol `.states.bz2` file containing every rovibronic level (total energy, degeneracy, quantum numbers v, J, electronic state label). `PartFuncExoMolStates` in `partfunc.py` parses this file and separates the total energy into three pieces:
- Te = minimum energy of v=0 levels for each electronic state (e.g., Te=0 for X, Te~32588 cm-1 for A)
- Evib = minimum energy of each (electronic state, v) group minus Te
- Erot = remainder: E - Te - Evib

This separation is done once at load time and cached.

**Step 2: Map transitions to separated energies**
Each transition in the line database has upper and lower total energies (Eu, El). In `_calculate_exomol_energies()` (`base.py`), these are matched to the `.states` data using `np.searchsorted` for fast nearest-neighbor lookup. After this step, every transition carries Evibu, Erotu, Te_u and their lower-state counterparts.

**Step 3: Assign bands**
OH is a CLASS3 diatomic. `add_bands()` in `bands.py` labels transitions by vibrational band (vl -> vu) when quantum numbers are available, or falls back to generic labels for ExoMol formats that lack them.

**Step 4: Compute populations**
When `sf.non_eq_spectrum(Ttrans, Trot, Tvib, Telec)` is called, `calc_populations_noneq()` in `base.py` builds each state's population from three independent Boltzmann factors:
[...]

PR #1031, opened 2026-07-19
Resolve Sphinx cross-reference warnings caused by lazy-loading re-exports 

prathamhole14:
### Description

To make the lazy loading work, in #842 explicitly defined `__all__` lists in these files. However, Sphinx's autodoc extension blindly trusts `__all__`. As soon as it saw classes like Spectrum and `SpecDatabase` explicitly exported in multiple `__init__.py` files, it started indexing and documenting them in all of those locations simultaneously.

This caused core classes to be documented two or three times (e.g., Spectrum was indexed as `radis.Spectrum`, `radis.spectrum.Spectrum`, and `radis.spectrum.spectrum.Spectrum`). This massive double-indexing resulted in over 400 duplicate object description and more than one target found for cross-reference warnings during make html-noplot.

### DIFF
--- a/docs/_templates/autosummary/module.rst
+++ b/docs/_templates/autosummary/module.rst
@@ -4,3 +4,4 @@
    :members:
    :show-inheritance:
    :undoc-members:
+   :no-index:
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -184,9 +184,29 @@ def _fix_math_backslashes(app, what, name, obj, options, lines):
             in_math = False
 
 
+def _skip_reimported_members(app, what, name, obj, skip, options):
+    """Skip members that are re-imported from submodules."""
+    # We only care about objects that have a __module__ attribute
+    obj_module = getattr(obj, "__module__", None)
+    if not obj_module:
+        return skip
+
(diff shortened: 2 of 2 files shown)


minouHub:
Not sure I understand sorry. I still find three occurrences of Spectrum:

- https://radis--1031.org.readthedocs.build/en/1031/source/radis.spectrum.html
- https://radis--1031.org.readthedocs.build/en/1031/spectrum/spectrum.html
- https://radis--1031.org.readthedocs.build/en/1031/source/radis.spectrum.spectrum.html

prathamhole14:
Sorry, I wasn't clear. The PR doesn't remove those entries from the pages it removes the duplicate cross-reference targets. `:no-index:` still renders the class, it just stops Sphinx registering it a second and third time as a target for `class:Spectrum`, which is what was generating the warnings.

Also one of those three isn't autodoc spectrum/spectrum.html is the hand-written guide, it just links to Spectrum rather than documenting it.

The measurable effect is the warning count in the build log (400+ "duplicate object description" / "more than one target found" before).

PR #1035, opened 2026-08-09
Detect truncated HITRAN downloads

prathamhole14:
Heads up: HAPI was silently swallowing incomplete responses and returning truncated line lists without throwing an error. We now check that downloaded files end on a clean record boundary, retrying if they don't, and raising an exception rather than passing along partial data.

Also, instead of guessing isotopes 1–9 and parsing the resulting KeyErrors, we now pull them directly from HAPI's isotopologue table.

### DIFF
--- a/radis/api/hitranapi.py
+++ b/radis/api/hitranapi.py
@@ -107,18 +107,18 @@ columns_2004 = OrderedDict(
 # fmt: on
 
 
-def get_hitran_isotopes(molecule, max_isotope=9):
-    """Isotopes of ``molecule`` available in HITRAN, read from HAPI's
+def get_hitran_isotopologues(molecule, max_isotopologue=9):
+    """Isotopologues of ``molecule`` available in HITRAN, read from HAPI's
     isotopologue table.
 
-    Isotopes above ``max_isotope`` (ex: 10-12 for CO2) are left out: they have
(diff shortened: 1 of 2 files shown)


dcmvdbekerom:
I came across this issue also, the hitran servers return smaller files than they report. Do you know why this happens?

minouHub (approved):
I'm glad you cleaned a bit this portion of code! Thank you

Small comment. The exact word for molecules is "isotopologues" -- instead of "isotopes" which is for atoms. I know this mistake is everywhere in RADIS, but let's have the new code with the proper wording please 🙏

prathamhole14:
> I came across this issue also, the hitran servers return smaller files than they report. Do you know why this happens?

Yeah, dug into it turns out nothing's actually missing. HITRAN computes `Content-Length1 as if records ended in CRLF, but sends them with just LF, so every response comes up short by exactly one byte per line.

- CO: declared 871722, got 866341 so 5381 short, and CO has exactly 5381 lines
- H2O iso 1: declared 52507602, got 52183481 so 324121 short, 324121 lines

So `delivered =n_lines × 161` and `declared = n_lines × 162`. The files are complete, the header's just wrong. Side effect is that `Content-Length` is useless for spotting a genuinely truncated download from HITRAN

PR #1037, opened 2026-08-13
Fix NumPy 2.5 string dtype compatibility

HajimeKawahara:
## Summary

- Normalize legacy fixed-width string dtype codes from `aN` to `SN` in `_format_dtype`.
- Add a fast regression test for the `a2` to `S2` conversion.

## Root cause

NumPy 2.5 removed the deprecated `a` dtype alias in favor of `S`. RADIS still passes codes such as `a1`, `a2`, and `a25` to `np.dtype` while parsing HITRAN-family fixed-width files, which raises:

```text
TypeError: data type 'a2' not understood
```

The normalization is kept in the shared formatter so existing column definitions and the HITRAN, HITEMP, CDSD, and GEISA parser paths continue to work without broader changes.

Related downstream issue: HajimeKawahara/exojax#735

## Validation

Using Python 3.12 and NumPy 2.5.2:

```text
5 passed in 0.51s
```

The selected tests cover the new regression test and the local HITRAN/HITEMP parser tests. The same five tests also pass with NumPy 2.1.0.

### DIFF
--- a/radis/api/tools.py
+++ b/radis/api/tools.py
@@ -188,9 +188,11 @@ def _format_dtype(dtype):
         formatted_dtype = [
             (
                 str(k),
-                f"S{c[1:]}"
-                if isinstance(c, str) and c.startswith("a") and c[1:].isdigit()
-                else c,
+                (
+                    f"S{c[1:]}"
+                    if isinstance(c, str) and c.startswith("a") and c[1:].isdigit()
+                    else c
+                ),
             )
             for k, c in dtype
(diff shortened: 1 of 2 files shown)


minouHub (approved):
I addressed this with [c8136f8](https://github.com/radis/radis/pull/860/commits/c8136f8629c291a4a173fc1e2aeb7ff4e756fd1f) but you were faster! Your solution is more elegant anyway. Approved

PR #1040, opened 2026-08-26
fix: forward-compatibility with matplotlib 3.13 and pandas 3.0

prathamhole14:
### Description
Fix two upstream deprecations that will become errors in upcoming releases.

`plot_diff` passes `fig.canvas` to matplotlib's `MultiCursor`. This is deprecated in 3.11 and will cause an error in 3.13. We can't just drop it because matplotlib < 3.11 (which Python 3.10 users get) still needs it, so we added a version check.

`test_hdf5.py` passes `key` to `to_hdf` positionally. Pandas 3.0 requires it to be a keyword argument. The behavior stays exactly the same.

#### Developer Confirmation
- [x] I understand all code changes in this PR (whether self-authored or AI-assisted).
- [x] I confirm that all changes are necessary and useful.

#### Checklist before review by admins:
- [x] This PR passes [linting](https://radis.readthedocs.io/en/latest/dev/developer.html#code-style-and-linting).
- [x] This PR passes all related tests.

If a test fails due to an external issue (e.g. database not downloaded), please post a comment rather than re-running the tests.

### DIFF
--- a/examples/6_Misc/plot_styles.py
+++ b/examples/6_Misc/plot_styles.py
@@ -48,7 +48,7 @@ for plotlib, context, style in [
     plt.title(", ".join(f"{v}" for v in radis.config["plot"].values()))
     plt.tight_layout()
 
-#%%
+# %%
 # You can also choose to not show the figure immediately, and instead return
 # the Matplotlib or Plotly object. This allows for **advanced customization**
 # (e.g., modifying the matplotlib figure).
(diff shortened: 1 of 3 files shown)


minouHub:
Oh gosh .... I still feel the pain of the transition to pandas 2.0

minouHub (changes requested):
Quick review. There is an issue in the docs, see https://app.readthedocs.org/projects/radis/builds/34246428/

`plot_chain_spectrum_edition.py`, `plot_fit_lineshape.py` and probably others are failing due to matplotlib issues. Probably not related to this PR changes but it is the same subject.

minouHub:
Note: vaex is not compatible with matplotlib>=3.11
https://github.com/vaexio/vaex/issues/2490

PR #1041, opened 2026-09-05
ci: add zizmor to lint GitHub Actions workflows

prathamhole14:
[zizmor](https://docs.zizmor.sh/) is a static analysis tool for GitHub Actions. It catches things like script injection, leaked tokens and over-broad job permissions.

I added it as a pre-commit hook, so it runs in the existing linting workflow and locally too. No new workflow file needed.

It found a few real issues, fixed in the same PR so CI stays green:

- `publish-Pypi.yml` put the release tag straight into a shell script with `${{ }}`. A tag name can contain shell code, so it now goes through `env` instead.
- `checkout` was leaving the job token in `.git/config`, where any later step could read it. Set `persist-credentials: false`. Nothing in these workflows pushes, so this is safe.
- Test and lint jobs ran with the default write permissions. Narrowed to `contents: read`.

Commit hash pinning is turned off in `.github/zizmor.yml`, since there is no bot here to keep hashes updated. Tag pins like `@v4` are fine.

Checked with `pre-commit run zizmor --all-files`, passes.


prathamhole14:
I will merge this in next 2 days since the changes are not critical.

minouHub (approved):

